PR #1, opened 2018-05-16
Fix: AppExtension Safe API Only

jleandroperez:
### Description:
Since WordPressKit is now being imported via **cocoapods**, and used within App Extensions, we're getting the `App Extension Safe API Only` flag enabled by default.

This translates into `UIApplication.shared` not being available anymore. This was being used by:

  **WordPressOrgXMLRPCApi** >> **HTTAuthenticationAlertController** >> **UIAlertController+Helpers**

Context: Displaying an AlertController to determine how to handle a HTTP Challenge.

### Details:
In this PR we're implementing a mechanism that relays back to the Host App the `Challenge` received by NSURLSession.

Since `HTTPAuthenticationAlertController` is open, the client app is free to instantiate the Alert, display it, and hit the callback whenever the user picks up an option.

If no callback is set, we'll go ahead with the default NSURLSession handling mechanism.

### Testing:
1. Checkout [this WordPress iOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/9386)
2. Attempt to log into a Self Hosted site (w/ self signed certificate)

Verify that an alert shows up!


SergioEstevao:
@jleandroperez just for me to understand on WordPressComKit there is not formal pod release version? is all done by pointing to master?

jleandroperez:
@SergioEstevao This one `WordPressKit` (which is not the same as `WordPressComKit`) *is* released via CocoaPods (1.0.1), and the idea is to roll out a new version with this fix.

But current WPiOS branch points to master + hash, because... this PR needs to be included for the integration to work. I'll switch to a formal release once the PR's are approved!

jleandroperez:
Thanks A LOT @SergioEstevao !!!!

PR #6, opened 2018-06-05
Remove AFNetworking part 2: Refactor WordPressComRestAPI

SergioEstevao:
This PR updates the WordPressComRestAPI to use AlamoFire instead of AFNetworking.

On the way I move the check for calls failing because of too many requests to the inside of the API object instead of being done on the Remote object.

How to test:
 - Make sure the unit tests are passing.
 - Test on the main app using this branch: `try/remove_af_networking_from_wpkit`
 - Because of the amount of places this code is used check several parts of the app:
  -  Post and page lists;
  -  Reader;
  - Site List;
  - Notifications;
  - Editor
  - Media Library, upload multiple images and videos and check memory impact
  - See if there is any slow downs while scrolling around
 - Test on a device!

### DIFF
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -128,6 +128,46 @@ open class WordPressComRestApi: NSObject {
 
     // MARK: - Network requests
 
+    private func request(method: HTTPMethod,
+                         urlString: String,
+                         parameters: [String: AnyObject]?,
+                         encoding: ParameterEncoding,
+                         success: @escaping SuccessResponseBlock,
+                         failure: @escaping FailureReponseBlock) -> Progress? {
+
+        guard let URLString = buildRequestURLFor(path: urlString) else {
+            let error = NSError(domain: String(describing: WordPressComRestApiError.self),
+                                code: WordPressComRestApiError.requestSerializationFailed.rawValue,
+                                userInfo: [NSLocalizedDescriptionKey: NSLocalizedString("Failed to serialize request to the REST API.", comment: "Error message to show when wrong URL format is used to access the REST API")])
+            failure(error, nil)
+            return nil
+        }
+
+        let progress = Progress(totalUnitCount: 1)
+        let progressUpdater = {(taskProgress: Progress) in
+            progress.totalUnitCount = taskProgress.totalUnitCount
+            progress.completedUnitCount = taskProgress.completedUnitCount
+        }
+
+        let dataRequest = sessionManager.request(URLString, method: method, parameters: parameters, encoding:encoding)
+            .validate()
+            .responseJSON(completionHandler: { (response) in
+            switch response.result {
+            case .success(let responseObject):
+                progress.completedUnitCount = progress.totalUnitCount
+                success(responseObject as AnyObject, response.response)
+            case .failure(let error):
+                let nserror = self.processError(response: response, originalError: error)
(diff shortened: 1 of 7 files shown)


WordPressKit/WordPressComRestApi.swift
```diff
@@ -417,13 +370,29 @@ final class WordPressComRestAPIResponseSerializer: AFJSONResponseSerializer {
...
+            responseString.contains("Limit reached") else {
+                return nil
+        }
+        var userInfo =  [String: Any]()
+        userInfo[WordPressComRestApi.ErrorKeyErrorCode] = "too_many_requests"
+        userInfo[WordPressComRestApi.ErrorKeyErrorMessage] = NSLocalizedString("Limit reached. You can try again in 1 minute. Trying again before that will only increase the time you have to wait before the ban is lifted. If you think this is in error, contact support.", comment: "Message to show when a request for a WP.com API endpoint is throttled")
+        userInfo[NSLocalizedDescriptionKey] = userInfo[WordPressComRestApi.ErrorKeyErrorMessage]
+        let nsError = WordPressComRestApiError.tooManyRequests as NSError
+        let errorWithLocalizedMessage = NSError(domain: nsError.domain, code: nsError.code, userInfo:userInfo)
+        return errorWithLocalizedMessage
```
jleandroperez: This would look GREAT GREAT GREAT as a Swift native `Error`

WordPressKit/WordPressComRestApi.swift
```diff
@@ -417,13 +370,29 @@ final class WordPressComRestAPIResponseSerializer: AFJSONResponseSerializer {
...
+    }
+
+    func checkForThrottleErrorIn(data: Data) -> NSError? {
+        // This endpoint is throttled, so check if we've sent too many requests and fill that error in as
+        // when too many requests occur the API just spits out an html page.
+        guard let responseString = String(data:data, encoding:.utf8),
+            responseString.contains("Limit reached") else {
+                return nil
+        }
+        var userInfo =  [String: Any]()
```
jleandroperez: Nipticky: Extra space between `=` and `[String: Any]`

WordPressKit/WordPressComRestApi.swift
```diff
@@ -238,93 +229,54 @@ open class WordPressComRestApi: NSObject {
                               requestEnqueued: RequestEnqueuedBlock? = nil,
                               success: @escaping SuccessResponseBlock,
                               failure: @escaping FailureReponseBlock) -> Progress? {
-        
+
+        guard let URLString = buildRequestURLFor(path: URLString) else {
+            let error = NSError(domain: String(describing: WordPressComRestApiError.self),
+                                code: WordPressComRestApiError.requestSerializationFailed.rawValue,
+                                userInfo: [NSLocalizedDescriptionKey: NSLocalizedString("Failed to serialize request to the REST API.", comment: "Error message to show when wrong URL format is used to access the REST API")])
+            failure(error, nil)
```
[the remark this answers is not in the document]
SergioEstevao: While I love do to do I wanted to minimize the impact on the rest of the code, that will cause deeper changes all the way to the Remote classes. I think that will be easier to do when we port the Remotes to swift.

WordPressKit/WordPressComRestApi.swift
SergioEstevao: Same as the above comment.

PR #7, opened 2018-06-06
Add ReaderSiteSearchServiceRemote and tests

frosty:
This PR looks big, but most of it is mock data!

This PR adds a new remote: `ReaderSiteSearchServiceRemote`, which I'll use to fetch a list of possible Reader sites for a user to follow: https://github.com/wordpress-mobile/WordPress-iOS/issues/9433. The remote has a single method to perform a search, and returns the results.

Results should include a feed ID, a title, and a URL. They may also include an optional blavatar URL and description.

Equivalent Android PR: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/822

**To test:**

* Check the code
* Build and run tests, and check everything passes
* Let me know if you'd like to actually try it out with WPiOS and I can push up a work in progress branch there that uses this remote.

PR #12, opened 2018-06-07
ReaderFeed: Parse blog ID if available

frosty:
Following on from the recent addition of `ReaderSiteSearchServiceRemote`: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/7

It turns out that not all feeds returned by the endpoint have a `feed_ID` value, and we were parsing that as a required property. I've made a few tweaks so that now:

* If feed ID doesn't exist, we can try and use blog ID instead
* If neither exist, we remove the feed from the list of results

I've also updated the unit tests to handle this.

**To test:**

* Build and run the tests, check they pass.
* Check out branch `issue/9433-reader-site-search` of WPiOS.
* Go to Reader > Search, and search for `macos 10.13.5`. You should see an error in the console:

```
2018-06-07 18:18:37:155 WordPress[50072:6170454] Error while performing Reader site search: decodingFailure
```
* Update the Podfile to point to this branch:

```
pod WordPressKit, :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :branch => 'fix/reader-feed-parsing'
```
* Perform the same search as before. You should see one result in the Sites tab, and no error in the console.

@jleandroperez Sorry, would you mind one more?

### DIFF
--- a/WordPressKit.xcodeproj/project.pbxproj
+++ b/WordPressKit.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		170589D520C9923500774C6F /* reader-site-search-blog-id-fallback.json in Resources */ = {isa = PBXBuildFile; fileRef = 170589D320C9923500774C6F /* reader-site-search-blog-id-fallback.json */; };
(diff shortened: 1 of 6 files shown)


frosty:
Thanks @jleandroperez! Yes, that's a WPiOS thing – already fixed locally :)

Thanks for the review!

PR #13, opened 2018-06-07
Extend `RemoteBlog` to check for capability like `Blog` does

mindgraffiti:
This PR is a dependency for fixing [WPiOS issue #8621](https://github.com/wordpress-mobile/WordPress-iOS/issues/8621).

This PR duplicates Blog+Capabilities (found in the WordPress-iOS project) to RemoteBlog, and uses lowerCamelCase for enum naming.

### DIFF
--- a/WordPressKit/RemoteBlog+Capabilities.swift
+++ b/WordPressKit/RemoteBlog+Capabilities.swift
@@ -1,0 +1,30 @@
+import Foundation
+
+extension RemoteBlog {
+    /// Enumeration that contains all of the RemoteBlog's available capabilities.
+    ///
+    public enum Capability: String {
+        case deleteOthersPosts  = "delete_others_posts"
+        case deletePosts        = "delete_posts"
+        case editOthersPages    = "edit_others_pages"
+        case editOthersPosts    = "edit_others_posts"
+        case editPages          = "edit_pages"
+        case editPosts          = "edit_posts"
+        case editThemeOptions   = "edit_theme_options"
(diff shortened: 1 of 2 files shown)


WordPressKit/RemoteBlog+Capabilities.swift
```diff
@@ -0,0 +1,48 @@
...
+        case promoteUsers       = "promote_users"
+        case publishPosts       = "publish_posts"
+        case uploadFiles        = "upload_files"
+        case viewStats          = "view_stats"
+    }
+
+
+    /// Returns true if a given capability is enabled. False otherwise
+    ///
+    public func isUserCapableOf(_ capability: Capability) -> Bool {
```
koke: I know this is just replicating an existing thing, and as far as nitpicks go this has to be on my top list, so feel free to ignore this 😁 

It looks to me like `isUserCapableOf` would take a capability in the form of a gerund, but they're currently infinitives, as the original core equivalent would be `current_user_can`. So "current user can delete posts" sounds right, and "is user capable of _deleting_ posts" would, but not "is user capable of _delete_ posts".

koke (approved):

bummytime (changes requested):
@mindgraffiti this PR may not be needed — see this comment for more deets: https://github.com/wordpress-mobile/WordPress-iOS/pull/9545#pullrequestreview-127185961

mindgraffiti:
Canceling PR. There's a better solution linked above

PR #17, opened 2018-06-30
Fix Xcode 10 warnings.

jklausa:
This gets rid of warnings about conflicting accessiblity specifiers in Xcode 10 ("Static implies final, use public instead").


jklausa:
@etoledom Thanks for the review :)!

PR #18, opened 2018-06-30
Improve `Activity` API.

jklausa:
This does two things: 
* removes an unused (and very easily misused) property on `Activity`
* improves handling of situations where API returns 0 `Activities` (currently only true for WPCom free sites which we don't show AL for, but probably will change in the future).


jklausa:
Thanks for the review, sir @jleandroperez !!
!

jleandroperez:
Anytime Jan!!

PR #25, opened 2018-07-12
Update getDomainSuggestions to pass new parameters

pinarol:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/9630 

This is for allowing to search for suggested domains such that not only belong to wordpress.com but also to other domains, because with 9630 we want to allow business plan users to register custom domains with the app. 

With this change we are allowing to set both "include_wordpressdotcom",
"only_wordpressdotcom". Their default values remain as true.

To test: will be tested on https://github.com/wordpress-mobile/WordPress-iOS/pull/9759


ScoutHarris:
@pinar1234 well done! This change looks good. We'll leave it open during your trial project, and in case more changes are needed to support that.

WordPressKit/DomainsServiceRemote.swift
```diff
@@ -130,12 +130,16 @@ public class DomainsServiceRemote: ServiceRemoteWordPressComREST {
...
+                                     includeWordPressDotCom: Bool = true,
+                                     onlyWordPressDotCom: Bool = true,
+                                     success: @escaping ([String]) -> Void,
+                                     failure: @escaping (Error) -> Void) {
         let endPoint = "domains/suggestions"
         let servicePath = path(forEndpoint: endPoint, withVersion: ._1_1)
         let parameters: [String: AnyObject] = ["query": query as AnyObject,
-                                               "include_wordpressdotcom": true as AnyObject,
-                                               "only_wordpressdotcom": true as AnyObject]
+                                               "include_wordpressdotcom": includeWordPressDotCom as AnyObject,
```
jklausa: I might be misunderstanding how this API is supposed to work server-side, but I don't think modeling those options as two `Bool`s is the best choice here. Not sure how the backend would behave, but this API as-is allows you to pass `includeWPCom` as `false` and `onlyWPCom` as `true`, which, logically, should always return empty set :) 

What about a three state enum, like 
```
enum TooTiredToFigureOutASmartNamePleaseForgiveMe {
    case noWordpressDotCom
    case includeWordPressDotCom
    case onlyWordPressDotCom
}
``` 
or something similar?

pinar1234: oh yes you are probably right but It looks like it is something that needs to be fixed on the API side first(I mean the serverside), isn't it?

pinar1234:
Hi, I've sent a new commit "Add getPlansForSite with version V3" getPlansForSite response has huge differences between v2 and v3 so I couldn't merge them in the old response together. 
I named the new one as PlanServiceV3Remote but I am open for suggestions, It doesn't sound very good currently, I just couldn't come up with sth better.

WordPressKit/DomainsServiceRemote.swift
jklausa: Hmm, not really, I think we can model this "correctly" on our end only. If you take the example enum I posted above, you can do something like that:

```swift
let parameters: [String: AnyObject]
switch includeParameter {
     case .noWordPressDotCom: parameters = ["include_wordpressdotcom": false]
     case .includeWordPressDotCom: parameters  = ["include_wordpressdotcom" true, "only_wordpressdotcom": false]
     case .onlyWordPressDotCom: parameters =  ["include_wordpressdotcom" true, "only_wordpressdotcom": true]
}
```

(or something to that effect)

We'd still get a sensible API for this on our end, and the backend can keep it's own weirdness ;P

WordPressKit/JSONDecoderExtension.swift
```diff
@@ -0,0 +1,40 @@
...
+    
+    static var supportMultipleDateFormats: JSONDecoder.DateDecodingStrategy {
+        return JSONDecoder.DateDecodingStrategy.custom({ (decoder) -> Date in
+            let container = try decoder.singleValueContainer()
+            let dateStr = try container.decode(String.self)
+            let dateNoTimeFormatter = DateFormatter()
+            dateNoTimeFormatter.dateFormat = "yyyy-mm-dd"
+            
+            let dateWithTimeFormatter = DateFormatter()
+            dateWithTimeFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
```
ScoutHarris: I don't _think_ having two `DateFormatter`s is necessary. Maybe something like this will work?
```
let dateFormatter = DateFormatter()
...
if len == 10 {
    dateFormatter.dateFormat = "yyyy-mm-dd"
    date = dateFormatter.date(from: dateStr)
} else if len == 19 {
    dateFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
    date = dateFormatter.date(from: dateStr)
}
```

WordPressKit/JSONDecoderExtension.swift
```diff
@@ -0,0 +1,40 @@
...
+            let len = dateStr.count
+            var date: Date? = nil
+            if len == 10 {
```
ScoutHarris: Can we make this `10` and the following `19` constants with descriptive names?

WordPressKit/PlanServiceV3Remote.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import WordPressShared
+import CocoaLumberjack
+
+@objc public class PlanServiceV3Remote: ServiceRemoteWordPressComREST {
```
ScoutHarris: I'm still not sure having a separate class is necessary. `PlanServiceV3Remote` and `PlanServiceRemote` do the same thing. The only difference is how they parse the response and what is returned. Would it not be possible to add an api version parameter to `getPlansForSite`  and decide how to proceed based on that parameter?

WordPressKit/RemotePlanDetail.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+
+@objc public class RemotePlanDetail: NSObject, Codable {
```
ScoutHarris: The name of this class is confusing. Initially I thought it would be details for the `RemotePlan` class. Boy was I wrong! 😉 I would request two things:
1. Rename this to something more specific, like `RemotePlanApiVersion1_3` (or something less crappy).
2. Add some comments as to why this exists and why we're not using `RemotePlan`.

ScoutHarris:
Hey @pinar1234 . I didn't catch this until now. The versions are not v2 or v3. They are v1.2 or v1.3. If something does need to be named specific to 1.3, can we use something like `ApiVersion1_3`? Thanks!!

WordPressKit/PlanServiceV3Remote.swift
pinar1234: I can sure remove the PlanServiceV3Remote class and add a new method to PlanServiceRemote. Actually at the beginning that was what I did that but I had to add a suffix like getPlansForSite_version1_3 I choosed adding the suffix to the class instead but I can revert.

But, on the other hand merging them both into the old getPlansForSite is very problematic. Because then I need to make them return the same response in the success completion, this means merging all parameters of v2 and v3 in the same response model, and there is a huge difference between them:

old one:
```
 "product_slug": "free_plan",
  "description": "",
  "cost": 0,
  "bill_period": -1,
  "product_type": "bundle",
  "available": "yes",
  "multi": 0,
  "bd_slug": "wp-free-plan",
  "bd_variation_slug": "wp-bundles",
  "outer_slug": null,
  "extra": null,
  "capability": "manage_options",
  "product_name_short": "Free",
  "icon": "https:\/\/s0.wordpress.com\/i\/store\/plan-free.png",
  "icon_active": "https:\/\/s0.wordpress.com\/i\/store\/plan-free-active.png",
  "bill_period_label": "for life",
  "price": "$0",
  "formatted_price": "$0",
  "raw_price": 0,
  "tagline": "Just get started",
  "currency_code": "USD",
  "features_highlight": [
                         {
                         "items": [
                                   "free-blog",
                                   "space",
                                   "support"
                                   ]
                         }
                         ],
  "formatted_original_price": "$0",
  "raw_discount": 0,
  "formatted_discount": "$0",
  "current_plan": true
  }
```
new one:
```
  "1008" : {
    "current_plan" : true,
    "discount_reason" : null,
    "currency_code" : "USD",
    "user_facing_expiry" : "2019-07-07",
    "auto_renew" : true,
    "product_name" : "WordPress.com Business",
    "formatted_original_price" : "$0",
    "formatted_price" : "$300",
    "raw_discount" : 0,
    "product_slug" : "business-bundle",
    "has_domain_credit" : true,
    "raw_price" : 300,
    "free_trial" : false,
    "id" : "10872031",
    "is_domain_upgrade" : null,
    "subscribed_date" : "2018-07-10 19:41:28",
    "auto_renew_date" : "2019-06-10",
    "expiry" : "2019-07-10",
    "user_is_owner" : true,
    "interval" : 365,
    "formatted_discount" : "$0"
  }
```

WordPressKit/JSONDecoderExtension.swift
pinar1234: sure

WordPressKit/DomainsServiceRemote.swift
pinar1234: Ok sure I'll update it on our side. I am just trying to understand how things are moving forward between back end and client teams

WordPressKit/JSONDecoderExtension.swift
pinar1234: + I did some additional refactoring

WordPressKit/JSONDecoderExtension.swift
pinar1234: I refactored this

WordPressKit/DomainsServiceRemote.swift
pinar1234: I've updated this

WordPressKitTests/TransactionsServiceTests.swift
```diff
@@ -0,0 +1,35 @@
+import XCTest
+import WordPressShared
+import CocoaLumberjack
+@testable import WordPressKit
+
+class TransactionsServiceTests: RemoteTestCase, RESTTestable {
```
ScoutHarris: 👍 for adding new tests!!

Looking at other remote tests, it looks like they all have `remote` in the class name and file name. Can we rename this to something like `TransactionsServiceRemoteTests`?

WordPressKitTests/TransactionsServiceTests.swift
```diff
@@ -0,0 +1,35 @@
...
+        let expect = expectation(description: "Get supported countries success")
+        
+        stubRemoteResponse("me/transactions/supported-countries/",
+                           filename: supportedCountriesSuccessFileName,
+                           contentType: .ApplicationJSON,
+                           status: 200)
+        
+        remote.getSupportedCountries(success: { (countryList) in
+            expect.fulfill()
+            XCTAssert(countryList.count == 239)
```
ScoutHarris: This seems awfully fragile. If the API adds or removes a supported country, the test will fail. I'm not sure that's a good reason for it to. Is that what we really want to happen?

WordPressKitTests/TransactionsServiceTests.swift
pinar1234: Sure

WordPressKitTests/TransactionsServiceTests.swift
pinar1234: service response is mocked on the client side so as soon as nobody changes the mock data it won't fail. Actually I don't see any other way. I need to check if the response parsing is successful, do you have any other suggestions?

WordPressKit/DomainsServiceRemote.swift
```diff
@@ -26,6 +26,27 @@ public class DomainsServiceRemote: ServiceRemoteWordPressComREST {
         case decodingFailed
     }
 
+    public enum DomainSuggestionType {
+        case noWordpressDotCom
+        case includeWordPressDotCom
+        case onlyWordPressDotCom
+        
+        fileprivate func parameters() -> [String: AnyObject] {
+            var result: [String: AnyObject] = [:]
```
jklausa: nitpick: if you don't assign an initial value, this could be a `let`

WordPressKit/JSONDecoderExtension.swift
```diff
@@ -0,0 +1,47 @@
...
+            let dateStr = try container.decode(String.self)
+            
+            let len = dateStr.count
+            var date: Date? = nil
+            if len == DateFormat.noTime.rawValue.count {
+                date = DateFormat.noTime.formatter.date(from: dateStr)
+            } else if len == DateFormat.dateWithTime.rawValue.count {
+                date = DateFormat.dateWithTime.formatter.date(from: dateStr)
+            }
+            guard let _date = date else {
```
jklausa: stylistic nitpick: I don't like the use of underscore here — I think it's fairly incosistent with the rest of the codebase.

I think this would read better like this:
```swift
let date: Date

switch len {
case DateFormat.noTime.rawValue.count:
    date = DateFormat.noTime.formatter.date(from: dateStr)
case DateFormat.dateWithTime.rawValue.count:
    date = DateFormat.dateWithTime.formatter.date(from: dateStr)
default:
    throw DecodingError.dataCorruptedError(
                     in: container,
                     debugDescription: "Cannot decode date string \(dateStr)"
                 )
}
```

That way you get to avoid both optionals and the mutability here :) (and I think it's slightly more clear what the code is doing, but that's a personal thing)

WordPressKit/PlanServiceRemote_ApiVersion1_3.swift
```diff
@@ -0,0 +1,69 @@
...
+                }
+            },
+            failure: {
+                error, _ in
+                failure(error)
+            }
+        )
+    }
+    
+    private static func mapPlansResponse(
```
jklausa: style nitpick: formatting here is weird. i think having the method signature on line line would be cleaner ;)

WordPressKit/State.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+@objc class State: NSObject, Codable {
```
jklausa: A one-line comment explaining what we're using this for would be good — when I think of `State` in the context of programming, this isn't the first thing that comes to mind :)

jklausa: (A similar comment could be added to `Country.swift`, however there the possibility of confusion is much smaller :))

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
```diff
@@ -114,4 +115,25 @@ class DomainsServiceRemoteRESTTests: RemoteTestCase, RESTTestable {
         
         waitForExpectations(timeout: timeout, handler: nil)
     }
+    
+    func testGetStatesSuccess() {
```
jklausa: I think it would be also useful to add a test that verify that we handle the countries that have no meaningful division into "states" correctly. (like `DE` or `PL` for example)

WordPressKit/PlanServiceRemote_ApiVersion1_3.swift
pinar1234: it is a very long line since return type is a tuple, I am updating it as below, how does that look?
    private static func mapPlansResponse(_ response: AnyObject) throws
        -> (activePlan: RemotePlan_ApiVersion1_3, availablePlans: [RemotePlan_ApiVersion1_3]) {

WordPressKit/JSONDecoderExtension.swift
pinar1234: oh I like what you suggested here, the thing is DateFormat.noTime.formatter.date(from: dateStr) also returns Optional result so I could only partially apply your solution. In the new commit I got rid of the underscore also.

WordPressKit/DomainsServiceRemote.swift
pinar1234: Done

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
pinar1234: for those countries the service returns empty array so I added a new test case for handling empty result, is that what you meant?

jklausa:
I've taken over this work in #43.

PR #26, opened 2018-07-27
Add support for AL for WPCom free sites.

jklausa:
This adds a bunch of glue code needed for the WPiOS app not to freak out when trying to fetch the `Rewind` status for non-Atomic WPCom sites. We shouldn't need it long-term, but we want to ship this before next code freeze, so c'est la vive.


aerych:
:shipit:

aerych:
Thank ya sir!

PR #27, opened 2018-07-27
1.2 Release.

jklausa:
Let's tag a release before upcoming code freeze, so we can pin the version in WPiOS to a release, not just a raw git SHA hash.


aerych:
:shipit: !

PR #28, opened 2018-07-27
Adds "external_users_only" field to remote publicize connections.

aerych:
A new field was added to the API results.  This PR stores its value in `externalUsersOnly`. 
Podspec is bumped to v1.3


aerych:
Thanks @jklausa !!!

PR #29, opened 2018-07-27
Tweak podspec number so we're bumping from 1.2 to 1.2.1, instead of 1.3

aerych:
Tweaking the podspec version to go back to a three digit schema.  Looks like that's what we were using prior to 1142bd739109190a72a51f0a6c3be545abcdfcb6.  With this change we'll be consistent with the naming schema in our other pods.


aerych:
Thank you sir!

PR #30, opened 2018-08-13
Support for Automated Transfer.

jklausa:
I'm still working on PR for the WPiOS part of it, but the WPKit part is ready, so I've decided to open it already :)

This adds support for WPKit to perform AT operations — checking for eligibility, initiating the process itself and checking the status of an operation.

It'll be used in WPiOS to let users kick-off AT by installing a plugin on a WPCom site with a business plan.

To test:
1. Make sure it builds
2. Read the code and see if it makes sense
3. If you don't trust me, wait for the WPiOS PR and verify it in action :P

### DIFF
--- a/WordPressKit/AutomatedTransferService.swift
+++ b/WordPressKit/AutomatedTransferService.swift
@@ -1,0 +1,137 @@
+import Foundation
+import Alamofire
+
+/// Class encapsualting all requests related to performing Automated Transfer operations.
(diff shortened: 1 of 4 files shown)


WordPressKit/AutomatedTransferService.swift
```diff
@@ -0,0 +1,137 @@
+import Foundation
+import Alamofire
```
etoledom: Is it necessary to import Alamofire?

PR #31, opened 2018-08-28
Mark as sticky Post

danielebogo:
This is related to this: [Mark as Sticky Post](https://github.com/wordpress-mobile/WordPress-iOS/issues/9661)
I updated the `RemotePost` model to map the `sticky` value from the JSON response. I also updated the methods to decode/encode the `RemotePost` model in order to create or update the model whenever it's used.
`PostServiceRemoteRESTTests.m` is updated using the new property as well.

**To Test**
- Make sure it builds
- Run the test

### DIFF
--- a/WordPressKit/PostServiceRemoteREST.m
+++ b/WordPressKit/PostServiceRemoteREST.m
@@ -363,6 +363,9 @@ static NSString * const RemoteOptionValueOrderByPostID = @"ID";
     post.commentCount = [jsonPost numberForKeyPath:@"discussion.comment_count"] ?: @0;
     post.likeCount = [jsonPost numberForKeyPath:@"like_count"] ?: @0;
 
+    NSNumber *stickyPost = [jsonPost numberForKeyPath:@"sticky"] ?: @0;
+    post.isStickyPost = stickyPost.boolValue;
+    
     // FIXME: remove conversion once API is fixed #38-io
     // metadata should always be an array but it's returning false when there are no custom fields
     post.metadata = [jsonPost arrayForKey:@"metadata"];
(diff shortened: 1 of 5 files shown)


WordPressKit/PostServiceRemoteREST.m
```diff
@@ -435,6 +438,8 @@ - (NSDictionary *)parametersWithRemotePost:(RemotePost *)post
     }
     parameters[@"featured_image"] = post.postThumbnailID ? [post.postThumbnailID stringValue] : @"";
     parameters[@"metadata"] = [self metadataForPost:post];
+    
+    parameters[@"sticky"] = post.isStickyPost ? @"1" : @"0";
```
aerych: I'm not %100 sure if the endpoint will accept string values of "1" and "0" correctly.  I did a quick test at `https://developer.wordpress.com/docs/api/console/` and tried creating new posts. The endpoint interprets a numeric 1 as true, numeric 0 as false, string "true" as true and string "false" as false, but a string "1" is not evaluated to true... it evaluated to false (if all that makes sense). 

Maybe double check and if you get the same results as I did we might change this line accordingly.

aerych:
> Is it ok for you if I change it to parameters[@"sticky"] = post.isStickyPost ? @"true" : @"false"; ?

I'm a fan. :)

danielebogo:
@aerych @jklausa I also update the `podspec` to a minor version 1.3.0

aerych:
Since we updated the podspec, can we also update the `Podfile.lock` so its consistent?

danielebogo:
@aerych done!

aerych:
:shipit: :)

PR #33, opened 2018-09-03
Sticky Post for self-hosted sites/XML-RPC-API 

danielebogo:
This PR is a followup of [this task](https://github.com/wordpress-mobile/WordPress-iOS/pull/10077).
Following @SergioEstevao suggestion, I updated the decode/encode methods to handle the `sticky` JSON property also for self-hosted non-Jetpack blogs.

### DIFF
--- a/WordPressKit/PostServiceRemoteXMLRPC.m
+++ b/WordPressKit/PostServiceRemoteXMLRPC.m
@@ -319,6 +319,9 @@ static NSString * const RemoteOptionValueOrderByPostID = @"ID";
     NSArray *terms = [xmlrpcDictionary arrayForKey:@"terms"];
     post.tags = [self tagsFromXMLRPCTermsArray:terms];
     post.categories = [self remoteCategoriesFromXMLRPCTermsArray:terms];
+    
+    NSNumber *stickyPost = [xmlrpcDictionary numberForKeyPath:@"sticky"] ?: @0;
+    post.isStickyPost = stickyPost.boolValue;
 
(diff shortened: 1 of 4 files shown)


WordPressKit/PostServiceRemoteXMLRPC.m
```diff
@@ -409,6 +412,8 @@ - (NSDictionary *)parametersWithRemotePost:(RemotePost *)post
     if ([post.metadata count] > 0) {
         postParams[@"custom_fields"] = post.metadata;
     }
+    
+    postParams[@"sticky"] = post.isStickyPost ? @"true" : @"false";
```
jklausa: This looks a bit suspect — we're assuming that backend will send us a `1`/`0` value (judging by the `numberForKeyPath` in L323), but we're sending it a string `"true"` and `"false"`?

danielebogo: Yes I kept the same behaviour introduced here `PostServiceRemoteREST.m`. Initially I used `1` or `0` but then I switched to `true/false`. Take a look [here](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/31#issuecomment-416931951). I tested `PostServiceRemoteXMLRPC` passing `true/false` and it works.

jleandroperez:
Note: While catching up with notifications, i've noticed the branch was merged but not deleted. Just doing some clean up (branch nuked!).

Nice work Daniele!!

PR #37, opened 2018-09-14
Issue/make active plan optional

aerych:
This PR tweaks the remote Plans service and model making the active plan optional, and not throwing an exception when an active plan is not matched to one of the plans returned from the v1.2 endpoint.   This will allow WPiOS to still show the plans returned from the endpoint instead of presenting an error.
See: https://github.com/wordpress-mobile/WordPress-iOS/issues/9856

It also removes the test for an empty response since that test was, in practice, checking for an exception when an active plan was not found. 

To test: 
Run tests and confirm they pass. 

@jklausa could I trouble you for a review of this one?


aerych:
Thanks Jan!

PR #39, opened 2018-09-28
Feature/quick start award

danielebogo:
This PR adds `markQuickStartChecklistAsComplete` function to the `SiteManagementServiceRemote`, in order to trigger a masterbar notification celebrating completion of mobile quick start.

**To Test:**
- Run Unit Test

### DIFF
--- a/WordPressKit/SiteManagementServiceRemote.swift
+++ b/WordPressKit/SiteManagementServiceRemote.swift
@@ -100,6 +100,27 @@ open class SiteManagementServiceRemote: ServiceRemoteWordPressComREST {
                 failure?(error)
         })
     }
+    
+    /// Trigger a masterbar notification celebrating completion of mobile quick start.
+    ///
+    /// - Parameters:
+    ///   - siteID: The WordPress.com ID of the site.
+    ///   - success: Optional success block
+    ///   - failure: Optional failure block with NSError
+    ///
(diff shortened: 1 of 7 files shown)


WordPressKit/SiteManagementServiceRemote.swift
```diff
@@ -100,6 +100,27 @@ open class SiteManagementServiceRemote: ServiceRemoteWordPressComREST {
...
     }
+    
+    /// Trigger a masterbar notification celebrating completion of mobile quick start.
+    ///
+    /// - Parameters:
+    ///   - siteID: The WordPress.com ID of the site.
+    ///   - success: Optional success block
+    ///   - failure: Optional failure block with NSError
+    ///
+    @objc open func markQuickStartAsCompleted(_ siteID: NSNumber, success: (() -> Void)?, failure: ((NSError) -> Void)?) {
```
nheagy: imo we should clarify _what_ part of Quick Start is being completed. Right now there are ways to complete each part of the quick start checklist, and this is completing the whole checklist itself. I'd prefer something like `markQuickStartChecklistAsComplete`, but if that's just to long then so be it.

danielebogo: Totally agree with you @nheagy !

PR #43, opened 2018-10-11
Support for custom domains.

jklausa:
This his mostly contents of #25, which was mostly reviewed already by me and @ScoutHarris before — just with some minor cleanups and with latest `develop` merged in.

I'm gonna leave it open after the review for a while, since it's possible I might need to do some changes while working on the WPiOS part of it, but it's close enough where I thought another pair of eyes wouldn't hurt :)


WordPressKit/JSONDecoderExtension.swift
```diff
@@ -0,0 +1,51 @@
...
+            default:
+                break
+            }
+            if let date = date {
+                return date
+            } else {
+                throw DecodingError.dataCorruptedError(
+                    in: container,
+                    debugDescription: "Cannot decode date string \(dateStr)"
+                )
```
danielebogo: Could the `if let` be `guard let date ...` and throw the error? So we can return the date after it without the `if else` statement.

danielebogo:
Hey @jklausa I only left 2 comments, but it looks great so far!

aerych:
Hey @jklausa 
I have nothing to add over @danielebogo's comments.  Once those are in and the conflicts resolved I think its good to go.

jklausa:
@aerych GitHub is having some weird hiccups lately — I've addressed those way back when, but it's not showing in the inline diffs — you have to go to "Files changed" up top to see it :)

jklausa:
@aerych can i consider your thumbs up reactji above as :shipit: :)?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

aerych:
Fo sho! :shipit: with much 🚢 ing

PR #45, opened 2018-10-18
Incremented podspec dependency on WordPressShared

stevebaranski:
With the update to `WordPressShared`, we also need to increment this `podspec` dependency.


ctarda:
Actually, @stevebaranski , the build failed locally with a cocoa pods error.

stevebaranski:
@nheagy thanks for bring this up. Hopefully our discussion via `p77Llu-b8E-p2` will prevent future recurrences.

PR #53, opened 2018-11-13
Blog: get all authors API

danielebogo:
This PR adds a new method to the `BlogRemoteService` which allows to load all the authors for a specific blog. This because the current method `- (void)getAuthorsWithSuccess:(UsersHandler)success failure:(void (^)(NSError *error))failure` loads only the default number (20) of authors without check if there are more.

The new method `- (void)getAllAuthorsWithNumber:(NSNumber *)number success:(UsersHandler)success failure:(void (^)(NSError *error))failure` recursively calls a private method passing the number of users returned and the offset.

This new API works with REST and XMLRPC services.

Using a SemVer spec I updated the new version to _1.5.0-beta.1_.

**To Test:**
- Simply run the Unit Test

**P.S.**
I also run a test on a WP iOS spike, testing with Wp.com blog, self-hosted jetpack and self-hosted non-jetpack blogs. Everything worked fine.

### DIFF
--- a/WordPressKit/BlogServiceRemote.h
+++ b/WordPressKit/BlogServiceRemote.h
@@ -14,13 +14,15 @@ typedef void (^SuccessHandler)(void);
 @protocol BlogServiceRemote <NSObject>
 
 /**
- *  @brief      Synchronizes a blog's authors.
- *
- *  @param      success     The block that will be executed on success.  Can be nil.
- *  @param      failure     The block that will be executed on failure.  Can be nil.
+ Synchronizes all blog's authors.
+
+ @param number Limit the total number of users returned.
+ @param success The block that will be executed on success.  Can be nil.
+ @param failure The block that will be executed on failure.  Can be nil.
  */
-- (void)getAuthorsWithSuccess:(UsersHandler)success
-                      failure:(void (^)(NSError *error))failure;
+- (void)getAllAuthorsWithNumber:(NSNumber *)number
+                        success:(UsersHandler)success
+                        failure:(void (^)(NSError *error))failure;
 
 /**
  *  @brief      Synchronizes a blog's post types.
--- a/WordPressKit/BlogServiceRemote.h
+++ b/WordPressKit/BlogServiceRemote.h
@@ -14,6 +14,15 @@ typedef void (^SuccessHandler)(void);
 @protocol BlogServiceRemote <NSObject>
 
 /**
+ *  @brief      Synchronizes a blog's authors.
+ *
+ *  @param      success     The block that will be executed on success.  Can be nil.
+ *  @param      failure     The block that will be executed on failure.  Can be nil.
+ */
+- (void)getAuthorsWithSuccess:(UsersHandler)success
+                      failure:(void (^)(NSError *error))failure;
+
+/**
  Synchronizes all blog's authors.
 
  @param number Limit the total number of users returned.
--- a/WordPressKit/BlogServiceRemote.h
+++ b/WordPressKit/BlogServiceRemote.h
@@ -14,24 +14,13 @@ typedef void (^SuccessHandler)(void);
 @protocol BlogServiceRemote <NSObject>
 
 /**
- *  @brief      Synchronizes a blog's authors.
- *
- *  @param      success     The block that will be executed on success.  Can be nil.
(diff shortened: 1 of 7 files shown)


WordPressKit/BlogServiceRemote.h
```diff
@@ -22,6 +22,17 @@ typedef void (^SuccessHandler)(void);
...
                       failure:(void (^)(NSError *error))failure;
 
+/**
+ Synchronizes all blog's authors.
+
+ @param number Limit the total number of users returned.
+ @param success The block that will be executed on success.  Can be nil.
+ @param failure The block that will be executed on failure.  Can be nil.
+ */
+- (void)getAllAuthorsWithNumber:(NSNumber *)number
```
aerych: Maybe its just me, but is it a little weird for a method named "get all authors" to then set a limit for the number returned?   Maybe just `getAuthorsWithNumber` ?

aerych: Would `NSInteger` rather than `NSNumber` be better in this case?

WordPressKit/BlogServiceRemoteREST.m
```diff
@@ -95,6 +95,62 @@ - (void)getAuthorsWithSuccess:(UsersHandler)success
...
+{
+    NSMutableArray *remoteUsers = [NSMutableArray array];
+    [self getAllAuthorsWithRemoteUsers:remoteUsers
+                                number:number
+                                offset:nil
+                               success:success
+                               failure:failure];
+}
+
+- (void)getAllAuthorsWithRemoteUsers:(NSMutableArray <RemoteUser *>*)remoteusers
```
aerych: Is there a reason we need to pass `remoteUsers` to this method rather than defining and instantiating in the method body?

WordPressKit/BlogServiceRemoteREST.m
```diff
@@ -95,6 +95,62 @@ - (void)getAuthorsWithSuccess:(UsersHandler)success
...
+                                number:(NSNumber *)number
+                                offset:(NSNumber *)offset
```
aerych: Maybe `NSIntegers` instead of `NSNumbers`?

WordPressKit/BlogServiceRemoteREST.m
aerych: Same comment about "all authors".  Maybe just 'getAuthors' since the number can be limited.

WordPressKit/BlogServiceRemote.h
aerych: I'm curious.  The `number` parameter is letting the caller define the batch size, but is there a need?  Would it make more sense to just have `getAllAuthorsWithSuccess:failures:` and define the batch size internally?  Just a thought.

WordPressKit/BlogServiceRemoteREST.m
```diff
@@ -95,6 +95,62 @@ - (void)getAuthorsWithSuccess:(UsersHandler)success
...
+                               success:(UsersHandler)success
+                               failure:(void (^)(NSError *error))failure
+{
```
aerych: Rather than need to define and pass `remoteUsers` in the calling method, you might consider checking for `remoteUsers==nil` in this method and instantiating a copy if one is needed.  It would keep the logic a bit more encapsulated.

WordPressKit/BlogServiceRemoteREST.m
```diff
@@ -95,6 +95,62 @@ - (void)getAuthorsWithSuccess:(UsersHandler)success
...
+                       parameters:parameters
+                          success:^(id responseObject, NSHTTPURLResponse *httpResponse) {
+                              if (success) {
+                                  NSArray *users = responseObject[@"users"];
+                                  
+                                  if (users == nil || users.count == 0) {
+                                      success(remoteusers);
+                                  } else {
+                                      [remoteusers addObjectsFromArray:[self usersFromJSONArray:users]];
+                                      [self getAllAuthorsWithRemoteUsers:remoteusers
```
aerych: Nice use of recursion :)

danielebogo:
@aerych thanks for the review! I fixed everything and test it again with a WP spike!

danielebogo:
Thanks @aerych !

PR #57, opened 2018-11-30
Fix/xmlrpc sticky post

danielebogo:
Refs. [#10370](https://github.com/wordpress-mobile/WordPress-iOS/issues/10370)

This change set the `sticky` value only if it exists. It has been changed with a NSNumber instead of a BOOL. It fixes a bug when an authors can not publish posts on self hosted sites.

**To Test:**
- Run the Unit tests


danielebogo:
Thanks @nheagy !

PR #62, opened 2018-12-21
Update podfile lock

aerych:
I neglected to run `pod install` after changing the podspec version.  This PR updates Podfile.lock so everything is nice and tidy.


aerych:
Thanks @jtreanor :)

PR #63, opened 2019-01-04
Exposed new WPCOM service method for enhanced site creation

stevebaranski:
### Description

Implements the service to fulfill the service implementation of [this issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/10711).

### Testing

1. Checkout the branch. Confirm that it builds & existing tests pass.

There no Step 2! Yet. 😄 Please allow me to explain.

### Explanation

There is an associated change to WPiOS noted in the aforementioned issue. That PR is forthcoming.

Moreover, the changes proposed to `sites/new` have not been implemented on the services side. This PR implements the specification agreed upon in `p9wMUP-bH-p2`. Any issues that arise will be identified & resolved during [integration testing](
https://github.com/wordpress-mobile/WordPress-iOS/issues/10712).

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "1.6.1-beta.1"
+  s.version       = "1.7.0-beta.1"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
   s.description   = <<-DESC
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "1.6.0-beta.1"
+  s.version       = "1.6.1-beta.1"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
   s.description   = <<-DESC
--- a/WordPressKitTests/SiteCreationRequestEncodingTests.swift
+++ b/WordPressKitTests/SiteCreationRequestEncodingTests.swift
@@ -84,7 +84,7 @@ class SiteCreationRequestEncodingTests: XCTestCase {
         XCTAssertNotNil(jsonDictionary["validate"])
     }
 
-    func testSiteCreationRequestEncoding_WorksWithoutValidation() {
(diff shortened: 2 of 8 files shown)


WordPressKitTests/SiteCreationRequestEncodingTests.swift
```diff
@@ -0,0 +1,206 @@
...
+        let jsonDictionary = serializedJSON as! [String : AnyObject]
+
+        // Then
+        XCTAssertNotNil(jsonDictionary["blog_name"])
+        XCTAssertNotNil(jsonDictionary["blog_title"])
+        XCTAssertNotNil(jsonDictionary["client_id"])
+        XCTAssertNotNil(jsonDictionary["client_secret"])
+        XCTAssertNotNil(jsonDictionary["public"])
+        XCTAssertNotNil(jsonDictionary["lang_id"])
+        XCTAssertNotNil(jsonDictionary["options"])
```
frosty: Should we verify the values of the json dictionary, to ensure they contain what we expect them to? Also, should we verify the contents of the `options` dictionary here, as it has a couple of sublevels?

WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "1.6.0-beta.1"
+  s.version       = "1.6.1-beta.1"
```
frosty: If we're following semantic versioning, should we roll to `1.7.0-beta.1` instead of `1.6.1` as this is a new feature, not a bug fix?

stevebaranski: This was tricky for me - it's a Swift implementation of an existing endpoint, with a few more parameters passed. After a second look, I agree, I will update this to `1.7.0-beta.1`.

WordPressKitTests/SiteCreationRequestEncodingTests.swift
```diff
@@ -0,0 +1,206 @@
...
+        XCTAssertNotNil(jsonDictionary["validate"])
+    }
+
+    func testSiteCreationRequestEncoding_WorksSansVertical() {
```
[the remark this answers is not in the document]
stevebaranski: I was using `Without` above (line `87`) for `false`, and `sans` as absent (e.g., Optional / `nil`). Naming is hard. 😂 

I'll rename the three methods in an additional commit.

stevebaranski:
@frosty I took the liberty to push a few commits that address each of the suggestions you raised. Would you mind taking another look?

PR #64, opened 2019-01-04
Exposed public initializer for SiteCreationRequest

stevebaranski:
### Description

Unfortunately during #63, I overlooked the need to designate the members & initializer of `SiteCreationRequest` as `public`. This PR corrects that oversight.

It should be noted that `1.7.0-beta.1` has not yet been tagged & released, so it should not be necessary to change `WordPressKit.podspec`.

### Testing

Checkout the branch. Confirm that it builds & existing tests pass.

### DIFF
--- a/WordPressKit/WordPressComServiceRemote+SiteCreation.swift
+++ b/WordPressKit/WordPressComServiceRemote+SiteCreation.swift
@@ -6,16 +6,39 @@ import Foundation
 /// This value type is intended to express a site creation request.
 ///
 public struct SiteCreationRequest: Encodable {
-    let segmentIdentifier: Int64
-    let verticalIdentifier: String?
-    let title: String
-    let tagline: String?
-    let siteURLString: String
-    let isPublic: Bool
-    let languageIdentifier: String
-    let shouldValidate: Bool
-    let clientIdentifier: String
-    let clientSecret: String
+    public let segmentIdentifier: Int64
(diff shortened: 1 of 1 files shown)


WordPressKit/WordPressComServiceRemote+SiteCreation.swift
```diff
@@ -6,16 +6,39 @@ import Foundation
...
+    public let title: String
+    public let tagline: String?
+    public let siteURLString: String
+    public let isPublic: Bool
+    public let languageIdentifier: String
+    public let shouldValidate: Bool
+    public let clientIdentifier: String
+    public let clientSecret: String
+
+    public init(segmentIdentifier: Int64,
```
jkmassel: Do we need a manual `init` here? Sorry if I missed something in here, its just looks like it's the same as the autogenerated one?

stevebaranski: I thought so too, but alas, when attempting to instantiate `SiteCreationRequest` in WPiOS, I encountered the following compiler error:
> 'SiteCreationRequest' initializer is inaccessible due to 'internal' protection level

jkmassel: Ohhhhhh right. I would've seen that from my unit test! In that case, :shipit:!

PR #71, opened 2019-01-22
[Site creation] Implementation of segments service

ctarda:
Implements part of [issue #10316 in WordPress-iOS](https://github.com/wordpress-mobile/WordPress-iOS/issues/10316)

In this PR:
- Added an extension to WordPressComServiceRemote, to implement the call to the Segments Service
- Added model objects to model the response and Result type.

To review:
- Check out the branch
- Make sure it builds and tests pass
- Check out the code.

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "1.7.0-beta.4"
+  s.version       = "1.7.0-beta.5"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
   s.description   = <<-DESC
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "1.7.1-beta.1"
+  s.version       = "1.8.0-beta.1"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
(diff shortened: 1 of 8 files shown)


WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "1.7.0"
```
stevebaranski: What motivated the decision to increment the patch version rather than just the beta?

WordPressKit/WordPressComServiceRemote+SiteSegments.swift
```diff
@@ -84,34 +84,16 @@ public enum SiteSegmentsResult {
     case failure(SiteSegmentsError)
 }
 
-public struct SiteSegmentsRequest: Encodable {
```
stevebaranski: Incidentally, I created [this issue](https://github.com/wordpress-mobile/WordPressKit-iOS/issues/72) to address the apparent duplication of the `locale` parameter.

WordPressKitTests/Mock Data/site-segments-single.json
```diff
@@ -0,0 +1,9 @@
+{
```
stevebaranski: It's unlikely that the service will return one result, but I wonder - would a payload for such a case include _just_ a dictionary, or would it be an array containing a single dictionary? 

The latter case evinces failures in `SiteSegmentsResponseDecodingTests`.

stevebaranski: On second thought, I don't think that concern impacts this test, as I see you're using it to decode a single `SiteSegment` instance.

Issue #78, opened 2019-01-26
v2 endpoints require modified locale key


stevebaranski:
### Expected behavior

As per `p77Llu-byF-p2`, v2 endpoints must pass `_locale` rather than `locale`. Prior versions should still use `locale`. 

_From the perspective of semantic versioning, implementing this may be a breaking change._

### Actual behavior

In the [default case](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/WordPressComRestApi.swift#L54) where `appendsPreferredLanguageLocale` is `true`, a localization identifier is [added to requests](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/WordPressComRestApi.swift#L433-L441) with a key of `locale`. This needs to change based on the endpoint version.

### Steps to reproduce the behavior

Inspect a request constructed via `WordPressKit` and scrutinize the query string. Note the value `locale` is  used for all endpoint versions.

##### WordPressKit Environment
All devices & versions of `WordPressKit`, Xcode & iOS.
 - WordPressKit Version [e.g. 22] : all versions
 - OS: [e.g. iOS 12.1.3 (16D40)] : all versions
 - Device: [e.g. iPhone XS] : all devices
 - Xcode Version: [e.g. Xcode 10.1 (10B61)] : all versions

##### Additional Context
N/A

stevebaranski:
### Context

I drafted a naive PR for this issue that has stimulated some interesting discussion about both `ServiceRemoteWordPressComREST` & `WordPressComRestApi`.

Quoting @SergioEstevao:
> I was considering the following alternative, instead of just one WordPressComRestAPI instance ( object not class) what about having two instances:
>
> - One for the Rest API v1.x
> - Another for the Rest API 2.x
>
> This way we could just set the locale parameter on the constructor of the API object, and we don't need to go around changing depending on the service. If the call is the REST API v2, we get the instance for v2 and make the request.
> This way in the future if need to do more changes like this we could even have a different class for API v2.

### Observations

I considered this input a bit further, and would like to continue the discussion here. For context, WPiOS services currently subclass `ServiceRemoteWordPressComREST` and initialize with an instance of `WordPressComRestApi`. 

At the time of this writing, the designated initializer takes a **single** instance of the WP API:
`-[ServiceRemoteWordPressComREST initWithWordPressComRestApi:]`

That class offers a single instance method:
[...]

PR #80, opened 2019-01-29
Modified locale handling to inspect request parameters

### LINKED ISSUE #72: Modify locale handling to inspect request parameters
### Description

The default behavior of `WordPressComRestApi` appends a `locale` parameter to requests. To do so, it inspects & modifies the path of a given request [if needed](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/WordPressComRestApi.swift#L433-L441).

If a caller were to specify `locale` in the request parameters, however, the specified value is not properly de-duplicated. For example, composing a `GET` request with the default behavior enabled, passing `locale : foo` in the request parameters evinces a request with the following URL : `https://public-api.wordpress.com/wpcom/v2/<path>?locale=en&locale=foo`.

### Potential Solution

The client should inspect the parameters associated with the request, and de-duplicate those if necessary. It might be worth using [`URLComponents`](https://developer.apple.com/documentation/foundation/urlcomponents) for this purpose.

stevebaranski:
### Description

Fixes #72, which observed that the existing framework behavior inspected a request's path for locale, and by default, appended it if missing. This change ensures that request parameters are also inspected for the presence of the parameter.

_From the standpoint of semantic versioning, I view this as a breaking change._

### Testing Details

- Checkout the branch and confirm that existing tests pass.
- Review the code and documentation for additional context.
- From [`WordPressAuthenticator`](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/), checkout the branch `try/wordpresskit-72-validation` and confirm that it compiles & tests pass.
- From [`WPiOS`](https://github.com/wordpress-mobile/WordPress-iOS/), checkout the branch `try/wordpresskit-72-validation` and confirm that it compiles & tests pass. Also launch the app and exercise requests to confirm that there are no regressions.

- [X] Please check here if your pull request includes additional test coverage.

### DIFF
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -338,7 +338,7 @@ open class WordPressComRestApi: NSObject {
         let existingLocaleQueryItems = existingQueryItems.filter { $0.name == localeKey }
         if let parameters = parameters, parameters[localeKey] == nil, existingLocaleQueryItems.isEmpty {
             let preferredLanguageIdentifier = WordPressComLanguageDatabase().deviceLanguage.slug
-            let localeQueryItem = URLQueryItem(name: WordPressComRestApi.localeKey, value: preferredLanguageIdentifier)
+            let localeQueryItem = URLQueryItem(name: localeKey, value: preferredLanguageIdentifier)
 
             existingQueryItems.append(localeQueryItem)
         }
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -43,7 +43,7 @@ open class WordPressComRestApi: NSObject {
     
     fileprivate let backgroundUploads: Bool
 
-    fileprivate static let localeKey = "locale"
+    static let localeKey = "locale"
 
     fileprivate let oAuthToken: String?
     fileprivate let userAgent: String?
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -299,8 +299,6 @@ open class WordPressComRestApi: NSObject {
         return "\(String(describing: oAuthToken)),\(String(describing: userAgent))".hashValue
     }
 
-    // AlamoFire our parameters via `SessionManager.request(_:method:parameters:encoding:headers:)`
-
     /// This method assembles a valid request URL for the specified path & parameters.
     /// The framework relies on a field (`appendsPreferredLanguageLocale`) to influence whether or not locale should be
     /// added to the path of requests. This approach did not consider request parameters.
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -145,7 +145,7 @@ open class WordPressComRestApi: NSObject {
                          success: @escaping SuccessResponseBlock,
                          failure: @escaping FailureReponseBlock) -> Progress? {
 
-        guard let URLString = buildRequestURLFor(path: urlString) else {
+        guard let URLString = buildRequestURLFor(path: urlString, parameters: parameters) else {
             let error = NSError(domain: String(describing: WordPressComRestApiError.self),
                                 code: WordPressComRestApiError.requestSerializationFailed.rawValue,
                                 userInfo: [NSLocalizedDescriptionKey: NSLocalizedString("Failed to serialize request to the REST API.", comment: "Error message to show when wrong URL format is used to access the REST API")])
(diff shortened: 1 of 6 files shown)


WordPressKit/WordPressComRestApi.swift
```diff
@@ -299,18 +299,54 @@ open class WordPressComRestApi: NSObject {
         return "\(String(describing: oAuthToken)),\(String(describing: userAgent))".hashValue
     }
 
-    fileprivate func buildRequestURLFor(path: String) -> String? {
-        let pathWithLocale = appendLocaleIfNeeded(path)
+    // AlamoFire our parameters via `SessionManager.request(_:method:parameters:encoding:headers:)`
```
frosty: Is this comment missing a word somewhere near the beginning?

WordPressKit/WordPressComRestApi.swift
```diff
@@ -299,18 +299,54 @@ open class WordPressComRestApi: NSObject {
...
+
+    /// This method assembles a valid request URL for the specified path & parameters.
+    /// The framework relies on a field (`appendsPreferredLanguageLocale`) to influence whether or not locale should be
+    /// added to the path of requests. This approach did not consider request parameters.
+    ///
+    /// This method now considers both the path and specified request parameters when performing the substitution.
+    /// It only accounts for the locale parameter. AlamoFire encodes other parameters via `SessionManager.request(_:method:parameters:encoding:headers:)`
```
frosty: Great explanation 👍

WordPressKit/WordPressComRestApi.swift
```diff
@@ -299,18 +299,54 @@ open class WordPressComRestApi: NSObject {
...
+            return urlComponents
+        }
+
+        var componentsWithLocale = urlComponents
+        var existingQueryItems = componentsWithLocale.queryItems ?? []
+
+        let existingLocaleQueryItems = existingQueryItems.filter { $0.name == localeKey }
+        if let parameters = parameters, parameters[localeKey] == nil, existingLocaleQueryItems.isEmpty {
+            let preferredLanguageIdentifier = WordPressComLanguageDatabase().deviceLanguage.slug
+            let localeQueryItem = URLQueryItem(name: WordPressComRestApi.localeKey, value: preferredLanguageIdentifier)
```
frosty: Should we use the passed in `localeKey` value here, not the constant directly?

If so, we could potentially add another test to cover this, passing a different `localeKey` to `buildRequestURLFor`.

SergioEstevao:
> From the standpoint of semantic versioning, I view this as a breaking change

If I'm correct the API for external users of the library didn't change, so we could live with an update only of the minor version number? 1.9?

WordPressKit/WordPressComRestApi.swift
```diff
@@ -421,25 +456,6 @@ extension WordPressComRestApi {
...
-
-    /// Append the user's preferred device locale as a query param to the URL path.
-    /// If the locale already exists the original path is returned.
-    ///
-    /// - Parameters:
-    ///     - path: A URL string. Can be an absolute or relative URL string.
-    ///
-    /// - Returns: The path with the locale appended, or the original path if it already had a locale param.
-    ///
-    @objc class public func pathByAppendingPreferredLanguageLocale(_ path: String) -> String {
```
SergioEstevao: Do we no longer need to use this method on ObjC in the main app?

WordPressKit/WordPressComRestApi.swift
stevebaranski: Good catch - I have pushed an additional commit that removes the stray comment.

WordPressKit/WordPressComRestApi.swift
stevebaranski: Good catch. I pushed a commit that sets the key to the value of the injected `localeKey`.

As for the test with different values, this is forthcoming via #78.

stevebaranski:
@frosty thanks for the comments. I have pushed two commits that address the items you pointed out. Regarding:

> I'm not sure if I missed something, but do these changes provide a way for a different localeKey to be provided to WordPressComRestApi

Not just yet. The majority of that work is slated for #78, but the change to `WordPressComRestApi.buildRequestURLFor(path:parameters:localeKey:)` paves the way for it. 

For that change, we will likely amend the signatures of these methods to inject the applicable locale key.

- `GET(_:parameters:success:failure:)`
- `POST(_:parameters:success:failure:)`
- `multipartPOST(_:parameters:fileParts:requestEnqueued:success:failure:)`

stevebaranski:
While tests pass locally, I should note that CircleCI is currently reporting a build error. I inquired about that Friday when others were observed, and was advised that the CircleCI transition is in progress for this repo, and CircleCI doesn’t have a config file for it just yet.

WordPressKit/WordPressComRestApi.swift
stevebaranski: I search for consumers of this in the `try/wordpresskit-72-validation` branch I created, and could not find any.

PR #82, opened 2019-01-31
Add support for locale key injection

### LINKED ISSUE #78: v2 endpoints require modified locale key

stevebaranski:
### Description

Fixes #78, which requires that `_locale` be passed as the key to v2 endpoints, while preserving the original `locale` key for prior versions.

This required special consideration, as `ServiceRemoteWordPressComREST` is responsible for endpoint versions, while `WordPressComRestApi` has responsibility for applying locale to requests.

### Testing Details

- Checkout the branch, confirm that it builds, and that both new & existing tests pass.
- Review the code and documentation for additional context.
- From [`WPiOS`](https://github.com/wordpress-mobile/WordPress-iOS/), checkout the branch `try/wpkit-locale-integration` and confirm that it compiles & tests pass. Also launch the app and exercise requests to confirm that there are no regressions.

- [X] Please check here if your pull request includes additional test coverage.


stevebaranski:
@frosty @SergioEstevao given that this work continues what began in #80, would you two be willing to take a look at this? I'd be especially interested to hear if you have any thoughts on make the change more "strict", or if you think the documentation of this change can be improved in some way.

I've also added @ctarda, as this is potentially relevant to the ongoing work with Site Creation.

Finally, the CircleCI error is due to the fact that this repository is not correctly configured at the moment.

SergioEstevao:
@stevebaranski While you approach in this PR looks ok, I think we are getting on a path that in the future may introduce too much complexity.

I was considering the following alternative, instead of just one WordPressComRestAPI instance ( object not class) what about having two instances:

 - One for the Rest API v1.x
 - Another for the Rest API 2.x

This way we could just set the locale parameter on the constructor of the API object, and we don't need to go around changing depending on the service. If the call is the REST API v2, we get the instance for v2 and make the request.

This way in the future if need to do more changes like this we could even have a different class for API v2.

stevebaranski:
I opted to continue our discussion in the root [issue](https://github.com/wordpress-mobile/WordPressKit-iOS/issues/78). I'd welcome your thoughts.

stevebaranski:
Closing this PR as discussed. I'll open a new one shortly.

PR #88, opened 2019-02-11
Stats Refresh: Insights fetching

jklausa:
### Description

This is a first step in rewrite of `WPStatsServiceRemote`, as a part of "Full Stats Refresh".

The old code was very complex, hard to read, obj-c, and did things that arguably don't belong in networking layer (localizing strings! date formatting! keeping track of which requests are currently in flight!). 

This is an attempt at simplifying it, specifically addressing fetching "Insights" for now, and with time-series data to follow shortly.

The core idea of this design is a simple protocol and generic function combo in `StatsServiceRemoteV2`:

```swift
public func getInsight<InsightType: InsightProtocol>(completion: @escaping ((InsightType?, Error?) -> Void))

public protocol InsightProtocol {
    static var queryProperties: [String: AnyObject] { get }
    static var pathComponent: String { get }

    init?(jsonDictionary: [String: AnyObject])
}
```

The protocol provides everything required to fetch a specific kind of insight — the path to where should the reqeust be sent to, optional query properties — and perhaps most importantly, a way to parse the response. Because of this, the code of the `StatsServiceRemoteV2` is itself fairly short and simple — most of the gnarly data parsing stuff is delegated to the specific classes conforming to `InsightProtocol`. 

This design also allows for a cleaner call-side implementation — all insights are fetched with a simple call to `getInsight(:_)`, differing only in the generic type being specialized:
```swift
api.getInsight { (lastPost: StatsLastPostInsight?, error) in
    dump("hello insight! \(lastPost)")
}

api.getInsight { (allTimesStats: StatsAllTimesInsight?, error) in
     dump("hello all time stats! \(allTimesStats)")       
}
```

#### Misc notes:
Some of the actual response parsing in the subclasses is not pretty, due to the weird shape of responses in some of the API endpoints. Not much I could do there :(

I also think I found a weird edge-case in `swiftc` with the generic design — one specific insight  (last post) requires two (mostly unrelated) network calls to compile, and therefore is implemented as a specialization of the generic `getInsight(:_)`.

However, if I then move the `StatsLastPostInsight` struct  definition outside of the same file where `StatsServiceRemoteV2` is defined, the project no longer builds, claiming that there's no such type as `StatsLastPostInsight` in the project! I asked fine folks from platform9 to investigate further if they have some spare cycles this week — if any of y'all have some ideas what might be going on here, I'm all ears!

### Testing Details

1. Make sure this PR builds
2. Take a corresponding WPiOS branch for a spin — https://github.com/wordpress-mobile/WordPress-iOS/pull/10994

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "2.0.0-beta.4"
+  s.version       = "2.0.0-beta.5"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
   s.description   = <<-DESC
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "2.0.0-beta.5"
+  s.version       = "2.1.0-beta.1"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
   s.description   = <<-DESC
--- a/WordPressKit/Insights/StatsAllTimesInsight.swift
+++ b/WordPressKit/Insights/StatsAllTimesInsight.swift
@@ -33,7 +33,7 @@ extension StatsAllTimesInsight: InsightProtocol {
     //MARK: -
     private static var dateFormatter: DateFormatter {
         let formatter = DateFormatter()
-        formatter.dateFormat = "yyyy-MM-dd";
+        formatter.dateFormat = "yyyy-MM-dd"
         return formatter
     }
 }
--- a/WordPressKit/Insights/StatsAllTimesInsight.swift
+++ b/WordPressKit/Insights/StatsAllTimesInsight.swift
@@ -10,10 +10,6 @@ public struct StatsAllTimesInsight {
 extension StatsAllTimesInsight: InsightProtocol {
 
     //MARK: - InsightProtocol Conformance
-    public static var queryProperties: [String : AnyObject] {
-        return [:]
-    }
-
     public init?(jsonDictionary: [String: AnyObject]) {
         guard
             let statsDict = jsonDictionary["stats"] as? [String: AnyObject],
--- a/WordPressKit/Insights/StatsAllTimesInsight.swift
+++ b/WordPressKit/Insights/StatsAllTimesInsight.swift
@@ -1,0 +1,43 @@
+public struct StatsAllTimesInsight {
+    public let postsCount: Int
+    public let viewsCount: Int
+    public let bestViewsDay: Date
+    public let visitorsCount: Int
+    public let bestViewsPerDayCount: Int
+}
+
+
+extension StatsAllTimesInsight: InsightProtocol {
+
+    //MARK: - InsightProtocol Conformance
+    public static var queryProperties: [String : AnyObject] {
+        return [:]
+    }
+
+    public init?(jsonDictionary: [String: AnyObject]) {
+        guard
+            let statsDict = jsonDictionary["stats"] as? [String: AnyObject],
+            let postsCount = statsDict["posts"] as? Int,
+            let viewsCount = statsDict["views"] as? Int,
+            let visitorsCount = statsDict["visitors"] as? Int,
+            let bestViewsPerDayCount = statsDict["views_best_day_total"] as? Int,
+            let bestViewsDayString = statsDict["views_best_day"] as? String,
+            let bestViewsDay = StatsAllTimesInsight.dateFormatter.date(from: bestViewsDayString)
+            else {
+                return nil
+        }
+
+        self.postsCount = postsCount
+        self.bestViewsPerDayCount = bestViewsPerDayCount
+        self.visitorsCount = visitorsCount
+        self.viewsCount = viewsCount
+        self.bestViewsDay = bestViewsDay
+    }
+
+    //MARK: -
+    private static var dateFormatter: DateFormatter {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyy-MM-dd";
+        return formatter
+    }
+}
--- a/WordPressKit/Insights/StatsAnnualAndMostPopularTimeInsight.swift
+++ b/WordPressKit/Insights/StatsAnnualAndMostPopularTimeInsight.swift
@@ -25,10 +25,6 @@ public struct StatsAnnualAndMostPopularTimeInsight {
 }
 
 extension StatsAnnualAndMostPopularTimeInsight: InsightProtocol {
-    public static var queryProperties: [String : AnyObject] {
-        return [:]
-    }
-
     public static var pathComponent: String {
         return "stats/insights"
     }
--- a/WordPressKit/Insights/StatsAnnualAndMostPopularTimeInsight.swift
+++ b/WordPressKit/Insights/StatsAnnualAndMostPopularTimeInsight.swift
@@ -1,0 +1,93 @@
+public struct StatsAnnualAndMostPopularTimeInsight {
+
+    /// - A `DateComponents` object with one field populated: `weekday`.
+    public let mostPopularDayOfWeek: DateComponents
+    public let mostPopularDayOfWeekPercentage: Int
+
+    /// - A `DateComponents` object with one field populated: `hour`.
+    public let mostPopularHour: DateComponents
+    public let mostPopularHourPercentage: Int
+
+    public let annualInsightsYear: Int
+
+    public let annualInsightsTotalPostsCount: Int
+    public let annualInsightsTotalWordsCount: Int
+    public let annualInsightsAverageWordsCount: Double
+
+    public let annualInsightsTotalLikesCount: Int
+    public let annualInsightsAverageLikesCount: Double
+
+    public let annualInsightsTotalCommentsCount: Int
+    public let annualInsightsAverageCommentsCount: Double
+
+    public let annualInsightsTotalImagesCount: Int
+    public let annualInsightsAverageImagesCount: Double
+}
+
+extension StatsAnnualAndMostPopularTimeInsight: InsightProtocol {
+    public static var queryProperties: [String : AnyObject] {
+        return [:]
+    }
+
+    public static var pathComponent: String {
+        return "stats/insights"
+    }
+
+    public init?(jsonDictionary: [String : AnyObject]) {
+        guard
+            let highestHour = jsonDictionary["highest_hour"] as? Int,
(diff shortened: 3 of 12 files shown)


WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "2.0.0-beta.4"
+  s.version       = "2.0.0-beta.5"
```
stevebaranski: Alas, through no fault of your own, I think we'll need to update the minor version & reset the beta counter (e.g., `2.1.0-beta.1`).

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -0,0 +1,193 @@
...
+
+}
+
+// This serves both as a way to get the query properties in a "nice" way,
+// but also as a way to narrow down the generic type in `getInisght:` method.
+public protocol InsightProtocol {
+    static var queryProperties: [String: AnyObject] { get }
+    static var pathComponent: String { get }
+
+    init?(jsonDictionary: [String: AnyObject])
```
stevebaranski: How would you feel about having your implementors conform to `Decodable`?

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -0,0 +1,193 @@
...
+                                }, failure:  { (error, _) in
+                                    completion(nil, error)
+                                }
+        )
+    }
+
+}
+
+// This serves both as a way to get the query properties in a "nice" way,
+// but also as a way to narrow down the generic type in `getInisght:` method.
```
stevebaranski: Small typo : `getInsight(completion:)` vs ``getInisght:`.

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -0,0 +1,193 @@
...
+public protocol InsightProtocol {
+    static var queryProperties: [String: AnyObject] { get }
```
stevebaranski: A fair number of these implementations append `as AnyObject` to conform to this type.

I wonder if it be worthwhile for this type to be `[String: String?]` instead. In my mind, this would offer the following benefits:

- It matches the semantics of [`URLQueryItem`](https://developer.apple.com/documentation/foundation/urlqueryitem), which would benefit us if we ever consider `URLSession` over `Alamofire`.
- It should be easily cast to `[String: AnyObject]` in/on `StatsServiceRemoteV2.swift:24`

Thoughts?

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -0,0 +1,193 @@
...
+        return "posts/"
+    }
+
+    public init?(jsonDictionary: [String: AnyObject]) {
+        fatalError("This shouldn't be ever called, instead init?(jsonDictionary:_ views:_) be called instead.")
+    }
+
+    //MARK: -
+
+    private static let dateFormatter: ISO8601DateFormatter = .init()
```
stevebaranski: ❤️ I have yet to use this since it was added to iOS.

WordPressKit/Insights/StatsAllTimesInsight.swift
```diff
@@ -0,0 +1,39 @@
...
+    public let bestViewsPerDayCount: Int
+}
+
+
+extension StatsAllTimesInsight: InsightProtocol {
+
+    //MARK: - InsightProtocol Conformance
+    public init?(jsonDictionary: [String: AnyObject]) {
+        guard
+            let statsDict = jsonDictionary["stats"] as? [String: AnyObject],
```
stevebaranski: I personally like this style for `guard/let` syntax.

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -0,0 +1,193 @@
...
+// This serves both as a way to get the query properties in a "nice" way,
+// but also as a way to narrow down the generic type in `getInisght:` method.
+public protocol InsightProtocol {
+    static var queryProperties: [String: AnyObject] { get }
+    static var pathComponent: String { get }
+
+    init?(jsonDictionary: [String: AnyObject])
+}
+
+extension InsightProtocol {
```
stevebaranski: Nice touch. :)

WordPressKit/Insights/StatsDotComFollowersInsight.swift
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        let followers = subscribers.compactMap { StatsFollower(jsonDictionary: $0) }
+
+        self.dotComFollowersCount = subscribersCount
+        self.topDotComFollowers = followers
+    }
+
+    //MARK: -
+    fileprivate static let dateFormatter: ISO8601DateFormatter = .init()
```
stevebaranski: Also consider:
```
private static let dateFormatter = ISO8601DateFormatter()
```

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -0,0 +1,193 @@
...
+    public static var queryProperties: [String: AnyObject] {
+        return [:]
+    }
+
+    public static var pathComponent: String {
+        return "stats/"
+    }
+}
+
+// For some god-forsaken reason Swift compiler freaks out if this is not declared _in this file_,
```
stevebaranski: The content of this comment is valid.

From what I can tell, this is the only type you refer to the static members of within this file. That is to say, none of the other eight `InsightProtocol`-conformers are causing problems because we don't attempt to access them. I tested one in a local branch to confirm.

As for the tone of this comment, I would respectfully request that we moderate it slightly. Instead of:
```
For some god-forsaken reason 
```
please consider:
```
At the moment, I am unable to explain why the
```

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -0,0 +1,193 @@
...
+                }
+
+                completion(insight, nil)
+            }
+        }, failure: {(error, _) in
+            completion(nil, error)
+        })
+    }
+
+    public func getPostViews(`for` postID: Int, completion: @escaping ((Int?, Error?) -> Void)) {
```
stevebaranski: Can this method be `private` for now?

WordPressKit/StatsServiceRemoteV2.swift
stevebaranski: As for the essence of the problem you observed here & rightly noted in the PR description, I tinkered with this a bit, and wanted to report my findings.

I noted that the problematic type (i.e., `StatsLastPostInsight`) is the one you explicitly reference the static members of. I encountered some difficulty, but ultimately, I wanted to share one approach for your consideration:

1. Create new file `StatsLastPostInsight.swift` and extract the struct & protocol conformance to it.
1. I created an extension of `StatsServiceRemoteV2` in `StatsLastPostInsight.swift`, and moved the "problem" methods there (i.e., `getInsight(completion:)`, `.getLastPostInsight(completion:)`, and `getPostViews(for:completion:)`.

Bad idea? Worst idea? 😂

WordPressKit/StatsServiceRemoteV2.swift
jklausa: I desperately wish I could just use Decodable for this, but the underlying layer of `WordPressComRestApi` already parses the response from `Data` into a JSON dictionary, and `Codable` doesn't have a dictionary `Decoder` :(

WordPressKit/StatsServiceRemoteV2.swift
jklausa: I actually meant to redact that comment before posting this PR, I'm embarassed that I forgot to do that. Thanks for pointing this out!

WordPressKit/StatsServiceRemoteV2.swift
jklausa: Thanks!

WordPressKit/StatsServiceRemoteV2.swift
jklausa: Good point!

WordPressKit/StatsServiceRemoteV2.swift
jklausa: It's a life-saver.

WordPressKit/Insights/StatsAllTimesInsight.swift
jklausa: :)

WordPressKit/StatsServiceRemoteV2.swift
jklausa: Thanks :)!

WordPressKit/Insights/StatsDotComFollowersInsight.swift
jklausa: Done!

WordPressKit/StatsServiceRemoteV2.swift
jklausa: Sure can!

WordPressKit/Insights/StatsAllTimesInsight.swift
```diff
@@ -0,0 +1,39 @@
...
+        self.bestViewsPerDayCount = bestViewsPerDayCount
+        self.visitorsCount = visitorsCount
+        self.viewsCount = viewsCount
+        self.bestViewsDay = bestViewsDay
+    }
+
+    //MARK: -
+    private static var dateFormatter: DateFormatter {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyy-MM-dd";
```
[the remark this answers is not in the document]
jklausa: dang, you caught me :P

WordPressKit/Insights/StatsAnnualAndMostPopularTimeInsight.swift
```diff
@@ -0,0 +1,89 @@
...
+
+        let mappedWeekday: ((Int) -> Int) = {
+            // iOS Calendar system is `1-based` and uses Sunday as the first day of the week.
+            // The data returned from WP.com is `0-based` and uses Monday as the first day of the week.
+            // This maps the WP.com data to iOS format.
+            if $0 == 6 {
+                return 0
+            }
+
+            return $0 + 2
```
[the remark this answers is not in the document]
jklausa: Good call!

jklausa:
@ScoutHarris @stevebaranski updated and ready for another look!

@stevebaranski re/ `StatsLastPostInsight` — I'd like to leave it as it is for now — I'll add an issue to the GH project in the WPiOS repo to come back to it and the end of the project. 

My hope is that Platform folks will be able to come up with something clever here — if not, your workaround suggestion sounds good to me.

ScoutHarris:
Thanks @jklausa ! You get a 👍 from me!

jklausa:
heyo let's get this party started then :P

PR #93, opened 2019-02-18
Update podspec author

aerych:
Updates the podspec author to be "WordPress" => "mobile@automattic.com" for consistency with our other pods.

Since there's no functionality changes we can probably skip adding a new tag or version for this one.


aerych:
Thanks Will!

PR #97, opened 2019-02-19
Stats: fetching clicks / referrer

jklausa:
### Description

Adds support for fetching clicks and referrer data.

They're so similar in shape I took the liberty of grouping them both in one PR

### Testing Details

* Verify the code builds
* Verify that the tests pass

### DIFF
--- a/WordPressKit/Time-based data/ClicksStatsType.swift
+++ b/WordPressKit/Time-based data/ClicksStatsType.swift
@@ -1,0 +1,71 @@
+public struct ClicksStatsType {
+    public let period: StatsPeriodUnit
+    public let periodEndDate: Date
+
+    public let totalClicksCount: Int
+    public let otherClicksCount: Int
+
+    public let clicks: [StatsClick]
+}
+
+public struct StatsClick {
+    public let title: String
+    public let clicksCount: Int
+    public let clickedURL: URL?
+    public let iconURL: URL?
+
+    public let children: [StatsClick]
+}
+
+extension ClicksStatsType: TimeStatsProtocol {
+    public static var pathComponent: String {
+        return "stats/clicks"
+    }
+
+    public init?(date: Date, period: StatsPeriodUnit, jsonDictionary: [String : AnyObject]) {
+        guard
+            let unwrappedDays = type(of: self).unwrapDaysDictionary(jsonDictionary: jsonDictionary),
+            let totalClicks = unwrappedDays["total_clicks"] as? Int,
+            let otherClicks = unwrappedDays["other_clicks"] as? Int,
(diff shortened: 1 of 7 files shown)


WordPressKit/Time-based data/ClicksStatsType.swift
```diff
@@ -0,0 +1,71 @@
...
+        } else {
+            children = []
+        }
+
+        let icon = jsonDictionary["icon"] as? String
+        let urlString = jsonDictionary["url"] as? String
+
+        self.title = title
+        self.clicksCount = clicksCount
+        self.clickedURL = urlString.flatMap { URL(string: $0) }
```
stevebaranski: I recognize why we'd use flatMap on a collection. It appears that the benefit in this case is to handle the optionality of `icon` & `urlString`, correct?

WordPressKit/Time-based data/ReferrerStatsType.swift
```diff
@@ -0,0 +1,84 @@
+public struct ReferrerStatsType {
+    public let period: StatsPeriodUnit
+    public let periodEndDate: Date
+
+    public let totalReferrerViews: Int
```
stevebaranski: In `ClicksStatsType`, there appears to be an analogous property called `totalClicksCount`. Is this worth renaming to `totalViewsCount` or `totalRefererredViewsCount` for consistency with that and `StatsReferrer` (`viewsCount`) below?

WordPressKit/Time-based data/ReferrerStatsType.swift
```diff
@@ -0,0 +1,84 @@
...
+    public let otherReferrerViews: Int
```
stevebaranski: If we embrace the suggestion above, it is also applicable here.

stevebaranski:
@jklausa I just noted that the hash in `Podfile.lock` was updated when I ran `pod install` to review this branch. If you find that's the case for you, would you please include a commit for that too?

WordPressKit/Time-based data/ClicksStatsType.swift
jklausa: Exactly! It saves us doing something like: 
```swift
let clickedURL: URL?
if let urlString = urlString {
    clickedURL = URL(string: urlString)
} else {
    clickedURL = nil
}
```

Just a bit of syntactic sugar.

WordPressKit/Time-based data/ReferrerStatsType.swift
jklausa: Yup, absolutely! Thanks for catching that!

PR #98, opened 2019-02-20
Stats: Add support for fetching top posts

jklausa:
### Description

Last set of those! I'll be posting few more PRs today, but that should be it.

Let's start with top posts :)!

### Testing Details

* Verify the library builds
* Verify the tests pass

### DIFF
--- a/WordPressKit/Time-based data/AuthorsStatsType.swift
+++ b/WordPressKit/Time-based data/AuthorsStatsType.swift
@@ -14,10 +14,21 @@ public struct StatsTopAuthor {
 
 
 public struct StatsTopPost {
+
+    public enum Kind {
+        case unknown
+        case post
+        case page
+        case homepage
+    }
+
+
     public let title: String
     public let postID: Int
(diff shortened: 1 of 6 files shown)


WordPressKit/Time-based data/AuthorsStatsType.swift
```diff
@@ -83,6 +94,20 @@ extension StatsTopPost {
         self.postID = postID
         self.viewsCount = viewsCount
         self.postURL = URL(string: postURL)
+        self.kind = type(of: self).kind(from: jsonDictionary["type"] as? String)
+    }
+
+    static func kind(from kindString: String?) -> Kind {
+        switch kindString {
+        case "post"?:
```
ScoutHarris: Are these `?` necessary? I removed them and the tests passed. Seems like they're not needed...

jklausa: Huh, thanks for that — I would have sworn I tried before and it failed to build before!

Issue #79, opened 2019-01-26
Modify WordPressKit for Swift 5 support


stevebaranski:
### Expected behavior

We should adapt the framework to support Swift 5.

### Actual behavior

At the moment the framework uses Swift 4.2.

### Steps to reproduce the behavior

For all targets in the project, `SWIFT_VERSION` is currently set to 4.2.

##### WordPressKit Environment
 - WordPressKit Version : `WordPressKit` v1.8.0 & before
 - OS : iOS 12.x
 - Device : all devices
 - Xcode Version : Xcode 10.2 beta

##### Additional Context
[Swift 5 Release Notes for Xcode 10.2 beta](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_2_beta_release_notes/swift_5_release_notes_for_xcode_10_2_beta)

PR #105, opened 2019-02-25
Apply some Swift 4 cleanup

stevebaranski:
### Description

Refs #79. Before we create a branch to consider the transition to Swift 5, there is some low-hanging fruit related to Swift 4 that we could stand to tackle. This PR takes care of those:

- `SWIFT_VERSION` is set to 4.2 at both the project & target levels.
- Renamed some `fileprivate` members as `private` where applicable

### Testing Details

- Checkout the branch, confirm that it builds and existing tests pass.
- Evaluate the `issue/wpkit-79-integration` branch from WPiOS and confirm that it has no ill-effect.

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "2.1.0"
+  s.version       = "2.1.1-beta.1"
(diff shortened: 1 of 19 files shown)


stevebaranski:
@jklausa would you be willing to take a look at this?

stevebaranski:
Thanks! I found these by searching; I'm not aware of a compiler-assisted way to do this. This would be a bit more daunting task for a larger project (e.g., WPiOS).

PR #116, opened 2019-03-06
Fix of numerical tag sync on WP.com

SergioEstevao:
### Description

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/10735

This PR updates the sending of the tags to the server to use a string instead of an array. The reason behind this is that the API endpoint was struggling when numbers are being sent on the array object.

### Testing Details

This fix can be tested by using the main app on the branch for this PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/11196


SergioEstevao:
@mindgraffiti I updated the code to use the same version of the endpoint as our Android colleagues and it now looks it can support the numerical tags correctly. Do you mind to review it again?

SergioEstevao:
Can you please try to do a rm -Rf Pods and then clean build and rebuild the
project. Just to make sure you got the correct version of WordPressKit.

mindgraffiti:
Tested with latest changes. This 💯% fixes the problem! Thank you Sergio 😀

PR #120, opened 2019-03-12
Adds invalidTokenHandler

koke:
### Description

Exposes a way to 

For https://github.com/wordpress-mobile/WordPress-iOS/pull/11245

**Note:** I created a `release/2.1.0` branch just for this since we'll want to do a hotfix, we should make sure this gets merged to develop afterwards.

### Testing Details

See https://github.com/wordpress-mobile/WordPress-iOS/pull/11245

### DIFF
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -60,6 +60,8 @@ open class WordPressComRestApi: NSObject {
 
     private let localeKey: String
 
+    private var invalidTokenHandler: (() -> Void)? = nil
+
     /**
      Configure whether or not the user's preferred language locale should be appended. Defaults to true.
      */
@@ -152,6 +154,14 @@ open class WordPressComRestApi: NSObject {
     @objc open func invalidateAndCancelTasks() {
         sessionManager.session.invalidateAndCancel()
         uploadSessionManager.session.invalidateAndCancel()
+    }
+
+    public func authTokenEquals(token: String) -> Bool {
+        return token == oAuthToken
+    }
+
+    @objc func setInvalidTokenHandler(_ handler: @escaping () -> Void) {
+        invalidTokenHandler = handler
     }
 
     // MARK: Network requests
@@ -443,6 +453,9 @@ extension WordPressComRestApi {
         ]
 
         let mappedError = errorsMap[errorCode] ?? WordPressComRestApiError.unknown
+        if mappedError == .invalidToken {
+            invalidTokenHandler?()
+        }
         userInfo[WordPressComRestApi.ErrorKeyErrorCode] = errorCode
         userInfo[WordPressComRestApi.ErrorKeyErrorMessage] = errorDescription
         userInfo[NSLocalizedDescriptionKey] =  errorDescription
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
This is an excellent addition.

WordPressKit/WordPressComRestApi.swift
```diff
@@ -154,6 +156,14 @@ open class WordPressComRestApi: NSObject {
         uploadSessionManager.session.invalidateAndCancel()
     }
 
+    public func authTokenEquals(token: String) -> Bool {
+        return token == oAuthToken
```
nheagy: I found this line a bit confusing because it's hard to tell which of `token` and `oAuthToken` is the one being considered.

Method signature like `authTokenEquals(token candidateToken: String)` might improve that? I dunno. What do you think?

nheagy (approved):
This looks good. I had one suggestion but :shipit: whenever you feel it's ready.

WordPressKit/WordPressComRestApi.swift
```diff
@@ -443,6 +453,9 @@ extension WordPressComRestApi {
         ]
 
         let mappedError = errorsMap[errorCode] ?? WordPressComRestApiError.unknown
+        if mappedError == .invalidToken {
+            invalidTokenHandler?()
```
stevebaranski: I'm seeing that this path is properly exercised via `WordPressComRestApiTests.testInvalidTokenFailedCall()`.

WordPressKit/WordPressComRestApi.swift
```diff
@@ -154,6 +156,14 @@ open class WordPressComRestApi: NSObject {
         uploadSessionManager.session.invalidateAndCancel()
     }
 
+    public func authTokenEquals(token: String) -> Bool {
+        return token == oAuthToken
+    }
+
+    @objc func setInvalidTokenHandler(_ handler: @escaping () -> Void) {
```
stevebaranski: While this might benefit from a `typealias` to better characterize the intent of the completion handler, this works fine as-is.

WordPressKit/WordPressComRestApi.swift
stevebaranski: I agree with @nheagy's observation above.

I should note, however, that in my testing, I did not see this method invoked in any code paths. For a hot fix, I'd generally advocate a "less is more" approach, but the changes do not discernably impact the behavior.

ghost (approved):
I left a couple of comments, but the changes here work as intended. 

I took a little bit of time to verify the contents of both the base & target branches : https://github.com/wordpress-mobile/WordPressKit-iOS/compare/master...release/2.1.0

Thanks for tackling this issue!

WordPressKit/WordPressComRestApi.swift
koke: Ah yes, that's unused now. Originally I used a NSNotification to handle the invalid token, and this helped checked which token had been invalidated (without exposing the token). Now that it uses a callback, WPAccount can capture `self` when setting the callback and make sure the right account gets cleared. I'll delete this

PR #124, opened 2019-03-18
[Jetpack Remote Install] New Blog service API

danielebogo:
### Description

Refs. [#10277](https://github.com/wordpress-mobile/WordPress-iOS/issues/10277)

This PR introduces the new API used to download/install/activate the _Jetpack_ plugin in a self-hosted site.

I created a new _BlogServiceRemoteREST_ extension adding the new API and the errors enum _JetpackInstallError_.
The expected API success response is:
```
{
    "status": true
}
```
The expected failure response is:
```
{
    "error": "ERROR_NAME",
    "message": "MESSAGE"
}
```

### Testing Details

The new API is fully covered by Unit Tests. Simply run `BlogServiceRemoteRESTTests+Jetpack`. It contains the Unit Tests to cover all these cases:
- Success -> `status: true`
- Failure -> `status: false`
- Error -> Every `JetpackInstallError` has been covered
- Unknown Error -> If an error is not a `JetpackInstallError` case.


WordPressKit/BlogServiceRemoteREST+Jetpack.swift
```diff
@@ -0,0 +1,55 @@
...
+    case installResponseError = "INSTALL_RESPONSE_ERROR"
+    case loginFailure = "LOGIN_FAILURE"
+    case siteIsJetpack = "SITE_IS_JETPACK"
+    case activationOnInstallFailure = "ACTIVATION_ON_INSTALL_FAILURE"
+    case activationResponseError = "ACTIVATION_RESPONSE_ERROR"
+    case activationFailure = "ACTIVATION_FAILURE"
+    case unknown
+
+    init?(error code: String?) {
+        guard let code = code else {
```
aerych: If you wanted to get rid of the `guard` you could add a default value to `code` like so:
`init?(error code: String = JetpackInstallError.unknown.rawValue)`

WordPressKit/BlogServiceRemoteREST+Jetpack.swift
```diff
@@ -0,0 +1,55 @@
...
+public extension BlogServiceRemoteREST {
+    public func installJetpack(url: String,
+                               username: String,
+                               password: String,
+                               completion: @escaping (Bool, JetpackInstallError?) -> Void) {
+        guard let escapedURL = url.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed) else {
+            completion(false, .unknown)
+            return
+        }
+        let path = String(format: "jetpack-install/%@/?locale=en_US", escapedURL)
```
aerych: `?locale=en_US` I _think_ this is unnecessary (and we probably don't want it hard-coded to en_US regardless). @stevebaranski recently updated things so the locale is automatically added and set to the users preferred device language (iirc).

WordPressKitTests/BlogServiceRemoteRESTTests+Jetpack.swift
```diff
@@ -0,0 +1,197 @@
...
+    let jetpackRemoteErrorForbiddenMockFilename = "blog-service-jetpack-remote-error-forbidden.json"
+    let jetpackRemoteErrorInstallFailureMockFilename = "blog-service-jetpack-remote-error-install-failure.json"
+    let jetpackRemoteErrorInstallResponseMockFilename = "blog-service-jetpack-remote-error-install-response.json"
+    let jetpackRemoteErrorLoginFailureMockFilename = "blog-service-jetpack-remote-error-login-failure.json"
+    let jetpackRemoteErrorSiteIsJetpackMockFilename = "blog-service-jetpack-remote-error-site-is-jetpack.json"
+    let jetpackRemoteErrorActivationInstallMockFilename = "blog-service-jetpack-remote-error-activation-install.json"
+    let jetpackRemoteErrorActivationResponseMockFilename = "blog-service-jetpack-remote-error-activation-response.json"
+    let jetpackRemoteErrorActivationFailureMockFilename = "blog-service-jetpack-remote-error-activation-failure.json"
+    
+    var endpoint: String { return "jetpack-install/\(encodedURL)/?locale=en_US" }
```
aerych: `?locale=en_US`
Just a note to remove locale here as well if its indeed unnecessary.

WordPressKit/BlogServiceRemoteREST+Jetpack.swift
```diff
@@ -0,0 +1,55 @@
...
+    case invalidCredentials = "INVALID_CREDENTIALS"
+    case forbidden = "FORBIDDEN"
+    case installFailure = "INSTALL_FAILURE"
+    case installResponseError = "INSTALL_RESPONSE_ERROR"
+    case loginFailure = "LOGIN_FAILURE"
+    case siteIsJetpack = "SITE_IS_JETPACK"
+    case activationOnInstallFailure = "ACTIVATION_ON_INSTALL_FAILURE"
+    case activationResponseError = "ACTIVATION_RESPONSE_ERROR"
+    case activationFailure = "ACTIVATION_FAILURE"
+    case unknown
```
aerych: I saw the tests mention `UNKNOWN_ERROR`.  Should that be the string used here or was that just for convenience?

WordPressKit/BlogServiceRemoteREST+Jetpack.swift
stevebaranski: Thanks for the ping @aerych!

Yes @danielebogo, if `appendsPreferredLanguageLocale` is `true` on the underlying `WordPressComRestApi` instance, this is indeed "free" behavior. Most of the changes I made to [this method](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/WordPressComRestApi.swift#L353) were focused on preventing duplicate parameter entries that were possible with the logic in-place.

Please let me know if you have questions or if you need additional information.

danielebogo: Thanks @aerych and @stevebaranski for the hint 😊

WordPressKit/BlogServiceRemoteREST+Jetpack.swift
danielebogo: Yes It's just for convenience. Only to test the default fallback to the `unknown` case.

WordPressKit/BlogServiceRemoteREST+Jetpack.swift
danielebogo: At the end I removed the guard without using the default value. Since the `init(rawValue:)` returns an optional I can use the nil-coalescing  to use `.unknown` as fallback.

danielebogo:
Thanks @aerych ! I fixed the locale and modified a bit the enum init. It's ready for another look!

danielebogo:
Thanks @aerych !

PR #125, opened 2019-03-19
Provide memberwise initializers for Stats Insights objects.

jklausa:
### Description

As part of Stats work in WPiOS, we need a way to map Stats object back and forth from Core Data to the format WPKit uses. 

Due to the fact that WPKit is in a separate module from WPiOS "proper", we need to manually provide memberwise initialisers for those objects. This accomplishes that!

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "3.1.2-beta.1"
+  s.version       = "3.1.2"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
   s.description   = <<-DESC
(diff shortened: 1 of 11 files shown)


WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "3.1.0"
+  s.version       = "3.1.2"
```
aerych: I think this would need to be `3.2.0-beta.2` assuming #124 lands first. 

🤔 but now that I think about it, you can @danielebogo could coordinate to roll all the changes into `3.2.0-beta.1` if you were so inclined.

danielebogo: @aerych I already merged my PR but I didn't create a tag yet. @jklausa let me know if it's ok for you use `3.2.0-beta.1` with also your PR merged

jklausa: I'll update mine to point to 3.2.0-beta.1 too :)

Issue #136, opened 2019-03-28
Update domains/suggestions endpoint to use new segment_id parameter


kwonye:
We have a new domains suggestion parameter which replaces all other parameters. Now, we simply pass `segment_id=INT` and if it matches a blog type, we'll get .blog subdomains back (coffee.shop.blog) along with the normal wordpress.com subdomains (coffee.wordpress.com)

If it doesn't match a blog type, we'll simply get wordpress.com subdomains back.

PR #137, opened 2019-03-28
Adding new remote call for domain suggestions

kwonye:
Addresses #136

There will be a WP-iOS PR using this soon.


WordPressKit/DomainsServiceRemote.swift
```diff
@@ -156,6 +156,32 @@ public class DomainsServiceRemote: ServiceRemoteWordPressComREST {
         }
     }
 
+    public func getDomainSuggestions(base query: String,
```
stevebaranski: @diegoreymendez for the site creation calls, we tried to avoid "parameter bloat" by employing a value object to encapsulate the attributes of a request when more than one could be found.

For example:

- [`SiteVerticalsRequest`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/f1940b9c8a07688077138e3b6fb2e3ac9f848902/WordPressKit/WordPressComServiceRemote%2BSiteVerticals.swift#L9)
- [`SiteCreationRequest`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/1490c59130e8f4dd5ca3d2633f72886a86d57070/WordPressKit/WordPressComServiceRemote%2BSiteCreation.swift#L8)

Now that we're adding a second parameter, would you consider following that pattern here for consistency?

diegoreymendez:
@ctarda - Can I ask you to validate if the changes requested by @stevebaranski in his review have been applied?

PR #139, opened 2019-03-29
Podfile: Reference 'WordPressKit' as a development pod

jtreanor:
### Description

This reverts the changes from https://github.com/wordpress-mobile/WordPressKit-iOS/pull/99 so that we once again use the recommended way of integrating pods being developed.

This will catch any issues with the podspec in CI and the Xcode project more closely resembles how WordPressKit is integrated into other projects. It also removes duplication of dependencies.

### Testing Details

- Run `bundle exec pod install` and run the tests.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -6,23 +6,23 @@ use_frameworks!
 platform :ios, '10.0'
 plugin 'cocoapods-repo-update'
 
+def wordpress_kit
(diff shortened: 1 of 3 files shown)


jtreanor:
I'm actually going to close this for now because there seems to be something strange about how the framework is linked. I think we should do this, but fully understand it first.

PR #141, opened 2019-04-03
Make WordPress.com base urls configurable

jtreanor:
### Description

Currently, `https://public-api.wordpress.com` and similar urls are hardcoded in a number of places around WordPressKit. This makes it very difficult to mock out network requests for testing purposes without resorting to Swizzling or similar hacks.

This is my proposal to make all of these urls configurable, while still defaulting to the current values.

I would love to hear any feedback people have and I'm happy to rethink the approach if there is a better alternative.

### Testing Details

- All tests pass and the podspec validates.
- Check out my corresponding WPiOS branch (`try/configure-base-urls`) and do a smoke test.


WordPressKit/WPStatsServiceRemote.m
```diff
@@ -34,19 +34,20 @@ @interface WPStatsServiceRemote ()
...
+    NSParameterAssert(baseUrlString != nil);
     
     self = [super init];
     if (self) {
         _oauth2Token = oauth2Token;
         _siteId = siteId;
         _siteTimeZone = timeZone;
-        _sitesPathPrefix = [NSString stringWithFormat:@"%@/sites/%@", WordPressComApiClientEndpointURL, _siteId];
-        _statsPathPrefix = [NSString stringWithFormat:@"%@/sites/%@/stats", WordPressComApiClientEndpointURL, _siteId];
+        _sitesPathPrefix = [NSString stringWithFormat:@"%@/rest/v1.1/sites/%@", baseUrlString, _siteId];
```
yaelirub: I am just getting myself familiar with the way we do API versioning so Im serious, is it possible to pass the version (as enum or something)?

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -24,15 +24,18 @@ public final class WordPressComOAuthClient: NSObject {
...
-        case socialLogin2FA = "https://wordpress.com/wp-login.php?action=two-step-authentication-endpoint&version=1.0"
-        case socialLoginNewSMS2FA = "https://wordpress.com/wp-login.php?action=send-sms-code-endpoint"
-        case oAuthRedirect = "https://wordpress.com/"
-
-        var url: URL {
-            return URL(string: self.rawValue)!
+        case oAuthBase = "/oauth2/token"
+        case socialLogin = "/wp-login.php?action=social-login-endpoint&version=1.0"
+        case socialLogin2FA = "/wp-login.php?action=two-step-authentication-endpoint&version=1.0"
+        case socialLoginNewSMS2FA = "/wp-login.php?action=send-sms-code-endpoint"
```
yaelirub: Just neat picking here but can we remove the "New"? and just have socialLoginSMS2FA? It made me wonder if there is something like socialLoginOldSMS2FA.

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -24,15 +24,18 @@ public final class WordPressComOAuthClient: NSObject {
...
+        case oAuthRedirect = "/"
```
yaelirub: where are we using this? Couldn't find anything in the code

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -66,23 +72,40 @@ public final class WordPressComOAuthClient: NSObject {
...
-    /// Creates a WordPresComOAuthClient initialized with the clientID and secret constants defined in the
-    /// ApiCredentials singleton
+    /// Creates a WordPresComOAuthClient initialized with the clientID and secrets provided
     ///
     @objc public class func client(clientID: String, secret: String) -> WordPressComOAuthClient {
         let client = WordPressComOAuthClient(clientID: clientID, secret: secret)
         return client
     }
 
+    /// Creates a WordPresComOAuthClient initialized with the clientID, secretand base urls provided
```
yaelirub: Theres a space missing in "secret and"

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -66,23 +72,40 @@ public final class WordPressComOAuthClient: NSObject {
...
+                                   wordPressComBaseUrl: String,
+                                   wordPressComApiBaseUrl: String) -> WordPressComOAuthClient {
+        let client = WordPressComOAuthClient(clientID: clientID,
+                                             secret: secret,
+                                             wordPressComBaseUrl: wordPressComBaseUrl,
+                                             wordPressComApiBaseUrl: wordPressComApiBaseUrl)
+        return client
+    }
+
     /// Creates a WordPressComOAuthClient using the defined clientID and secret
```
yaelirub: Should this comment be updated too to include the new parameters?

WordPressKit/WordPressComRestApi.swift
```diff
@@ -44,8 +44,8 @@ open class WordPressComRestApi: NSObject {
     public typealias SuccessResponseBlock = (_ responseObject: AnyObject, _ httpResponse: HTTPURLResponse?) -> ()
     public typealias FailureReponseBlock = (_ error: NSError, _ httpResponse: HTTPURLResponse?) -> ()
 
-    @objc public static let apiBaseURLString: String = "https://public-api.wordpress.com/"
-    
+    @objc public static let BaseURLStringDefault = "https://public-api.wordpress.com"
```
yaelirub: should this be with a lower case b? "baseURLStringDefault"?

jtreanor: I'll admit I'm not super familiar with convention here. This matches the other static constants here like `LocaleKeyDefault`, `LocaleKeyV2` and `SessionTaskKey`.

Happy to change to `baseURLStringDefault` if that is more correct though.

WordPressKit/WordPressComOAuthClient.swift
jtreanor: Good point! I hadn't spotted that this wasn't used.

WordPressKit/WordPressComOAuthClient.swift
jtreanor: That makes more sense to me too but I'm lacking the context on why it is this way so I'm hesitant to change it. There is also a `socialNewSMS2FASessionManager`. Perhaps this is something that should be refactored in another change?

WordPressKit/WPStatsServiceRemote.m
jtreanor: So in almost every other case a subclass of `ServiceRemoteWordPressComREST` is used which has a method, `pathForEndpoint:withVersion:` to do exactly this.

I'm not sure why `WPStatsServiceRemote` is an exception and does not subclass this. I have made a small update to match more closely what is there now (the oath is in a constant).

jtreanor:
Thanks for the review @yaelirub!

WordPressKit/WordPressComRestApi.swift
yaelirub: Yeah its a mix there, ah? defaultBackgroundSessionIdentifier is lower case and so was apiBaseURLString. To me lower case makes more sense.

SergioEstevao: Lower case is the standard on Swift.

stevebaranski: 👋 Hello there. I agree that the case conventions here are consistent; the majority of them are capitalized. Should you decide to make this change, please keep in mind the visibility of these properties (i.e., `public`), and the subsequent semantic version impact of making them lowercase.

Regarding naming conventions, at the moment our [Swift Style Guide](https://github.com/wordpress-mobile/swift-style-guide) does not express an opinion, but I tend to prefer "Pascal case" for types & protocols, and "lower camel case" for everything else. If we proceed with the change you're proposing, perhaps we could also amend the style guide to reflect this new precedent? We could then enable [this](https://github.com/realm/SwiftLint/blob/master/Rules.md#type-name) SwiftLint rule.

jtreanor: Thanks for the input folks. I’ll go with the lower camel case as suggested. My day to day exposure to Swift is pretty low so I’m happy to lean on your feedback.

I’m aware of the versioning implications here. As is, this is clearly a breaking change. I could make the API changes additive only by keeping this constant as `apiBaseUrlString` but that naming is a little misleading since it’s now just the default url.

jtreanor: Although maybe `apiBaseUrlString` isn’t such a bad name 🤔 It is still a constant with the real WP.com endpoint.

jtreanor:
Thanks, @bummytime. I've renamed the constant back to `apiBaseUrlString` as suggested so these changes are now additive only.

PR #150, opened 2019-04-29
Issues/non localized plans for all sites

aerych:
This PR adds a new method to fetch simple plan descriptions for the current user's sites.

We call the `me/sites` to fetch the list of the user's sites.  The call passes the `fields=siteID,plan` parameter to limit the response to just the desired information.  We also specify the `locale` in order to override WordPressKit's default locale behavior. For our use case in WPiOS, we want to specify an `en` locale in order to return plan descriptions in English on Zendesk tickets. 

To test:
Run the new tests.  Confirm that they pass. 

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.1.0-beta.2"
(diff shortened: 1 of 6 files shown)


aerych:
Thanks @ScoutHarris!  I needed to bump the podspec version again since another PR claimed beta.2.  I'll merge as soon as CI is happy :)

PR #156, opened 2019-05-06
Implemented savePost (calling autoSave endpoint)

yaelirub:
In order to preview published posts with their current updates, we need to call the autoSave endpoint.
This PR adds that call.
The successful response to this call contains a previewURL which we returning in the success block so that the caller can use it to preview correctly.

Addresses: wordpress-mobile/WordPress-iOS#9287

This branch is used in this WPiOS PR

### DIFF
--- a/WordPressKit/PostServiceRemoteREST.m
+++ b/WordPressKit/PostServiceRemoteREST.m
@@ -183,7 +183,7 @@ static NSString * const RemoteOptionValueOrderByPostID = @"ID";
            }];
 }
 
-- (void)savePost:(RemotePost *)post
+- (void)autoSave:(RemotePost *)post
          success:(void (^)(RemotePost *, NSString *))success
          failure:(void (^)(NSError *))failure
 {
--- a/WordPressKit/PostServiceRemoteREST.m
+++ b/WordPressKit/PostServiceRemoteREST.m
@@ -183,6 +183,33 @@ static NSString * const RemoteOptionValueOrderByPostID = @"ID";
            }];
 }
 
+- (void)savePost:(RemotePost *)post
+         success:(void (^)(RemotePost *, NSString *))success
+         failure:(void (^)(NSError *))failure
+{
+    NSParameterAssert([post isKindOfClass:[RemotePost class]]);
+    
+    NSString *path = [NSString stringWithFormat:@"sites/%@/posts/%@/autosave", self.siteID, post.postID];
+    NSString *requestUrl = [self pathForEndpoint:path
+                                     withVersion:ServiceRemoteWordPressComRESTApiVersion_1_1];
+    
+    NSDictionary *parameters = [self parametersWithRemotePost:post];
+    
+    [self.wordPressComRestApi POST:requestUrl
+                        parameters:parameters
+                           success:^(id responseObject, NSHTTPURLResponse *httpResponse) {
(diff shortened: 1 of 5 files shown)


diegoreymendez:
Considering this is the first endpoint (that I know of) that works just for the WP.com Rest API and not for XMLRPC, I think we may want to offer some mechanism to let the calling App the service is not available.

I also kind of feel either the success or failure block should always be executed.

One possibility could be to return a specific type of error in the failure block.

This would allow the calling App to know that this service is unavailable, and follow a different logic path if necessary.

Thoughts, @yaelirub & @aerych?

aerych:
> Considering this is the first endpoint (that I know of) that works just for the WP.com Rest API and not for XMLRPC,

We have others.  For example there is 
```
- (void)createPost:(RemotePost *)post
         withMedia:(RemoteMedia *)media
   requestEnqueued:(void (^)(NSNumber *taskID))requestEnqueued
           success:(void (^)(RemotePost *))success
           failure:(void (^)(NSError *))failure;
```
which is defined in PostServiceRemoteREST.h instead of `PostServiceRemote.h`.  

I think its been an accepted practice to put methods specific to one API or the other in their respective headers. (I wasn't thinking about this earlier, the ☕️ is just now kicking in 😬).  So maybe the better approach would be to move ```
- (void)savePost:(RemotePost *)post
         success:(void (^)(RemotePost *post, NSString *previewURL))success
```
to `PostServiceRemoteREST.h`.  This would help prevent it from being called when it shouldn't and remove the need for calling `failure` or logging.   Thoughts?

File: WordPressKit/PostServiceRemoteREST.m [hunk 1]
```diff
@@ -183,6 +183,33 @@ - (void)updatePost:(RemotePost *)post
            }];
 }
 
+- (void)savePost:(RemotePost *)post
```
aerych: I might be overthinking this.
Do you think `savePost` could potentially be conflated with `updatePost`?  I mean, would it be reasonable to think that calling a save method should update the saved content in the canonical post record?  The reason I ask is I was looking at the behavior of the /autosave endpoint and got to thinking about expectations when I saw this:
```
// Drafts and auto-drafts are just overwritten by autosave for the same user if the post is not locked
// Non drafts or other users drafts are not overwritten. The autosave is stored in a special post revision for each user.
```
So depending on post status and authorship a revision might be created instead of updating the post. 

Would it make sense to name things `autosave` to reflect the endpoint's specific behavior?

yaelirub:
@diegoreymendez , I am creating a specialized error and returning it in the implementation here: https://github.com/wordpress-mobile/WordPress-iOS/blob/596287851d317e1176e67cbde2adae89c846da55/WordPress/Classes/Services/PostService.m#L277

WordPressKit/PostServiceRemoteREST.m
yaelirub: I believe you're right. I will update the name and add that comment in the method description here too.
As a side note, I would also like to add documentation to this endpoint after the main issue is merged.

yaelirub:
@aerych , thank you so much for your review. I updated the PR based on your comments and added tests. Ready for another look.

aerych:
Looks good to me!  All tests pass.  :shipit: when ready! :)

PR #166, opened 2019-07-15
Implement POST editor settings

etoledom:
This PR adds the ability of setting the default mobile editor to Remote.
Fixes part of https://github.com/wordpress-mobile/WordPress-iOS/issues/10870

WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12123

### Testing Details

Follow the instructions from https://github.com/wordpress-mobile/WordPress-iOS/pull/12123

### DIFF
--- a/WordPressKit/EditorSettings.swift
+++ b/WordPressKit/EditorSettings.swift
@@ -11,7 +11,7 @@ public enum EditorSettings: String {
 }
 
 extension EditorSettings {
-    init (with response: AnyObject) throws {
+    init(with response: AnyObject) throws {
         guard let response = response as? [String: AnyObject] else {
             throw NSError(domain: NSURLErrorDomain, code: NSURLErrorBadServerResponse, userInfo: nil)
         }
--- a/WordPressKit/EditorSettings.swift
+++ b/WordPressKit/EditorSettings.swift
@@ -1,0 +1,27 @@
+private struct RemoteEditorSettings: Codable {
+    let editor_mobile: String
+    let editor_web: String
+}
+
+public enum EditorSettings: String {
+    case gutenberg
+    case aztec
+
+    static let `default` = EditorSettings.aztec
+}
+
+extension EditorSettings {
+    init (with response: AnyObject) throws {
+        guard let response = response as? [String: AnyObject] else {
+            throw NSError(domain: NSURLErrorDomain, code: NSURLErrorBadServerResponse, userInfo: nil)
+        }
+
+        let data = try JSONSerialization.data(withJSONObject: response, options: .prettyPrinted)
+        let editorPreferenesRemote = try JSONDecoder().decode(RemoteEditorSettings.self, from: data)
+        self.init(with: editorPreferenesRemote)
+    }
+
+    private init(with remote: RemoteEditorSettings) {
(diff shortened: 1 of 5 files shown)


WordPressKit/EditorSettings.swift
```diff
@@ -0,0 +1,27 @@
...
+
+public enum EditorSettings: String {
+    case gutenberg
+    case aztec
+
+    static let `default` = EditorSettings.aztec
+}
+
+extension EditorSettings {
+    init (with response: AnyObject) throws {
```
koke: The space after the init looks off. @wordpress-mobile/platform-9 we don't have swiftlint running on all repos?

WordPressKit/EditorSettings.swift
```diff
@@ -0,0 +1,27 @@
+private struct RemoteEditorSettings: Codable {
+    let editor_mobile: String
```
koke: That underscore in the name looks odd. I think other models use the standard camel case and then use `CodingKeys` to specify the JSON names.

WordPressKit/EditorSettings.swift
etoledom: Nice catch! Fixed 👍

WordPressKit/EditorSettings.swift
etoledom: Changed to camelCase.
Using `CodingKeys` seems to be remanence from an older Swift Version (in some cases at least?) 🤔 
Unless we want to map to a completely different property name.

Fixed using a helper from JSONDecoder to parse snake case into camel case. This was already in place 👍

koke: 🤯 `.convertFromSnakeCase` TIL

koke (approved):

WordPressKit/EditorSettings.swift
jtreanor: Yes, we don't have swiftlint running in the non-app repos. Feel free to open a request for us to add that 😄

WordPressKit/EditorSettings.swift
```diff
@@ -22,6 +32,8 @@ extension EditorSettings {
     }
 
     private init(with remote: RemoteEditorSettings) {
-        self = EditorSettings(rawValue: remote.editorMobile) ?? .default
```
koke: Considering this is only being called from another initializer that `throws`, I wonder if it'd be cleaner to avoid those `default` values and just throw an error if the JSON has unexpected values. Then the app can decide what's the default/fallback value if it can't read one

etoledom: Sounds good. I was unsure about those defaults myself. It's something I was considering asking 🙂

koke:
I'm seeing a few changes since I approved this one, does this need another review?

etoledom:
@koke 

What's new is:
- `getEditorSettings` GET request
- Handling the case when the server responds with an empty `mobile_gutenberg` property.

It's probably good to check it out again and publish the new POD version when approved.
in the meantime I'll update from develop.

etoledom:
By the commit history, it looks like the GET was already there when you reviewed 🤔 
But the second point is new for sure.

PR #172, opened 2019-07-24
[Update Username] Map new property in account settings GET and new method

danielebogo:
### Description
This is part of [#9705](https://github.com/wordpress-mobile/WordPress-iOS/issues/9705)
This PR add a new property `usernameCanBeChanged` to map when the `GET me/settings` is called.
This property is used to define if a user is able to change the username. 

There's also a new method added to validate the current username. It's a GET `me/username/validate/USERNAME`

### Testing Details
- Run the test `AccountSettingsRemoteTests`
- Run [#12198](https://github.com/wordpress-mobile/WordPress-iOS/pull/12198)


WordPressKit/AccountSettingsRemote.swift
```diff
@@ -90,6 +90,26 @@ public class AccountSettingsRemote: ServiceRemoteWordPressComREST {
...
 
+    /// Validate the current user's username
+    ///
+    /// - Parameters:
+    ///   - username: The new username
+    ///   - success: block for success
+    ///   - failure: block for failure
+    public func validateUsername(to username: String, success: @escaping () -> Void, failure: @escaping (Error) -> Void) {
+        let endpoint = "me/username/validate/\(username)"
+        let path = self.path(forEndpoint: endpoint, withVersion: ._1_1)
```
aerych: Do we need `self.` here?

WordPressKitTests/AccountSettingsRemoteTests.swift
```diff
@@ -62,6 +62,7 @@ class AccountSettingsRemoteTests: RemoteTestCase, RESTTestable {
         stubRemoteResponse(meSettingsEndpoint, filename: getAccountSettingsSuccessMockFilename, contentType: .ApplicationJSON)
         remote.getSettings(success: { settings in
             XCTAssertEqual(settings.username, self.username, "The usernames should be equal.")
+            XCTAssertEqual(settings.usernameCanBeChanged, false, "The usernames should be equal.")
```
aerych: Was the message in this line a copy/paste from the line above? Is there something more descriptive we could add?

WordPressKit/AccountSettingsRemote.swift
```diff
@@ -90,6 +90,26 @@ public class AccountSettingsRemote: ServiceRemoteWordPressComREST {
...
+        
+        wordPressComRestApi.GET(path,
+                                parameters: nil,
+                                success: { responseObject, httpResponse in
+                                    success()
```
aerych: I'm curious why nothing is done with the responseObject.  The results contain a list of available actions that include: `new`, `none`, `rename_discard` and `rename_keep` which determine the behavior a username change has on the user's blog.  Is the expectation always that a change will be made with `none`?  

Maybe add a comment here to clarify what the intent is?

WordPressKit/AccountSettingsRemote.swift
danielebogo: It's to avoid the `Variable used within its own initial value` error. I left it because it's used in all the other methods of the same service.

WordPressKit/AccountSettingsRemote.swift
danielebogo: Testing the API response I can see only `new` and `none` as options:
```
{
    "success": true,
    "allowed_actions": {
        "new": "Yes, create a new blog to match my new user name",
        "none": "No, don't create a matching blog address."
    }
}
```
In Calypso too. I left it _blank_ because the `changeUsername` method set the action as `none`. I'll leave a comment saying that the block should be changed if the `allowed_actions` is required.

WordPressKit/AccountSettingsRemote.swift
aerych: So it is.  Interesting.  Also kind of annoying that the linter doesn't recognize the full method signature when validating.  `self.` here shouldn't be necessary 😠.  grumble grumble grumble.  I'll stop ranting now.  Thanks for clarifying!

WordPressKit/AccountSettingsRemote.swift
aerych: I looked a little at the endpoint code earlier but I don't remember the circumstances that would return all four values. Fwiw, this is what I see from my `aerychtest` account.
Screen Shot 2019-07-25 at 12 52 06 PM

PR #174, opened 2019-07-29
Stats: add File Download support

ScoutHarris:
### Description

Ref #https://github.com/wordpress-mobile/WordPress-iOS/issues/11963

This adds fetching and storing data for File Download stats.

There are a couple of caveats:
- The endpoint isn't live yet, and thus isn't returning any data. So there is a `TODO` to update the keys when it is.
- Also because of the above, the test validates what the endpoint is currently returning. There is a `TODO` there as well to update later.

### Testing Details

- Verify the tests run, especially the new `testFetchDownloadsData` test.

### DIFF
--- a/WordPressKitTests/StatsRemoteV2Tests.swift
+++ b/WordPressKitTests/StatsRemoteV2Tests.swift
@@ -20,6 +20,7 @@ class StatsRemoteV2Tests: RemoteTestCase, RESTTestable {
     let getVisitsMonthMockFilename = "stats-visits-month.json"
     let getPostsMockFilename = "stats-posts-data.json"
     let getPublishedPostsFilename = "stats-published-posts.json"
+    let getDownloadsDataFilename = "stats-file-downloads.json"
     let getPostsDetailsFilename = "stats-post-details.json"
 
     // MARK: - Properties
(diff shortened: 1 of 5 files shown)


WordPressKitTests/StatsRemoteV2Tests.swift
```diff
@@ -364,7 +365,35 @@ class StatsRemoteV2Tests: RemoteTestCase, RESTTestable {
...
+        remote.getData(for: .month, endingOn: date) { (fileDownloads: StatsFileDownloadsTimeIntervalData?, error: Error?) in
+            XCTAssertNil(error)
+            XCTAssertNotNil(fileDownloads)
+
+            XCTAssertEqual(fileDownloads?.otherDownloadsCount, 0)
+            XCTAssertEqual(fileDownloads?.totalDownloadsCount, 0)
+
+            XCTAssertEqual(fileDownloads?.fileDownloads.count, 0)
+//            XCTAssertEqual(fileDownloads?.fileDownloads.first!.file, "")
+//            XCTAssertEqual(fileDownloads?.fileDownloads.first!.downloadCount, 666)
```
danielebogo: Do we need these tests? If not can we delete these comments?

ScoutHarris: Hey @danielebogo . I left these in so I wouldn't forgot to add them back in later. I'll add a `TODO` explaining why they are there. 🤞 they'll be updated before this is released.

PR #182, opened 2019-08-22
Add Apple specific parameters to social service connect

ScoutHarris:
### Description

Fixes #n/a

This does two things:
1) Adds an `apple` service name. This was needed for the associated WPAuth change.
2) Adds Apple specific parameters to the `social-login/connect` endpoint. This wasn't needed yet, but hey, while we're here...

### Testing Details

This can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12360
Ref WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/117

### DIFF
--- a/WordPressKit/AccountServiceRemoteREST+SocialService.swift
+++ b/WordPressKit/AccountServiceRemoteREST+SocialService.swift
@@ -2,6 +2,7 @@ import Foundation
 
 public enum SocialServiceName: String {
     case google
+    case apple
 }
 
 
--- a/WordPressKit/AccountServiceRemoteREST+SocialService.swift
+++ b/WordPressKit/AccountServiceRemoteREST+SocialService.swift
@@ -13,19 +13,35 @@ extension AccountServiceRemoteREST {
     /// - Parameters:
     ///     - service The name of the social service.
     ///     - token The OpenID Connect (JWT) ID token identifying the user on the social service.
+    ///     - appleEmail The email address associated with an Apple ID. Optional.
+    ///     - appleFullName The full name associated with an Apple ID. Optional.
     ///     - oAuthClientID The WPCOM REST API client ID.
     ///     - oAuthClientSecret The WPCOM REST API client secret.
     ///     - success The block that will be executed on success.
     ///     - failure The block that will be executed on failure.
-    public func connectToSocialService(_ service: SocialServiceName, serviceIDToken token: String, oAuthClientID: String, oAuthClientSecret: String, success:@escaping (() -> Void), failure:@escaping ((NSError) -> Void)) {
+    public func connectToSocialService(_ service: SocialServiceName,
+                                       serviceIDToken token: String,
+                                       appleEmail: String? = nil,
+                                       appleFullName: String? = nil,
+                                       oAuthClientID: String,
+                                       oAuthClientSecret: String,
+                                       success:@escaping (() -> Void),
+                                       failure:@escaping ((NSError) -> Void)) {
         let path = self.path(forEndpoint: "me/social-login/connect", withVersion: ._1_1)
 
-        let params = [
+        var params = [
             "client_id": oAuthClientID,
             "client_secret": oAuthClientSecret,
             "service": service.rawValue,
             "id_token": token,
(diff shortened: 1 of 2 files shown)


WordPressKit/AccountServiceRemoteREST+SocialService.swift
```diff
@@ -12,19 +13,35 @@ extension AccountServiceRemoteREST {
...
+    ///     - appleEmail The email address associated with an Apple ID. Optional.
+    ///     - appleFullName The full name associated with an Apple ID. Optional.
     ///     - oAuthClientID The WPCOM REST API client ID.
     ///     - oAuthClientSecret The WPCOM REST API client secret.
     ///     - success The block that will be executed on success.
     ///     - failure The block that will be executed on failure.
-    public func connectToSocialService(_ service: SocialServiceName, serviceIDToken token: String, oAuthClientID: String, oAuthClientSecret: String, success:@escaping (() -> Void), failure:@escaping ((NSError) -> Void)) {
+    public func connectToSocialService(_ service: SocialServiceName,
+                                       serviceIDToken token: String,
+                                       appleEmail: String? = nil,
```
frosty: I feel like these apple parameters don't really belong in here 🤔 The method can be used for any service, so having apple specific parameters there if you're using Google doesn't feel quite right to me (even though you can ignore them because of the defaults).

How about if we just accept a parameters dictionary, which we can append to our own `params`? We could add a separate method to generate it, like `appleSignInParameters(email:fullName:)`, so the client doesn't need to know the correct keys for those items.

frosty: Or we could just pass in `userEmail` and `userFullName`, and the services can do with them what they want (or ignore them)

ScoutHarris:
Hey @frosty . Changes made and ready for another review please.

WordPressKit/AccountServiceRemoteREST+SocialService.swift
```diff
@@ -13,16 +13,14 @@ extension AccountServiceRemoteREST {
...
+    ///     - appleConnectParameters Dictionary containing endpoint parameters specific to connecting to Apple service.
     ///     - oAuthClientID The WPCOM REST API client ID.
     ///     - oAuthClientSecret The WPCOM REST API client secret.
     ///     - success The block that will be executed on success.
     ///     - failure The block that will be executed on failure.
     public func connectToSocialService(_ service: SocialServiceName,
                                        serviceIDToken token: String,
-                                       appleEmail: String? = nil,
-                                       appleFullName: String? = nil,
+                                       appleConnectParameters: [String:AnyObject]? = nil,
```
frosty: So could we just name this `connectParameters`? Perhaps the doc comment could say "currently only used for the Apple service".

WordPressKit/AccountServiceRemoteREST+SocialService.swift
```diff
@@ -37,9 +35,8 @@ extension AccountServiceRemoteREST {
         ] as [String: AnyObject]
         
         // Append Apple specific parameters
-        if service == .apple {
-            params["user_email"] = appleEmail as AnyObject
-            params["user_name"] = appleFullName as AnyObject
+        if service == .apple, let appleConnectParameters = appleConnectParameters {
```
frosty: We could just always append the parameters regardless of the service, then we don't need special conditional code for Apple?

ScoutHarris: I figured it's better not to append them just in case, but ok.

WordPressKit/AccountServiceRemoteREST+SocialService.swift
ScoutHarris: Sure.

ScoutHarris:
> make the method less Apple centric.
I'm not trying to make it Apple centric. I'm trying to point out that these params are for Apple only. But ok, I'll generalize everything and just have a comment.

ScoutHarris:
OK @frosty . One more time (🤞 ) please kind sir.

PR #190, opened 2019-10-17
Added signup_flow_name attribute

yaelirub:
to account creation with email

### Description
https://github.com/wordpress-mobile/WordPress-iOS/issues/12724

To test: 
 in WPiOS update WordPressKit pod to: 
```pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :branch => 'task/add_signup_attribute'```
and test signup with email works

Seems like `createWPComAccountWithEmail` is not being used anymore (which is where I originally added the param. I created an issue to remove this method : #191 

### DIFF
--- a/WordPressKit/AccountServiceRemoteREST.m
+++ b/WordPressKit/AccountServiceRemoteREST.m
@@ -245,7 +245,7 @@ static NSString * const UserDictionaryEmailVerifiedKey = @"email_verified";
     NSMutableDictionary *params = [NSMutableDictionary dictionaryWithDictionary:@{
                                                                                   @"email": email,
                                                                                   @"client_id": clientID,
-                                                                                  @"client_secret": clientSecret,
+                                                                                  @"client_secret": clientSecret
     }];
 
     if (![@"wordpress" isEqualToString:scheme]) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by yaelirub** [comment]
@leandroalonso , updated and ready for another look

File: WordPressKit/AccountServiceRemoteREST.m [hunk 1]
```diff
@@ -240,7 +240,8 @@ - (void)requestWPComMagicLinkForEmail:(NSString *)email
     NSMutableDictionary *params = [NSMutableDictionary dictionaryWithDictionary:@{
                                                                                   @"email": email,
                                                                                   @"client_id": clientID,
-                                                                                  @"client_secret": clientSecret
+                                                                                  @"client_secret": clientSecret,
+                                                                                  @"signup_flow_name": @"mobile-ios"
```
leandroalonso: This method is used both for requesting:

* A magic link to sign in
* A magic link to sign up

I guess we only want the `signup_flow_name` for the last one. Is that right?

If so, we can add an extra parameter and set it here: https://github.com/wordpress-mobile/WordPressKit-iOS/blob/c67da35d2bb75f62cfeeb89fdbce57117588f57c/WordPressKit/AccountServiceRemoteREST.m#L221

yaelirub:
@leandroalonso , thanks for this review. Updated accordingly. 

I checked the requests body for both login and signup in charles proxy and the new parameter is added to signup and not login.

IMG_4999

leandroalonso:
@yaelirub Works as expected. :shipit:

PR #192, opened 2019-10-27
Add autosave attributes to RemotePost

guarani:
### Description

Post [autosave](https://wordpress.org/support/article/revisions/#autosaves) attributes are added to `RemotePost` - the type that describes a post (or page) fetched from the API. Autosave data is obtained by including `autosave` in the `meta` parameter, but at the time of writing this is undocumented.

Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/12141

### REST API

The title, content, excerpt and date of last modification are fetched from the following REST API endpoint:
> https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/

⚠️I think autosave is not currently supported on the XML RPC API as it has not been included in the Android changes I've come across (please let me know if this seems incorrect) ⚠️

https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1357/files
https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1391/files

### Testing Details

These changes are best tested alongside the matching WordPress-iOS PR https://github.com/wordpress-mobile/WordPress-iOS/pull/12826

### Unit Tests

The new `meta` parameter uses the existing mechanism for specifying options, `PostServiceRemoteOptions`. There is already a test that asserts that arbitrary options are correctly set on the request payload (`- testThatGetPostsOfTypeWithOptionsWorks`), so I think the current tests have this covered.


WordPressKit/RemotePost.h
```diff
@@ -34,6 +34,10 @@ extern NSString * const PostStatusDeleted;
 @property (nonatomic, strong) NSString *postThumbnailPath;
 @property (nonatomic, strong) NSString *type;
 @property (nonatomic, strong) NSString *format;
+@property (nonatomic, copy, nullable) NSString *autosaveTitle;
+@property (nonatomic, copy, nullable) NSString *autosaveContent;
+@property (nonatomic, copy, nullable) NSString *autosaveExcerpt;
+@property (nonatomic, copy, nullable) NSDate *autosaveModifiedDate;
```
guarani: `copy` keeps these properties immutable - even though most of the other properties are mutable.

guarani:
> ⚠️I think autosave is not currently supported on the XML RPC API as it has not been included in the Android changes I've come across (please let me know if this seems incorrect) ⚠️
Update: Autosave is _not_ supported by the XML RPC API.

WordPressKit/PostServiceRemoteREST.m
```diff
@@ -432,7 +436,13 @@ - (RemotePost *)remotePostFromJSONDictionary:(NSDictionary *)jsonPost {
     post.tags = [self tagNamesFromJSONDictionary:jsonPost[@"tags"]];
 
     post.revisions = [jsonPost arrayForKey:@"revisions"];
-    
+
+    NSDictionary *autosave = jsonPost[@"meta"][@"data"][@"autosave"];
+    post.autosaveTitle = autosave[@"title"];
+    post.autosaveContent = autosave[@"content"];
+    post.autosaveExcerpt = autosave[@"excerpt"];
+    post.autosaveModifiedDate = [NSDate dateWithWordPressComJSONString:autosave[@"modified"]];
```
shiki: As mentioned in https://github.com/wordpress-mobile/WordPress-iOS/pull/12781:

- The `s` should be capitalized. 
- Let's add the other available properties even if we don't need them. Since they come for free in here, it's more expensive to modify `WordPressKit` again and add them later if we find out we need them.

shiki: What do you think about making a separate class for this? Just so the `RemotePost` is a bit closer to the API result. 

```swift 
class RemotePostAutoSave {
    let title: String?
    let content: String?
    ...
}
```

The WPiOS flat properties are fine since they are used in Core Data.

guarani: > What do you think about making a separate class for this? Just so the RemotePost is a bit closer to the API result.

I agree - and with more attributes being mapped it's especially useful to keep them in a separate class.

Leaving `s` uncapitalized as per discussion: https://github.com/wordpress-mobile/WordPress-iOS/pull/12781#discussion_r341083406

guarani:
@shiki, yes. I've now added a test `testFetchPostsPerformsAutosaveMappingSuccessfully` 👍

WordPressKit/RemotePostAutosave.swift
```diff
@@ -0,0 +1,15 @@
...
+@objcMembers
+open class RemotePostAutosave: NSObject {
+    open var title: String?
+    open var excerpt: String?
+    open var content: String?
+    open var modifiedDate: Date?
+    open var identifier: NSNumber?
+    open var authorID: String?
+    open var postID: NSNumber?
+    open var previewURL: String?
```
shiki: As I understand it, `open` would allow consumers of this library to create a subclass of `RemotePostAutosave`. I don't think we really need that. The `open` attribute would probably add some confusion about the intent. What do you think about making this class and its properties `public` instead? 

I experimented with `public internal(set)` but that does not seem to work well with Objective-C. 😞

shiki:
@guarani Would you mind incrementing the pod's beta version as well? Just like this commit: 9c3a55e3ce5a55809a7ba7ce5c39b1da1a170a45

WordPressKit/RemotePostAutosave.swift
guarani: I missed that, updated now. 

I tested `internal(set)` now too. It didn't work either and I found a [SO question](https://stackoverflow.com/q/32700078/1305067) which linked to [this Apple doc](https://developer.apple.com/documentation/swift/imported_c_and_objective-c_apis/importing_swift_into_objective-c):

> Methods and properties that are marked with the internal modifier and declared within a class that inherits from an Objective-C class are accessible to the Objective-C runtime. However, they're inaccessible at compile time and don't appear in the generated header for a framework target.

So they're there at runtime - just not at compile time. 😔

WordPressKit/RemotePost.h
```diff
@@ -34,7 +35,7 @@ extern NSString * const PostStatusDeleted;
 @property (nonatomic, strong) NSString *postThumbnailPath;
 @property (nonatomic, strong) NSString *type;
 @property (nonatomic, strong) NSString *format;
-
+@property (nonatomic, strong) RemotePostAutosave *autosave;
```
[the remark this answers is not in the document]
guarani: Thanks, I agree a comment here helps a lot until we can audit all the properties at once.

guarani:
Thanks again for the review, @shiki! I've updated the PR to address your comments.

WordPressKit/PostServiceRemoteREST.m
```diff
@@ -432,7 +436,21 @@ - (RemotePost *)remotePostFromJSONDictionary:(NSDictionary *)jsonPost {
     post.tags = [self tagNamesFromJSONDictionary:jsonPost[@"tags"]];
 
     post.revisions = [jsonPost arrayForKey:@"revisions"];
-    
+
+    NSDictionary *autosaveAttributes = jsonPost[@"meta"][@"data"][@"autosave"];
+    if (autosaveAttributes) {
```
guarani: If `autosaveAttributes` is `null` (in the response JSON), it will be deserialized into `NSNull` in the dictionary - only if the key is not present will it return `nil`. I'm going to update this to `wp_isValidObject` to handle this case.

https://github.com/wordpress-mobile/WordPressKit-iOS/blob/78a0a0ef53e7a5c5890930cd876328c3688e467c/WordPressKit/ObjectValidation.swift#L9

shiki: Great find! Thanks for fixing it.

guarani:
Thank you, @shiki!

PR #197, opened 2019-11-08
Fix XML-RPC comment moderation to use correct comment ID

rachelmcr:
### Description

iOS issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/12909

Updates the method to moderate a comment so it uses the correct comment ID, instead of the boolean value in the `wp.editComment` response. (This is the same fix used for a similar issue [here](https://github.com/wordpress-mobile/WordPress-iOS/commit/ed2a107b1fd2b82fc543d65bdb7678e20d97d774).)

### Testing Details

Test instructions in this PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12910

### DIFF
--- a/WordPressKit/CommentServiceRemoteXMLRPC.m
+++ b/WordPressKit/CommentServiceRemoteXMLRPC.m
@@ -119,7 +119,7 @@
     NSParameterAssert(comment.commentID != nil);
     NSNumber *commentID = comment.commentID;
     NSArray *extraParameters = @[
-                                 comment.commentID,
+                                 commentID,
                                  @{@"status": comment.status},
                                  ];
     NSArray *parameters = [self XMLRPCArgumentsWithExtra:extraParameters];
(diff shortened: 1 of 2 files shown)


WordPressKit/CommentServiceRemoteXMLRPC.m
```diff
@@ -117,6 +117,7 @@ - (void)moderateComment:(RemoteComment *)comment
                 failure:(void (^)(NSError *))failure
 {
     NSParameterAssert(comment.commentID != nil);
+    NSNumber *commentID = comment.commentID;
     NSArray *extraParameters = @[
                                  comment.commentID,
```
aerych: One tiny suggestion.  Maybe use `commentID` here as well rather than going through the `comment` instance as it currently does? This would make how the comment ID was referenced consistent throughout this method.

PR #202, opened 2019-11-28
[PoC] WP-API and Cookie Authentication

koke:
### Description

We are currently lacking a way to perform requests to the [WordPress REST API](https://developer.wordpress.org/rest-api/) because we lack a good authentication method. The only method available today is based on [cookies and nonces](https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/). There might be support for [Basic Authentication](https://core.trac.wordpress.org/ticket/42790) and [oAuth2](https://github.com/wp-api/authentication) eventually, but there's no timeline for that.

At the same time, we are coming across more and more features in gutenberg-mobile that require getting data from the API. In https://github.com/wordpress-mobile/WordPress-iOS/pull/12714 we managed to work around this issue because we could get enough information through an unauthenticated request. As part of https://github.com/wordpress-mobile/gutenberg-mobile/issues/1513 we will need to get the theme's color palette that will be exposed by https://core.trac.wordpress.org/ticket/48798, but only for authenticated users.

In this scenario, I think it makes sense to make an attempt at implementing a solution for the only available auth method, as fragile as that might be, so here it is.

This new API Client supports both Token and Cookie authentication, although only the latter has been tested so far (`GutenbergNetworkRequest` is already reusing the existing WordPress.com API Client for WordPress.com sites), but the basic design would allow for the other future authentication methods (detecting which one we should use might make things interesting, but that's not a problem for today). 

The `RequestAdapter` and `RequestRetrier` patterns that Alamofire offer seem pretty convenient for authentication, and if this moves forward, it would be interesting to refactor the existing `WordPressComRestApi` client, since there is a lot of overlap with the .org one, and a lot of the specifics could be configurable via adapters. 

### How it works

When an API request fails with a 401, the CookieNonce authenticator will POST to `wp-login.php` and ask to redirect to the new post page (`wp-admin/post-new.php`). The editor is the only page I've found where we can extract a usable nonce. Then the authenticator will look at the HTML source and extract the nonce from the Gutenberg initialization code. This also means the method won't work if they don't have Gutenberg as the default editor for new posts (pre-5.0 or classic editor plugin), but we don't have much of an alternative there.

Nonces have an expiration of 1 day, so instead of complicating things too much with storage and invalidation, I'm just keeping the nonce in memory as part of the Authenticator. This is tied to the lifetime of the API Client, which is tied to the Blog. I'm not checking for the presence of an existing nonce or cookie, so if a request fails because either expired, we would try to authenticate again and resolve the issue.

This also means that we still have to find a way to flag when the authentication fails. We don't want to be retrying the login on every API request if it's not going to work, and could lead to hosting providers blocking the user 😱 

### Testing Details

I've tested this from WPiOS https://github.com/wordpress-mobile/WordPress-iOS/pull/13020 and the steps described in https://github.com/wordpress-mobile/WordPress-iOS/pull/12714:

- Run metro server on develop
- Add a image block to gutenberg.
- Select an image from the WordPress media Library.
- Select settings and choose a different image setting.
- Change to HTML mode.
- Check that the image URL has changes according to the selected image size.


pinarol:
I am trying to test this on my self-hosted site but couldn't make it work:

imageSize

I can provide admin credentials from slack

WordPressKit/Secret.swift
```diff
@@ -0,0 +1,49 @@
...
+///
+/// Usage example
+///
+/// ```
+/// let password = Secret("my secret password")
+/// print(password)             // Prints "--redacted"
+/// print(password.secretValue) // Prints "my secret password"
+/// ```
+///
+public struct Secret<T> {
```
pinarol: TIL

WordPressKit/WordPressOrgRestApi.swift
```diff
@@ -0,0 +1,90 @@
...
+        return request(method: .get, path: path, parameters: parameters, completion: completion)
+    }
+
+    private func request(method: HTTPMethod,
+                         path: String,
+                         parameters: [String: AnyObject]?,
+                         completion: @escaping Completion) -> Progress? {
+        let relativePath = path.removingPrefix("/")
+        guard let url = URL(string: relativePath, relativeTo: apiBase) else {
+            // FIXME: make tese .org errors
```
pinarol: was going to tell the same thing then I saw this FIXME :)

pinarol:
As you pointed out in slack redirecting to this screen was the problem:

Screen Shot 2019-12-02 at 11 04 32

After solving that issue the current solution started working, it has a bit of a latency compared to .com but I think it is not a big problem:

imageSize2

Thank you for putting effort to this! This is totally some dark magic :)

This is absolutely making us support much more auth API calls on self hosted sites. But can we say that we shouldn't lean on to this solution much?

First reason is its fragility(it's not working if login page redirects us to some other page as above ☝️). 

Secondly, it is hard to tell how this could go wrong so we might need to revert this one day?

koke:
CircleCI is failing to validate the pod spec because the required Secret wrapper is in a WordPressShared beta. I'll update that once the final release is out, but the can be reviewed in the meantime.

WordPressKit/Authenticator.swift
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    // MARK: Request Adapter
+
+    public func adapt(_ urlRequest: URLRequest) throws -> URLRequest {
+        guard let nonce = nonce else {
+            return urlRequest
+        }
+        var adaptedRequest = urlRequest
+        adaptedRequest.addValue(nonce.secretValue, forHTTPHeaderField: "X-WP-Nonce")
```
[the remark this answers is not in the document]
koke: `nonce` is only in the constructor for testing purposes, but in the normal flow, it's actually set by the retrier if `startLoginSequence` is successful.

I understand that the adapter and the retrier might look unnecessarily coupled in this first iteration, but the idea was to make this the base for the upcoming auth mechanisms. Most authentication methods (if not all) will require some way to inject something in the request, and to be able to inspect the retry and potentially retry after re-authenticating.

I think I can simplify this a bit by making it a protocol instead of a struct.

> And in the future if we want to pass a different purpose RequestRetrier or RequestAdapter we wouldn't need to change this class. Thus, better interface segregation.

I also wondered about this, but I believe that if we want to do that we'll need to figure out a way to combine adapters and retriers, because you can only assign one to a session.

WordPressKit/Authenticator.swift
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    enum Error: Swift.Error {
+        case invalidNewPostURL
+        case postLoginFailed(Swift.Error)
+        case missingNonce
+        case unknown(Swift.Error)
+
+        var retriable: Bool {
+            return false
```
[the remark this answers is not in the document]
koke: I thought this was a good idea but then I couldn't figure out a good way to find out which errors should be retriable. I think it's probably best to remove that logic from this first iteration if we're not going to use it.

WordPressKit/Authenticator.swift
```diff
@@ -0,0 +1,231 @@
...
+        var components = URLComponents()
+        components.queryItems = parameters
+        request.httpBody = components.percentEncodedQuery?.data(using: .utf8)
+        return request
+    }
+
+}
+
+public extension CookieNonceAuthenticator {
+    static func cookieNonce(username: String, password: String, loginURL: URL, adminURL: URL, version: String) -> Authenticator {
```
pinarol: > static func cookieNonce(...)

This method doesn't look necessary since CookieNonceAuthenticator is public. One could simply use the initializer instead.

But, if the intention was to keep CookieNonceAuthenticator private and provide a kind of factory method for `Authenticator`(which I think would be better) then it would make sense to keep it but make it an extension for Authenticator instead:

```
public extension Authenticator { 
     static func cookieNonce(username: String, password: String, loginURL: URL, adminURL: URL, version: String) -> Authenticator {
....
```

pinarol: Same is true for:

```
public extension TokenAuthenticator {
    static func token(_ token: String, shouldAuthenticate: RequestAuthenticationValidator? = nil) -> Authenticator {
        return TokenAuthenticator(token: token, shouldAuthenticate: shouldAuthenticate)
    }

}

```

WordPressKit/Authenticator.swift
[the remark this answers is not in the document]
pinarol: > Most authentication methods (if not all) will require some way to inject something in the request, and to be able to inspect the retry and potentially retry after re-authenticating.

That makes sense. And I don't have a better suggestion currently about how to combine them. I can think of some other ways but can't claim they are any better.

WordPressKit/WordPressOrgRestApi.swift
```diff
@@ -0,0 +1,98 @@
...
+}
+
+@objc
+open class WordPressOrgRestApi: NSObject {
+    public typealias Completion = (Swift.Result<Any, Error>, HTTPURLResponse?) -> Void
+    private let apiBase: URL
+    private let authenticator: Authenticator?
+    private let userAgent: String?
+
+    public init(apiBase: URL, authenticator: Authenticator? = nil, userAgent: String? = nil) {
```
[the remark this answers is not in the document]
pinarol: Resolving this one since `Authenticator` interface has changed

WordPressKit/Authenticator.swift
koke: I agree that they're not very useful now. Those were initially created to be able to use the authenticators more like enums, so you would be able to do something like:

```swift
let authToken = "secret"
let api = WordPressOrgRestApi(
    url: baseURL,
    authenticator: .token(authToken)
)
```

However, since I ended up with [a convenience initializer](https://github.com/wordpress-mobile/WordPress-iOS/blob/92d7354c7342c41caaec4c8d4488161f0a7b3f64/WordPress/Classes/Networking/WordPressOrgRestApi%2BWordPress.swift#L37) that takes a `Blog`, these aren't being used in that way and can be removed.

SergioEstevao: For the record, I needed to change the code from being extension of Authenticator to the specific classes because of an error that the Swift compiler was returning. 
I assume this must have been some recent change on the Swift compiler, and I will remove the methods in favor of the blog init helpers.

SergioEstevao:
@etoledom and @pinarol can you test/review this again? I think it's ready to merge now.

pinarol:
I didn't test again but changes look good to me 👍

PR #226, opened 2020-03-10
Improve loading of large media libraries

SergioEstevao:
### Description

Fixes #https://github.com/wordpress-mobile/WordPress-iOS/issues/11644

This PR changes the interface of `getMediaLibraryWithSuccess:Failure` to support notification of page loads instead of all the media library only.

### Testing Details


WordPressKit/MediaServiceRemote.h
```diff
@@ -39,11 +39,13 @@
 /**
  *  Get Media items from blog using the options parameter.
  *
+ *  @param pageLoad a block to be executed when each page of media is loaded.
  *  @param success a block to be executed when the request finishes with success.
  *  @param failure a block to be execute when the request fails.
  */
-- (void)getMediaLibraryWithSuccess:(void (^)(NSArray *))success
-                           failure:(void (^)(NSError *))failure;
+- (void)getMediaLibraryWithPageLoad:(void (^)(NSArray *))pageLoad
```
frosty: It would be nice if we could introduce nullability here and make this nullable – but perhaps for a future PR :)

WordPressKit/MediaServiceRemoteXMLRPC.m
```diff
@@ -53,14 +55,17 @@ - (void)getMediaLibraryStartOffset:(NSUInteger)offset
                          return;
                      }
                      NSArray *pageMedia = [self remoteMediaFromXMLRPCArray:responseObject];
+                     if(pageLoad) {
```
frosty: The REST version of this only calls `pageLoad` if media has a > 0 item count, but it seems like this version calls it every time. I'm not sure what the 'correct' option would be, but I think these two should be consistent.

Issue #228, opened 2020-03-11
Error parsing plans response - raw price can be decimal.


maxme:
### Expected behavior

No error in the logs when loading the Plan screen.

### Actual behavior

It seems the parser is expecting int values for the prices but they should instead be doubles.

```
2020-03-04 15:56:59:708 Error parsing plans response for site dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "rawPrice", intValue: nil)], debugDescription: "Parsed JSON number <9.5> does not fit in Int.", underlyingError: nil))
```

diegoreymendez:
@maxme - I updated the ticket title and description to describe the actual issue, hope that's ok.

PR #230, opened 2020-03-12
Use Doubles for price values.

diegoreymendez:
### Description

Fixes #https://github.com/wordpress-mobile/WordPressKit-iOS/issues/228

The actual issue we were seeing is that some values are decimal, whereas our code was expecting integers.

I modified a unit test to have a decimal number so that it failed.  I then proceeded to fix our code to be able to handle that scenario, and now the unit tests run well once again.

### Testing Details

Just evaluating the logic of the changes should be enough, since the unit test validates that we're able to deal with decimal numbers now.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    CFPropertyList (3.0.1)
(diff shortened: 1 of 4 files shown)


diegoreymendez:
Thanks @aerych!

PR #231, opened 2020-03-12
Pin the ruby version for this project

jkmassel:
This ensures we’re using a consistent version, which will prevent bundler weirdness.

**To Test:**
1. See that CI is passing.
2. Check out this branch, and run `ruby -v`. The output should be something like `ruby 2.6.4p104 (2019-08-28 revision 67798) [x86_64-darwin19]`.


aerych:
(I was about to post this, but then I see that you've updated the version :) Still posting because it looked like the version in the .ruby-version file wasn't being picked up for me.) 

Testing as described, I saw the following when the .ruby-version is set to 2.5.5: 
```
$ ruby -v
ruby 2.6.4p104 (2019-08-28 revision 67798) [x86_64-darwin19]
```
I'm using rvm:
```
$ rvm list
   ruby-2.3.0 [ x86_64 ]
   ruby-2.5.5 [ x86_64 ]
   ruby-2.6.3 [ x86_64 ]
=* ruby-2.6.4 [ x86_64 ]

# => - current
# =* - current && default
#  * - default
```

jkmassel:
Ah, sorry, bad timing. You had the updated commit, but not the updated testing directions.

It looks like you've got the right version – should be `2.6.4`.

Sorry again!!

aerych:
Pretty sure I had the first commit only. 
```
$ cat .ruby-version 
2.5.5
```

aerych:
So for rvm folks, there is an extra `rvm use` call we need to make. :til:

PR #235, opened 2020-03-17
Fixes a SignUp regression.

diegoreymendez:
## Description

Fixes a signup regression.  Existing accounts should be redirected to login, instead of displaying an error.

## Associated branches:

WPiOS: `try/fix-signup-issue`
WPAuth: `try/fix-signup-issue`

### Test 1:

1. Check out the WPiOS branch.
2. Install the pods.
3. Run the app from scratch.
4. Try to sign up with an existing account.

Verify that you're redirected to login and that you an complete the login process.

### Test 2:

Same test steps as in Test 1, but try to sign up with a mailinator account instead (any random account will do, like `testing@mailinator.com`).

Verify that there's an error message explaining this email provider is not supported by us.

### Test 3:

Same test steps as in Test 1 and 2, but try to sign up using a valid / supported email account instead.

Verify that signup works.

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.6.0-beta.7"
+  s.version       = "4.6.0-beta.8"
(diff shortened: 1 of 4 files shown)


diegoreymendez:
Thanks @mindgraffiti !

PR #251, opened 2020-06-15
Remove ReaderPost formatter from the service

leandroalonso:
### Description

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/14313

### Testing Details

Details here: https://github.com/wordpress-mobile/WordPress-iOS/pull/14314

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.10.0"
+  s.version       = "4.11.0"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
   s.description   = <<-DESC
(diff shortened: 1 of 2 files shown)


WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "5.0.0-beta.2"
+  s.version       = "5.0.0-beta.1"
```
aerych, line 3: Should this be `5.0.0-beta.3` ?

leandroalonso: Yes!

I'll change it before merging. I just didn't it yet because I'll have to change WordPressShared too (is hardcoded to `5.0.0-beta.1`)

aerych: Ahh.. Gotcha!

PR #256, opened 2020-07-01
Add Reader Card service

leandroalonso:
Part of https://github.com/wordpress-mobile/WordPress-iOS/issues/14399

Review after #255 is merged

### Description

This PR adds the service to retrieve Cards from the Reader API.

So far it's able to parse Cards that represent a post or a collection of interests.

### How to test

The best way is to check `ReaderPostServiceRemoteCardTests`. The service API and the results can be debugged there.

### DIFF
--- a/WordPressKit/Decodable+Dictionary.swift
+++ b/WordPressKit/Decodable+Dictionary.swift
@@ -15,6 +15,7 @@ struct JSONCodingKeys: CodingKey {
 
 
 /// Add support do decode to a Dictionary
+/// From: https://stackoverflow.com/q/44603248
 extension KeyedDecodingContainer {
 
     func decode(_ type: Dictionary<String, Any>.Type, forKey key: K) throws -> Dictionary<String, Any> {
--- a/WordPressKit/Decodable+Dictionary.swift
+++ b/WordPressKit/Decodable+Dictionary.swift
@@ -1,0 +1,100 @@
+struct JSONCodingKeys: CodingKey {
+    var stringValue: String
+
+    init?(stringValue: String) {
+        self.stringValue = stringValue
+    }
+
+    var intValue: Int?
+
+    init?(intValue: Int) {
+        self.init(stringValue: "\(intValue)")
+        self.intValue = intValue
+    }
+}
+
+
+/// Add support do decode to a Dictionary
+extension KeyedDecodingContainer {
+
+    func decode(_ type: Dictionary<String, Any>.Type, forKey key: K) throws -> Dictionary<String, Any> {
+        let container = try self.nestedContainer(keyedBy: JSONCodingKeys.self, forKey: key)
+        return try container.decode(type)
+    }
+
+    func decodeIfPresent(_ type: Dictionary<String, Any>.Type, forKey key: K) throws -> Dictionary<String, Any>? {
+        guard contains(key) else {
+            return nil
+        }
+        guard try decodeNil(forKey: key) == false else {
+            return nil
+        }
+        return try decode(type, forKey: key)
+    }
+
+    func decode(_ type: Array<Any>.Type, forKey key: K) throws -> Array<Any> {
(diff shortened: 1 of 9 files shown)


WordPressKit/ReaderPostServiceRemote+Cards.swift
```diff
@@ -0,0 +1,34 @@
...
+                                        let data = try JSONSerialization.data(withJSONObject: response, options: [])
+                                        let envelope = try decoder.decode(ReaderCardEnvelope.self, from: data)
+
+                                        success(envelope.cards)
+                                    } catch {
+                                        DDLogError("\(error)")
+                                        failure(error)
+                                    }
+        }, failure: { error, _ in
+            DDLogError("\(error)")
```
emilylaguna, line 33:
```suggestion
            DDLogError("Error fetching reader cards: \(error)")
            
```

WordPressKit/Decodable+Dictionary.swift
```diff
@@ -0,0 +1,100 @@
...
+        self.init(stringValue: "\(intValue)")
+        self.intValue = intValue
+    }
+}
+
+
+/// Add support do decode to a Dictionary
+extension KeyedDecodingContainer {
+
+    func decode(_ type: Dictionary<String, Any>.Type, forKey key: K) throws -> Dictionary<String, Any> {
```
emilylaguna, line 20: Should we add unit tests for this extension?

WordPressKit/Decodable+Dictionary.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    init?(intValue: Int) {
+        self.init(stringValue: "\(intValue)")
+        self.intValue = intValue
+    }
+}
+
+
+/// Add support do decode to a Dictionary
+extension KeyedDecodingContainer {
```
emilylaguna, line 18: How about adding a link to where this originated from?

WordPressKit/ReaderPostServiceRemote+Cards.swift
```diff
@@ -0,0 +1,34 @@
...
+                                success: { (response, _) in
+
+                                    do {
+                                        let decoder = JSONDecoder()
+                                        let data = try JSONSerialization.data(withJSONObject: response, options: [])
+                                        let envelope = try decoder.decode(ReaderCardEnvelope.self, from: data)
+
+                                        success(envelope.cards)
+                                    } catch {
+                                        DDLogError("\(error)")
```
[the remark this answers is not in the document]
leandroalonso: This will log not only the error message but also the file, function, and line: https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/CocoaLumberjack.swift#L86

In that case, I don't think we need to be descriptive in the error message. Wdyt?

WordPressKit/ReaderPostServiceRemote+Cards.swift
leandroalonso: Same as above.

WordPressKit/Decodable+Dictionary.swift
leandroalonso: Done!

WordPressKit/Decodable+Dictionary.swift
leandroalonso: Yes, we should! df0da44bf24ed784990a9d1489c26a3eaeed77b9

leandroalonso:
@emilylaguna ready for another 👀

WordPressKit/ReaderPostServiceRemote+Cards.swift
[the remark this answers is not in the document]
emilylaguna: I think adding a small description of what the code was trying to do when it error'd out is important. This is also pretty common in other aspects of the app that use `DDLogError`.

PR #264, opened 2020-07-21
Plugin service tests

charliescheer:
Description
This PR adds unit tests to the parts of WordPressKit-iOS that handle plugins. The updates cover these files:
PluginDirectoryFeedPage.swift
PluginState.swift
SitePluginCapabilities.swift
SitePlugin.swift
PluginDirectoryEntry.swift
PluginServiceRemote.swift
PluginDirectoryServiceRemote.swift

This PR also begins the creation of dynamically generated mocks to be used as a way to stress test different functions associated with the plugins and plugin remotes.  These mocks are still fairly basic and I am continuing to expand on their capabilities and how the can be used in tests.

Testing Details
PR only deals with the addition of new tests, to test it, simply run the testing suite WordPressKitTests. All tests should pass.


Podfile.lock
```diff
@@ -68,4 +68,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: c118a2a371f6c1c49231939b145ae4b52eefa18f
 
-COCOAPODS: 1.8.4
+COCOAPODS: 1.9.1
```
jkmassel, line 71: I think this change is causing tests to fail, because it's a different version of CocoaPods than what is specified in https://github.com/wordpress-mobile/WordPressKit-iOS/blob/3f6f92b9a238e81551add3b2eae420a97a7b4e1d/Gemfile#L2.

WordPressKit/PluginState.swift
```diff
@@ -84,8 +83,10 @@ public struct PluginState: Equatable, Codable {
             && lhs.automanaged == rhs.automanaged
             && lhs.url == rhs.url
     }
+
```
jkmassel, line 86: We should get rid of whitespace changes that aren't "remove the extra spaces/tabs from this line" unless they improve code readability. Let's fix that across all files in this PR before we land this?

PR #267, opened 2020-08-05
Fix error when toggling Jetpack security settings.

etoledom:
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14590

Avoid sending `jetpack_protect_global_whitelist` if `protect` module is disabled.

Doing so, the server responds with a 500 error.

#### Problem:

When "Block malicious login attempts" is disabled, sending `jetpack_protect_global_whitelist` key in the body results on a server 500 error, even though all other settings are set properly.

We receive this error and display it to the user:

#### Solution:

Filtering the request body to not add the `jetpack_protect_global_whitelist` key if `Block malicious login attempts` is disabled solves the problem.

### Testing Details

- Have a Jetpack-connected self-hosted site.
- Go to Settings -> [Jetpack] Security.
- With `Block malicious login attempts` disabled, toggle `Allow WordPress.com login on and off.
  - Check that there are no error alerts.
- Enable `Block malicious login attempts`.
- Go to `Whitelisted IP addressed`
- Add an IP to the list.
  - Check on web that the IP was synchronised property.

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.15.0-beta.2"
+  s.version       = "4.15.0-beta.3"
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
 
   s.description   = <<-DESC
--- a/WordPressKit/BlogJetpackSettingsServiceRemote.swift
+++ b/WordPressKit/BlogJetpackSettingsServiceRemote.swift
@@ -211,15 +211,18 @@ private extension BlogJetpackSettingsServiceRemote {
                                                 serveImagesFromOurServers: serveImagesFromOurServersValue)
     }
 
-    func dictionaryFromJetpackSettings(_ settings: RemoteBlogJetpackSettings) -> [String: AnyObject] {
+    func dictionaryFromJetpackSettings(_ settings: RemoteBlogJetpackSettings) -> [String: Any] {
         let joinedIPs = settings.loginWhiteListedIPAddresses.joined(separator: ", ")
-        return [Keys.monitorEnabled: settings.monitorEnabled as AnyObject,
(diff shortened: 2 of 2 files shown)


guarani:
I see that the proposal here is to set `whiteListedIPAddresses` to `nil` when "Block malicious login attempts" is disabled. 
Since when we were previously sending the list, it returned a 500 error, that seems to indicate an unexpected error that might be an issue on the server itself, not a bad request. Do you think the solution might be a fix on the backend instead of here?

I had a look for [this endpoint](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/78a0a0ef53e7a5c5890930cd876328c3688e467c/WordPressKit/BlogJetpackSettingsServiceRemote.swift#L15) on the [WordPress.com API reference](https://developer.wordpress.com/docs/api/), but couldn't find it — do you know where to find it?

etoledom:
> I see that the proposal here is to set whiteListedIPAddresses to nil when "Block malicious login attempts" is disabled.

We are not sending `null` to the server though, `compactMap` will remove any key with a nil value from the dictionary. That seems to be the expected behaviour from the server.
`settingsDictionary.compactMapValues { $0 }`

>  that seems to indicate an unexpected error that might be an issue on the server itself

True that, let's communicate this to Jetpack people also. Still they might not allow this request (even though getting a proper error code this time) so I'd say it would be good to go ahead with this change. This is how web behaves too.

guarani:
> We are not sending null to the server though, compactMap will remove any key with a nil value from the dictionary. That seems to be the expected behaviour from the server.

Ah, I missed that, sorry. 

> True that, let's communicate this to Jetpack people also. Still they might not allow this request (even though getting a proper error code this time) so I'd say it would be good to go ahead with this change. This is how web behaves too.

Yeah I agree 👍

guarani (approved):

etoledom:
Thank you @Guarani for the review!

This is the Jetpack PR that solves this issue: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/267
Still, let's merge this PR.

PR #269, opened 2020-08-15
Issue/posts api subscribe

momo-ozawa:
### Description

This PR adds the following endpoints:

- [Get subscription state](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/%24post/subscribers/mine/)
- [Subscribe to a post comments stream](https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/posts/%24post/subscribers/new/)
- [Unsubscribe from a post comments stream](https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/posts/%24post/subscribers/mine/delete/)

### Testing Details
You can review this PR's functionality in the parent WordPress-iOS issue:
- https://github.com/wordpress-mobile/WordPress-iOS/issues/14574

### DIFF
--- a/WordPressKit/ReaderPostServiceRemote.h
+++ b/WordPressKit/ReaderPostServiceRemote.h
@@ -76,6 +76,32 @@ Fetches the subscription status of the specified post for the current user.
                                failure:(void (^)(NSError *))failure;
 
 /**
+ Mark a post as subscribed by the user.
+
+ @param postID The ID of the post.
+ @param siteID The ID of the site.
+ @param success block called on a successful fetch.
+ @param failure block called if there is any error. `error` can be any underlying network error.
+ */
+- (void)subscribeToPost:(NSUInteger)postID
+                forSite:(NSUInteger)siteID
+                success:(void (^)(void))success
+                failure:(void (^)(NSError *error))failure;
+
+/**
+ Mark a post as unsubscribed by the user.
+
+ @param postID The ID of the post.
+ @param siteID The ID of the site.
+ @param success block called on a successful fetch.
+ @param failure block called if there is any error. `error` can be any underlying network error.
+ */
+- (void)unsubscribeFromPost:(NSUInteger)postID
+                    forSite:(NSUInteger)siteID
+                    success:(void (^)(void))success
+                    failure:(void (^)(NSError *error))failure;
+
+/**
  Mark a post as liked by the user.
 
  @param postID The ID of the post.
--- a/WordPressKit/ReaderPostServiceRemote.h
+++ b/WordPressKit/ReaderPostServiceRemote.h
@@ -63,6 +63,19 @@
                failure:(void (^)(NSError *error))failure;
 
 /**
+Fetches the subscription status of the specified post for the current user.
+
+@param postID The ID of the post.
+@param siteID The ID of the site.
+@param success block called on a successful fetch.
+@param failure block called if there is any error. `error` can be any underlying network error.
+*/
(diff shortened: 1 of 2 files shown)


momo-ozawa:
@leandroalonso 
I'm thinking this will be a beta version bump? Let me know! 😄

leandroalonso:
>  I'm thinking this will be a beta version bump? Let me know! 😄

Yep!

WordPressKit/ReaderPostServiceRemote.h
```diff
@@ -62,6 +62,45 @@
...
 
+/**
+Fetches the subscription status of the specified post for the current user.
+
+@param postID The ID of the post.
+@param siteID The ID of the site.
+@param success block called on a successful fetch.
+@param failure block called if there is any error. `error` can be any underlying network error.
+*/
+- (void)fetchSubscriptionStatusForPost:(NSUInteger)postID
```
leandroalonso, line 73: We've been trying to write Swift when it comes to new code. Of course, that's a decision that always depends on multiple factors but I think for these new endpoints it makes sense. Wdyt?

WordPressKit/ReaderPostServiceRemote.h
```diff
@@ -62,6 +62,45 @@
                success:(void (^)(RemoteReaderPost *post))success
                failure:(void (^)(NSError *error))failure;
 
+/**
```
leandroalonso, line 65: I was thinking if this should really belong to `ReaderPostServiceRemote`.

We're implementing this feature under Reader in the WPiOS app... but thinking about the domain — and not the feature — maybe this can fit better somewhere else? Not sure though. What are your thoughts?

momo-ozawa: Hmmmm. That's a good question.

I decided to define the new subscribe/unsubscribe endpoints in `ReaderPostServiceRemote`, because the like/unlike endpoints are also defined there. I get your point about domain vs feature, and I'm trying to think of a better place, but I'm not so sure either. (Perhaps PostsService..?)

As for you suggestion about adding the new endpoints in Swift, how about I define them in `...ServiceRemote+Subscription.swift`?

momo-ozawa:
@leandroalonso 
Also, just want to make sure about versioning... does `4.15.0-beta.1` sound good? 
Since we're adding functionality, and it's not just a bugfix?

WordPressKit/ReaderPostServiceRemote.h
momo-ozawa: > Also, wdyt about adding some unit tests to the new endpoints? :)

Will do! 👍

leandroalonso: > because the like/unlike endpoints are also defined there

Hmm, in that case let's just keep it where is is then! :)

momo-ozawa: Added tests! 8922f20

leandroalonso:
@momo-ozawa, regarding:

>  Also, just want to make sure about versioning... does 4.15.0-beta.1 sound good?

Yes!  `4.15.0-beta.1` sounds good!

WordPressKit/ReaderPostServiceRemote.h
momo-ozawa: Fixed! 6e4ec66

momo-ozawa:
@leandroalonso 
Fixed an issue where isSubscribed as always returning `false`! 9d3b17e
Will update WPKit in the WP PRs 👍

PR #271, opened 2020-08-18
Add refresh counter to the fetchCards method

emilylaguna:
### Description

Adds a refresh parameter to the fetchCards method to allow for random posts on pull to refresh.

### Testing Details

You can test this with: https://github.com/wordpress-mobile/WordPress-iOS/pull/14667

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.14.0"
+  s.version       = "4.14.1-beta.1"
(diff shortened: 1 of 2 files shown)


leandroalonso (approved):

WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.14.0"
+  s.version       = "4.14.1-beta.1"
```
leandroalonso, line 3: This is not a bugfix, so `"4.15.0-beta.1"` makes more sense IMHO

PR #280, opened 2020-09-04
Upgrade Xcode to 12.0

jkmassel:
### Description

Makes changes to the project required to support Xcode 12, including:

- Fixes an error when building caused by an old version of CocoaLumberjack
- Fixes a compilation failure
- Fixes many warnings about modular includes using quotes instead of angle brackets
- Updates CocoaPods to support Xcode 12


WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -51,12 +51,20 @@ open class WordPressOrgXMLRPCApi: NSObject {
         let sessionManager = Alamofire.SessionManager(configuration: sessionConfiguration)
 
         let sessionDidReceiveChallengeWithCompletion: ((URLSession, URLAuthenticationChallenge, @escaping(URLSession.AuthChallengeDisposition, URLCredential?) -> Void) -> Void) = { [weak self] session, authenticationChallenge, completionHandler in
-            return self?.urlSession(session, didReceive: authenticationChallenge, completionHandler: completionHandler)
+            guard let strongSelf = self else {
```
bjtitus, line 54: @jkmassel I think it's sufficient to not return the result here. The return type is `Void` anyway.

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -51,12 +51,20 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
+                return
+            }
+
+            return strongSelf.urlSession(session, didReceive: authenticationChallenge, completionHandler: completionHandler)
```
bjtitus: @jkmassel I think it's sufficient to not return the result here. The return type is `Void` anyway.

This means we don't need the `guard let strongSelf = self` above either.

jkmassel: Yep you're right!

Thanks @bjtitus!!

bjtitus (approved):
Build and ran using Xcode 12 and did not see any warnings.

:shipit:

jkmassel:
Thanks @bjtitus!!

shiki:
Thanks for fixing this!

PR #296, opened 2020-10-25
Feature/remote content hash

Deco354:
### Description
Adds a hash to `RemotePost` so we can identify when a remote posts content has been updated.

A lot of this logic was already present in `AbstractPost` so the non specialized parts of the hash have been extracted so any client of WordPressKit can generate a hash. It's also been adapted to be faster seeing as this hash will not be uploaded nor need to meet cryptographic standards but will be ran for every post the server sends and will block the user if it runs slow. `AbstractPost` will still be able to use the slower better collision proof version of this hash.

### Fixes

Part of [Conflict Resolution Dialog](https://github.com/wordpress-mobile/WordPress-iOS/issues/14798).

At present we use `dateModified` to identify if a remote post has changed which is not fully reliable.

### Testing Details

Running the tests and analyzing that they do indeed test that this is functional should suffice.

### DIFF
--- a/WordPressKit/RemotePost.m
+++ b/WordPressKit/RemotePost.m
@@ -50,7 +50,9 @@ NSString * const PostStatusDeleted = @"deleted"; // Returned by wpcom REST API w
 
 /// A hash used to determine if the remote content has changed.
 ///
-/// This hash must remain constant regardless of iOS version, app restarts or instances used
+/// This hash must remain constant regardless of iOS version, app restarts or instances used. `Hasher` or NSObject's `hash` were not used for these reasons.
+///
+/// - Note: `dateModified` is not included within the hash as it is prone to change wihout the content having been changed and is the reason this hash is necessary.
 - (NSString *)contentHash
 {
     NSString *hashedContents = [NSString stringWithFormat:@"%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@",
--- a/WordPressKit/RemotePost.m
+++ b/WordPressKit/RemotePost.m
@@ -52,7 +52,12 @@ NSString * const PostStatusDeleted = @"deleted"; // Returned by wpcom REST API w
 ///
 /// This hash must remain constant regardless of iOS version, app restarts or instances used. `Hasher` or NSObject's `hash` were not used for these reasons.
 ///
-/// - Note: `dateModified` is not included within the hash as it is prone to change wihout the content having been changed and is the reason this hash is necessary.
+/// `dateModified` is not included within the hash as it is prone to change wihout the content having been changed and is the reason this hash is necessary.
+///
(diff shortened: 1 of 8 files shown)


WordPressKit/SHAHasher.m
```diff
@@ -0,0 +1,81 @@
...
+
++ (NSData *)hashForBool:(BOOL)boolean {
+    unsigned char digest[CC_SHA256_DIGEST_LENGTH];
+
+    CC_SHA256(&boolean, sizeof(boolean), digest);
+
+    return [NSData dataWithBytes:digest length:CC_SHA256_DIGEST_LENGTH];
+}
+
++ (NSString *)sha256StringFromData:(NSData *)data {
```
diegoreymendez: I think this method is poorly named (I know it existed already) because it doesn't really do anything in terms of SHA256.  As it is right now, this is more like `hexStringFromData`.

WordPressKit/RemotePost.m
```diff
@@ -47,4 +48,48 @@ - (NSDictionary *)debugProperties {
...
+///
+/// `dateModified` is not included within the hash as it is prone to change wihout the content having been changed and is the reason this hash is necessary.
+///
+/// `autosave` properties are intentionally omitted as remote autosaves are always discarded in favor of local autosaves (aka `revision`s)
+///
+/// - Note: At the time of writing the backend will occasionally create updates that neither come from autosaves or user initiated saves, these will modify the hash
+/// and at present are treated as genuine updates as they are triggered by the user changing their posts content.
+- (NSString *)contentHash
+{
+    NSString *hashedContents = [NSString stringWithFormat:@"%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@%@/%@/%@",
```
diegoreymendez, line 63: Considering we have support to hash each component in the methods below, and then combine the hash of each component (through `combineHashes`), is there a reason behind the choice of using a big string concatenating all of the post elements?

diegoreymendez (commented):
Adding some initial feedback.

WordPressKit/RemotePost.m
palringo-dec: Performance mostly.

Hashing every property will likely minimize collisions but it took ~ 0.1s when I performance tested this. I fear this could result in a significant delay when dealing with many posts.

#### Property hash performance
image


For an `AbstractPost` upload I don't think this is a problem as a post upload isn't an action that blocks the user. However for downloading a `RemotePost` update this could result in real user delays as the non-arrival of this update will block them from continued use of the app.

#### ConcatenatedStringHash Performance
After concatenating the strings we're down to < 0.01s

image

I'm still a little bit concerned that this has the potential to be a bit too slow if we're going to be checking large numbers of posts each update. The swift algorithm I wrote earlier this month just performs a bitwise shift operation, this would be much faster than the others if it prove necessary.

Let me know your thoughts, perhaps I've missed something somewhere.
How are posts currently downloaded when in large numbers? Do we download them in batches?

PR #298, opened 2020-11-02
Merge 4.20 into trunk

jkmassel:
Copies changes from the `4.20` release into `trunk`


jkmassel:
Admin-merging this since it's just a release

PR #300, opened 2020-11-09
Upgrade to Alamofire 5.4.0

pmusolino:
### Description
In this PR, I handled the migration of [Alamofire](https://github.com/Alamofire/Alamofire) from version 4 to the major version 5.
Most APIs have changed in Alamofire 5. While most top level request APIs remain the same, nearly every other type has changed in some way.

## Testing
@Gio2018 opened a new PR under `WordPressAuthenticator-iOS` that use this branch. The branch of `WordPressAuthenticator-iOS` should be used by `WPiOS` and `WCiOS` for testing that everything works like before. I'll update the PR with the links to the branches opened for `WPiOS` and `WCiOS` if available.

**What to test?**
- Make sure that you are able to authenticate in the app like before.
- Make sure that every request succeeded.
- Make sure that errors are correctly handled like before.
- Make sure that uploading a new file/image works like before.
- Try to use the app under all their functionality to make sure that nothing changed.
- CI

## Changes:
- Updated the local Podfile for using Alamofire 5.4.0.
- Updated the `podspec` file for using Alamofire 5.4.0. There was also a version upgrade.
- Refactoring: renamed all the references to `Alamofire.SessionManager` that now becomes `Alamofire.Session`.
- Refactoring: some of the `Alamofire` class references becomes `AF`.
- Refactoring: the majority of Alamofire `Result` type has been removed. Alamofire now uses Swift’s `Result` type.
- Fix of a typo: `FailureReponseBlock` -> `FailureResponseBlock`
- `MultipartFormData`’s API has changed and the top level `upload` methods to create and upload `MultipartFormData` have been updated to match other request APIs, so it’s no longer necessary to deal with the `Result` of the multipart encoding. This means that I updated consequentially the method used by `WordPressComRestApi`.
- Alamofire now returns `AFError` for all of its APIs, wrapping any underlying system or custom APIs in `AFError` instances. Furthermore, AFError now has several new cases, so switching over it exhaustively has been handled. I updated the `processError` method under `WordPressComRestApi` and I also implemented a `getNSError()` method under a `AFError` private extension for getting the correct error for every case.
- I removed the `onChallenge` callback because the authentication system changed in Alamofire 5 and the methods that we were using previously are no more available, in fact, the `delegate` property of the `sessionManager` doesn’t contain anymore some completion handler that we were using, like `sessionDidReceiveChallengeWithCompletion` or `taskDidReceiveChallengeWithCompletion`. @Gio2018 is trying to reintroduce it since it is used by `WPiOS`. Since the authentication system changed, I also updated accordingly our `makeSessionManager` method.
- Under `Authenticator`, I handled all the changes introduced with the new request adapter and retrier and I updated the tests.


WordPressKit/WordPressComRestApi.swift
```diff
@@ -323,34 +323,29 @@ open class WordPressComRestApi: NSObject {
...
-                    dataRequest.cancel()
-                }
-            case .failure(let encodingError):
-                failure(encodingError as NSError, nil)
+        }, to: URLString)
+        .uploadProgress(closure: progressUpdater)
+        .validate()
+        .responseJSON { (dataResponse) in
+            switch dataResponse.result {
+            case .success(_):
```
Gio2018, line 335: here is where I am getting the crash. Maybe we should still pass the value associated with `success` to the `success(_:, _:)`  closure? I did not do extensive tests but doing so appears to solve the crash on `WPiOS`

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -51,23 +43,15 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
-
-        let sessionDidReceiveChallengeWithCompletion: ((URLSession, URLAuthenticationChallenge, @escaping(URLSession.AuthChallengeDisposition, URLCredential?) -> Void) -> Void) = { [weak self] session, authenticationChallenge, completionHandler in
-            self?.urlSession(session, didReceive: authenticationChallenge, completionHandler: completionHandler)
-        }
-        sessionManager.delegate.sessionDidReceiveChallengeWithCompletion = sessionDidReceiveChallengeWithCompletion
-
-        let  taskDidReceiveChallengeWithCompletion: ((URLSession, URLSessionTask, URLAuthenticationChallenge,  @escaping(URLSession.AuthChallengeDisposition, URLCredential?) -> Void) -> Void) = { [weak self] session, task, authenticationChallenge, completionHandler in
-            self?.urlSession(session, task: task, didReceive: authenticationChallenge, completionHandler: completionHandler)
-        }
-        sessionManager.delegate.taskDidReceiveChallengeWithCompletion = taskDidReceiveChallengeWithCompletion
```
Gio2018, line 70: this would need to be reinstated, as in `WPiOS` we do need the authentication challenge for self-hosted sites that use it

Gio2018 (commented):
Still need to do a complete review, but added a couple of comments (some of them might be redundant with what already discussed on Slack)

mokagio:
Closing because of @crazytonyli Alamofire removal work, see #736

PR #302, opened 2020-11-10
Add additional parameters in getActivityForSite

leandroalonso:
Added 3 new parameters in getActivityForSite:

* `after`
* `before`
* `group`

These will be used in the new filters for Activity Log in WPiOS. Documentation about each parameter can be found in the code.


emilylaguna (approved):
Looks good! 🚢

PR #304, opened 2020-11-17
Adds in Jetpack Scan API with models

emilylaguna:
### Description
🙇‍♀️ I apologize this PR got bigger than I originally intended. The threat models ended up being a lot more complex.

There isn't a related WPiOS PR yet, so the tests being green and a thorough code review will work 👍  

This PR introduces the `JetpackScanServiceRemote` class that contains the method for starting a scan and getting scan information (status, availability, threats, etc). Since the API only uses 1 endpoint for getting scan information (`sites/SITE_ID/scan`) I've added a few "helper" methods to reduce the amount of information returned by the `getScanForSite` method:

- `getScanAvailableForSite` just returns the `isEnabled` bool
- `getCurrentScanStatusForSite` just returns the current scan status if it exists
- `getThreatsForSite` just returns the threats array

#### Threats
I had originally planned on creating sub threat objects like `JetpackThreatCore`, `JetpackThreatDatabase` and only including the relevant properties for that type, but I couldn't get that to work within the decodable structure. Looking for suggestions to make this a bit cleaner.  

The `JetpackScanThreat` object is A LOT to look at so I'll provide any relevant information I can think of:
- The API does not return the threat type so we have to figure that out ourselves. Luckily [wp-calypso](https://github.com/Automattic/wp-calypso/blob/5a6b257ad97b361fa6f6a6e496cbfc0ef952b921/client/components/jetpack/threat-item/utils.ts#L11) already had this logic.
- The fixable property points to a "Fixer" object that defines how a threat will be fixed if it can be. This information will be used to generate a string later such as "Jetpack will update this plugin to version 1.3.3.7"
    - If a threat can not be fixed this object is nil. 

- `JetpackThreatContext`:
  - This defines a code preview for file based threats with highlighted sections
  - was created with the intent of using it later to generate an NSAttributedString which is why I opted to use NSRange's
  - The value from the API can be:
     - Not present for non-file threats
     - An empty string `context: ""`
     - A dictionary
   - The logic for this is kind of messy 😑 looking for suggestions to improve.

### Testing Details
🟢  Tests should be good!


WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /// Whether the scan feature is available or not
+    public var isEnabled: Bool = false
+
+    /// The state of the current scan
+    public var state: JetpackScanState = .idle
+
+    /// If there is a scan in progress, this will return its status
+    public var current: JetpackScanStatus? = nil
```
leandroalonso:
```suggestion
    public var current: JetpackScanStatus?
```

WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    /// Scan Status for the most recent scan
+    /// This will be nil if there is currently a scan taking place
+    public var mostRecent: JetpackScanStatus? = nil
```
leandroalonso:
```suggestion
    public var mostRecent: JetpackScanStatus?
```

WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    /// An array of the current threats
+    /// During a scan this will return the previous scans threats
+    public var threats: [JetpackScanThreat]? = nil
```
leandroalonso:
```suggestion
    public var threats: [JetpackScanThreat]?
```

WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    /// A limited representation of the users credientals for each role
+    public var credentials: [JetpackScanCredentials]? = nil
```
leandroalonso:
```suggestion
    public var credentials: [JetpackScanCredentials]?
```

WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,95 @@
...
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        isInitial = try container.decode(Bool.self, forKey: .isInitial)
+        progress = try container.decode(Int.self, forKey: .progress)
+        startDate = try container.decode(ISO8601Date.self, forKey: .timestamp)
+        duration = try container.decodeIfPresent(TimeInterval.self, forKey: .duration)
+
+        if duration == nil, let date = startDate {
+            duration = NSDate().timeIntervalSince(date)
```
leandroalonso: Why `NSDate`?

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,294 @@
...
+    public let status: ThreatStatus?
+
+    /// The date the threat was fixed on
+    public let fixedOn: Date?
+
+    /// More information if the threat is a extension type (plugin or theme)
+    public let `extension`: JetpackThreatExtension?
+
+    /// The type of threat this is
+    private(set) var type: ThreatType
```
leandroalonso:
```suggestion
    private(set) var type: ThreatType = .unknown
```

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,294 @@
...
+    public let diff: String?
+
+    // Database threats will contain row information
+    public let rows: [String: Any]?
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        // Define a placeholder type
+        type = .unknown
```
leandroalonso:
```suggestion
```

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,294 @@
...
+        case edit
+
+        case unknown
+    }
+
+    /// The suggested threat fix type
+    public let type: ThreatFixType
+
+    /// The file path of the file to be fixed
+    public var file: String? = nil
```
leandroalonso:
```suggestion
    public var file: String?
```

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,294 @@
...
+
+    /// The target version to fix to
+    public var target: String? = nil
```
leandroalonso:
```suggestion
    public var target: String?
```

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case type
+        case slug
+        case name
+        case version
+        case isPremium
+    }
+}
```
leandroalonso:
```suggestion
/// Represents plugin or theme additional metadata
public struct JetpackThreatExtension: Decodable {
    public enum JetpackThreatExtensionType: String, Decodable {
        case plugin
        case theme

        case unknown
    }

    public let slug: String
    public let name: String
    public let type: JetpackThreatExtensionType
    public let isPremium: Bool
    public let version: String
}
```

This same logic can be used to simplify other entities I guess.

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,294 @@
...
+
+public struct JetpackThreatContext {
+    public struct ThreatContextLine {
+        var lineNumber: Int
+        var contents: String
+        var highlights: [NSRange]? = nil
```
leandroalonso:
```suggestion
        var highlights: [NSRange]?
```

WordPressKit/JetpackScanServiceRemote.swift
```diff
@@ -0,0 +1,73 @@
...
+        }, failure: failure)
+    }
+
+    public func getCurrentScanStatusForSite(_ siteID: Int, success: @escaping(JetpackScanStatus?) -> Void, failure: @escaping(Error) -> Void) {
+        getScanForSite(siteID, success: { scan in
+            success(scan.current)
+        }, failure: failure)
+    }
+
+    public func getThreatsForSite(_ siteID: Int, success: @escaping([JetpackScanThreat]?) -> Void, failure: @escaping(Error) -> Void) {
```
leandroalonso, line 19: Just a question about those methods: I understand (and completely agree) with the separation here. But this wouldn't cause the app to perhaps make 2 or 3 requests to retrieve the same set of data?

leandroalonso (changes requested):
@emilylaguna a few suggestions to reduce boilerplate code. :)

WordPressKit/JetpackScanServiceRemote.swift
emilylaguna: It would if someone uses the same methods multiple times on the same view. I don't envision that happening since each of the methods should be used on separate screens.

WordPressKit/Jetpack Scan/JetpackScan.swift
emilylaguna: Didn't think it makes much of a difference  since I'm just calculating a time interval. I've switched it to Date()

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
emilylaguna: So I can't remove this one, or else the the last line `type = ThreatType(threat: self)` won't work since I need to pass self to it, but it will throw `Variable 'self.type' used before being initialized`.

I tried working around it but wasn't able to. Any ideas?

emilylaguna:
@leandroalonso Good idea about reducing boilerplate. Can you take another look?

WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,90 @@
...
+        case provisioning
+        
+        case unknown
+    }
+
+    /// Whether the scan feature is available or not
+    public var isEnabled: Bool = false
+
+    /// The state of the current scan
+    public var state: JetpackScanState
```
leandroalonso:
```suggestion
    public var state: JetpackScanState = .unknown
```

WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,90 @@
...
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        state = (try? container.decode(JetpackScanState.self, forKey: .state)) ?? .unknown
+        isEnabled = (state != .unavailable) && (state != .unknown)
+        current = try? container.decode(JetpackScanStatus.self, forKey: .current)
+        mostRecent = try? container.decode(JetpackScanStatus.self, forKey: .mostRecent)
+        threats = try? container.decode([JetpackScanThreat].self, forKey: .threats)
+        credentials = try? container.decode([JetpackScanCredentials].self, forKey: .credentials)
+    }
```
leandroalonso:
```suggestion
```

WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,90 @@
...
+    /// If there was an error finishing the scan
+    /// This will only be available for past scans
+    public var error: Bool?
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        isInitial = try container.decode(Bool.self, forKey: .isInitial)
+        progress = try container.decode(Int.self, forKey: .progress)
+        startDate = try container.decode(ISO8601Date.self, forKey: .timestamp)
```
leandroalonso: You can also reduce the amount of boilerplate code here by doing something like:

```swift
/**Custom decoder for dates*/
class DecoderDates: JSONDecoder {

    override func decode<T>(_ type: T.Type, from data: Data) throws -> T where T : Decodable {
        let decoder = JSONDecoder()
        let dateFormatter = DateFormatter()
        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS"
        decoder.dateDecodingStrategy = .formatted(dateFormatter)
        return try decoder.decode(T.self, from: data)
    }

}
```

And then adding `duration` logic in a `didSet`. You will be able to remove the whole `init`. :)

Just a suggestion though, not a strong opinion.

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,263 @@
...
+        case replace
+        case delete
+        case update
+        case edit
+
+        case unknown
+    }
+
+    /// The suggested threat fix type
+    public let type: ThreatFixType
```
leandroalonso, line 141:
```suggestion
    public var type: ThreatFixType = .unknown
```

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,263 @@
...
+    /// The target version to fix to
+    public var target: String?
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        type = (try? container.decode(ThreatFixType.self, forKey: .type)) ?? .unknown
+        file = try container.decodeIfPresent(String.self, forKey: .file)
+        target = try container.decodeIfPresent(String.self, forKey: .target)
+    }
```
leandroalonso:
```suggestion
```

leandroalonso (changes requested):
@emilylaguna a few more suggestions to remove some additional lines. :)

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
emilylaguna: From my understanding this wouldn't work because I need the `type` to default to `unknown` if the value of `type` is outside the enum values. 

That's why I left the boilerplate code in that I did, because I need to manually set some of the properties up.

Thoughts?

WordPressKit/Jetpack Scan/JetpackScan.swift
emilylaguna: Same comment as: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304#discussion_r527217475

WordPressKit/Jetpack Scan/JetpackScan.swift
emilylaguna: I don't think this would work because I need to be able to differentiate between ISO8601 with and without milliseconds since both are sent from the server.

WordPressKit/Jetpack Scan/JetpackScan.swift
```diff
@@ -0,0 +1,90 @@
...
+
+    /// The progress of the scan from 0 - 100
+    public var progress: Int
+
+    /// How long the scan took / is taking
+    public var duration: TimeInterval?
+
+    /// If there was an error finishing the scan
+    /// This will only be available for past scans
+    public var error: Bool?
```
yaelirub: how about didFail instead of error?

WordPressKit/Jetpack Scan/JetpackScan.swift
yaelirub: >I need to be able to differentiate between ISO8601 with and without milliseconds

@emilylaguna , just curious, why?

emilylaguna: For the scan status the server returns an ISO8601 formatted date, but for threats it returns ISO8601 dates with milliseconds. Which is why I have both `ISO8601Date` and `ISO8601WithMSDate`

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
yaelirub: @emilylaguna , have you read this: https://www.latenightswift.com/2019/02/04/unknown-enum-cases/ or similar? I also wonder if we can reduce some of the decoder boilerplate code in this PR

WordPressKit/Jetpack Scan/JetpackScan.swift
yaelirub: Does it fail to parse if you use the decode without milliseconds?

emilylaguna: Yep.

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
emilylaguna: @yaelirub awesome, I have not seen that. I've gone ahead and implemented that in.

WordPressKit/Jetpack Scan/JetpackScan.swift
emilylaguna: Done!

emilylaguna:
@leandroalonso was able to switch to using the `apiDecoder` which handles the multiple dates formats. I was then able to remove a lot of boilerplate except for `JetpackScanThreat`. 

I also added the `UnknownCaseRepresentable` handling

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
```diff
@@ -0,0 +1,253 @@
...
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        id = try container.decode(Int.self, forKey: .id)
+        signature = try container.decode(String.self, forKey: .signature)
+        fileName = try container.decodeIfPresent(String.self, forKey: .fileName)
+        description = try container.decode(String.self, forKey: .description)
+        firstDetected = try container.decode(Date.self, forKey: .firstDetected)
+        fixedOn = try container.decodeIfPresent(Date.self, forKey: .fixedOn)
+        fixable = try? container.decodeIfPresent(JetpackScanThreatFixer.self, forKey: .fixable) ?? nil
```
leandroalonso, line 96:
```suggestion
        fixable = try? container.decodeIfPresent(JetpackScanThreatFixer.self, forKey: .fixable)
```

leandroalonso (approved):

WordPressKit/Jetpack Scan/JetpackScanThreat.swift
emilylaguna: I left that in because the value can either be `false` or an `object` so I'm using `try?` to allow me to default to nil and prevent a decoding failure because the value isn't an object.

PR #306, opened 2020-11-27
Add get activity groups endpoint

momo-ozawa:
### Description

- Added a new endpoint to [get available activity groups](https://github.com/Automattic/host-partner-documentation/blob/master/jetpack/jetpack-activity-endpoints/activity-count.md) (i.e. activity filter types)
- Added tests for the new endpoint

This new endpoint will be used to get the available **Activity Types** in the Activity Log. Documentation about the parameters can be found in the code.

### Testing Details
🟢  Tests should be good!


WordPressKit/ActivityServiceRemote.swift
```diff
@@ -148,5 +180,22 @@ private extension ActivityServiceRemote {
...
     }
+    
+    func mapActivityGroupsResponse(_ response: AnyObject) throws -> ([ActivityGroup]) {
+        
+        guard let json = response as? [String: AnyObject],
+              let rawGroups = json["groups"] as? [String: AnyObject] else {
+                throw ActivityServiceRemote.ResponseError.decodingFailure
+        }
+        
+        let groups: [ActivityGroup] = try rawGroups.map { (_, value) -> ActivityGroup in
```
momo-ozawa: Looking at Calypso, it seems like the `Activity Type` filters are displaying `ActivityGroup.name` (_e.g. Posts and Pages, Media_) so I think we can ignore the dynamic key names in the `groups` dict (_e.g. post, attachment_).
```
{
  "groups": {
    "post": {
      "name": "Posts and Pages",
      "count": 10
    },
    "attachment": {
      "name": "Media",
      "count": 5
    },
    ...
  }
}
```
Screen Shot 2020-11-27 at 17 06 42

WordPressKit/ActivityServiceRemote.swift
```diff
@@ -67,6 +67,38 @@ open class ActivityServiceRemote: ServiceRemoteWordPressComREST {
...
+    /// Retrieves activity groups associated with a site.
+    ///
+    /// - Parameters:
+    ///     - siteID: The target site's ID.
+    ///     - success: Closure to be executed on success.
+    ///     - failure: Closure to be executed on error.
+    ///
+    /// - Returns: An array of available activity groups for a site.
+    ///
+    open func getActivityGroupsForSite(_ siteID: Int,
```
leandroalonso, line 87: We also need two arguments here: `before` and `after`, just like the getActivities. When the user filters by a date range we pass this into this request so only groups inside the given range are returned.

Another thing is that if we pass just a `after` OR `before` we should make the request using `on`. Eg.:

`https://public-api.wordpress.com/wpcom/v2/sites/165243437/activity/count/group?_envelope=1&on=2020-11-17%2003%3A00%3A00&number=1000`

WordPressKit/Activity.swift
```diff
@@ -158,6 +158,30 @@ public struct ActivityStatus {
     public static let warning = "warning"
 }
 
+public class ActivityGroup {
+    public let name: String
```
leandroalonso, line 163: We also need to store the key here. Eg.: for "Posts and Pages" the key is "post", which is the one we use to actually filters results. Eg.:

`https://public-api.wordpress.com/wpcom/v2/sites/165243437/activity?_envelope=1&aggregate=true&on=2020-11-17%2003%3A00%3A00&group%5B%5D=post&number=1000`

WordPressKit/ActivityServiceRemote.swift
leandroalonso: I added a comment about the key just above. :)

leandroalonso (changes requested):
@momo-ozawa added a few comments. Basically, we need to:

* Add date range parameters
* Add the key value to the model (which is used to filter the results)

WordPressKit/ActivityServiceRemote.swift
momo-ozawa: Fixed! 6a7698b

WordPressKit/Activity.swift
momo-ozawa: Thanks for the hint 😄 
Fixed! 5b7f2ed

leandroalonso (approved):
:shipit:

PR #309, opened 2020-12-03
Fix JetpackScan model parsing errors

emilylaguna:
### Description

This PR fixes some issues while parsing the scan data: 
- The Jetpack Scan credentials may not include all the fields so I made them optional
- The context parsing was throwing when `context: ""` (context is an empty string)
- `isEnabled` is always false because `didSet` is not called on initialization

You can test it using the following WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15427

### Testing Details
Follow the testing steps on the PR here: https://github.com/wordpress-mobile/WordPress-iOS/pull/15427

### DIFF
--- a/WordPressKit/Jetpack Scan/JetpackCredentials.swift
+++ b/WordPressKit/Jetpack Scan/JetpackCredentials.swift
@@ -2,10 +2,10 @@ import Foundation
 
 /// A limited representation of the users Jetpack credentials
 public struct JetpackScanCredentials: Decodable {
-    public let host: String
-    public let port: Int
-    public let user: String
-    public let path: String
+    public let host: String?
+    public let port: Int?
+    public let user: String?
+    public let path: String?
     public let type: String
     public let role: String
     public let stillValid: Bool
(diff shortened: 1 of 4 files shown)


WordPressKit/Jetpack Scan/JetpackCredentials.swift
```diff
@@ -2,10 +2,10 @@ import Foundation
...
 public struct JetpackScanCredentials: Decodable {
-    public let host: String
-    public let port: Int
-    public let user: String
-    public let path: String
+    public let host: String?
+    public let port: Int?
+    public let user: String?
+    public let path: String?
     public let type: String
```
leandroalonso, line 9: Are you sure these others will always be present? Just wondering if we shouldn't change them to Optionals too...

leandroalonso (approved):

WordPressKit/Jetpack Scan/JetpackCredentials.swift
emilylaguna: From what I can tell the `type`, `role`, `stillValid` are always present.

PR #311, opened 2020-12-04
Activity log: fix activies request when using multiple groups

leandroalonso:
### Description

This PR changes how the request to retrieve activities is made to split the groups into multiple `groups[]` params.

It also fixes a decoding error from the group's endpoint that should be treated as empty results.

You can test the PR here: https://github.com/wordpress-mobile/WordPress-iOS/pull/15440

### DIFF
--- a/WordPressKit/ActivityServiceRemote.swift
+++ b/WordPressKit/ActivityServiceRemote.swift
@@ -42,7 +42,7 @@ open class ActivityServiceRemote: ServiceRemoteWordPressComREST {
 
         path?.queryItems = group.map { URLQueryItem(name: "group[]", value: $0) }
 
-        let pageNumber = (offset / count + 1)
+        let pageNumber = (offset / count) + 1
         path?.queryItems?.append(URLQueryItem(name: "number", value: "\(count)"))
         path?.queryItems?.append(URLQueryItem(name: "page", value: "\(pageNumber)"))
 
--- a/WordPressKit/ActivityServiceRemote.swift
+++ b/WordPressKit/ActivityServiceRemote.swift
@@ -47,7 +47,7 @@ open class ActivityServiceRemote: ServiceRemoteWordPressComREST {
         path?.queryItems?.append(URLQueryItem(name: "page", value: "\(pageNumber)"))
 
         if let after = after, let before = before,
-           let lastSecondOfBeforeDay = Calendar.current.date(byAdding: .second, value: 86399, to: before) {
+           let lastSecondOfBeforeDay = before.endOfDay() {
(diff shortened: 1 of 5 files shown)


WordPressKit/ActivityServiceRemote.swift
```diff
@@ -37,28 +37,31 @@ open class ActivityServiceRemote: ServiceRemoteWordPressComREST {
...
+        var path = URLComponents(string: "sites/\(siteID)/activity")
+
+        path?.queryItems = group.map { URLQueryItem(name: "group[]", value: $0) }
+
+        let pageNumber = (offset / count + 1)
+        path?.queryItems?.append(URLQueryItem(name: "number", value: "\(count)"))
+        path?.queryItems?.append(URLQueryItem(name: "page", value: "\(pageNumber)"))
 
         if let after = after, let before = before,
            let lastSecondOfBeforeDay = Calendar.current.date(byAdding: .second, value: 86399, to: before) {
```
emilylaguna: I know this isn't new, but I just had a thought about moving moving `86399` to an extension since it's used in a couple places? 

`Calendar.current.endOfDay(for date: Date) -> Date?`

WordPressKit/ActivityServiceRemote.swift
```diff
@@ -37,28 +37,31 @@ open class ActivityServiceRemote: ServiceRemoteWordPressComREST {
...
-        ]
 
-        if !group.isEmpty {
-            parameters["group[]"] = group.joined(separator: ",") as AnyObject
-        }
+        var path = URLComponents(string: "sites/\(siteID)/activity")
+
+        path?.queryItems = group.map { URLQueryItem(name: "group[]", value: $0) }
+
+        let pageNumber = (offset / count + 1)
```
emilylaguna: I was a bit confused about the intention of the calculation, what about shifting the parentheses to make it a bit more clear?

```suggestion
        let pageNumber = (offset / count) + 1
```

emilylaguna (changes requested):
Nice, working well. Had a couple ideas below. Let me know your thoughts.

WordPressKit/ActivityServiceRemote.swift
leandroalonso: Done in 1e13a77d9a5d1809425b611ea392e19d87d84e7d

WordPressKit/ActivityServiceRemote.swift
leandroalonso: Done in f3bda1433353ce9f59a01f942d5ce0eea6e7e881

emilylaguna (approved):
LGTM! 🚢

PR #327, opened 2021-01-13
Add support for marking Reader posts seen/unseen

ScoutHarris:
### Description

Ref https://github.com/wordpress-mobile/WordPress-iOS/issues/15355
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15638

This adds functionality to toggle a post's `is_seen` status.

### Testing Details


WordPressKit/ReaderPostServiceRemote+V2.swift
```diff
@@ -44,4 +44,43 @@ extension ReaderPostServiceRemote {
...
+        let path = self.path(forEndpoint: endpoint, withVersion: ._2_0)
+        
+        let params = [
+            "feed_id": feedID,
+            "feed_item_ids": [feedItemID],
+            "source": "reader-ios"
+        ] as [String: AnyObject]
+        
+        wordPressComRestApi.POST(path, parameters: params, success: { (responseObject, httpResponse) in
+            success()
```
aerych, line 79: Probably we should check what's returned in the response object.  It should be a dictionary with a single key `status`.  The value of status can be true or false depending on if the call succeeded.  Looks like there are a few (unlikely) scenarios where false could be returned.

aerych (changes requested):
Hola @ScoutHarris 👋 
I had just one nitpick looking at the code. 
Back to you!

WordPressKit/ReaderPostServiceRemote+V2.swift
ScoutHarris: Good catch. And done!

aerych (approved):
LGTM! :shipit:

ScoutHarris:
Thanks @aerych !

PR #337, opened 2021-01-27
Jetpack Scan: Add rollback case to ThreatFixType 

momo-ozawa:
### Description
- Added rollback case to ThreatFixType

### Testing Details
Test should be 🟢


emilylaguna (approved):
LGTM!

emilylaguna:
Just a heads up looks like a new WPKit version was pushed so you'll need to bump the version before releasing.

PR #338, opened 2021-01-28
Include parameters in multipart POST request

### LINKED ISSUE #333: Parameters are not included in multipart POST requests
### Expected behavior

Parameters argument of [`multipartPOST`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/WordPressComRestApi.swift#L308) function should be included in the POST request.

### Actual behavior

Parameters are not included.

`multipartPOST` function is only used in the following places:
- `MediaServiceRemoteREST.uploadProgress` ([one item version](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/MediaServiceRemoteREST.m#L209), [multiple items version](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/MediaServiceRemoteREST.m#L148))
- [`PostServiceRemoteREST.createPost with media`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/PostServiceRemoteREST.m#L167) (looks like this function is not currently being used)

The easiest way to reproduce the bug is by adding items from the free photo library as they include as the caption value its photo credit in the upload parameters.

1. Go to the media screen.
2. Tap on `+` button.
3. Tap on `Free Photo Library`.
4. Search any term.
5. Tap one or multiple items.
6. Tap on `Add` button.
7. Wait for the items to be uploaded.
8. Tap on one of the uploaded items.
9. Check that the caption value is empty.

##### WordPressKit Environment
 - WordPressKit Version 4.25-beta
 - OS: iOS 14.0
 - Device: iPhone 8 (simulator)
 - Xcode Version: Xcode 12.0

##### Additional Context

No.

fluiddot:
### Description

Fixes https://github.com/wordpress-mobile/WordPressKit-iOS/issues/333

Unfortunately Alamofire doesn't provide a way to serialise the parameters object to be included in the `multipartFormData` object, as a workaround I'm only adding the `String` and `Int` values that are converted to `Data`.

### Testing Details

The easiest way to test this is by adding items from the free photo library as they include as the caption value its photo credit in the upload parameters.

1. Go to the media screen.
2. Tap on `+` button.
3. Tap on `Free Photo Library`.
4. Search any term.
5. Tap one or multiple items.
6. Tap on `Add` button.
7. Wait for the items to be uploaded.
8. Tap on one of the uploaded items.
9. Check that the caption value is defined.

### DIFF
--- a/WordPressKit/PostServiceRemoteREST.m
+++ b/WordPressKit/PostServiceRemoteREST.m
@@ -160,8 +160,8 @@ static NSString * const RemoteOptionValueOrderByPostID = @"ID";
 
     
     BodyPart *contentPart =[[BodyPart alloc] initWithName:@"content" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
-    BodyPart *titlePart = [[BodyPart alloc] initWithName:@"title" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
-    BodyPart *statusPart = [[BodyPart alloc] initWithName:@"status" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
+    BodyPart *titlePart = [[BodyPart alloc] initWithName:@"title" data:[post.title dataUsingEncoding:NSUTF8StringEncoding]];
+    BodyPart *statusPart = [[BodyPart alloc] initWithName:@"status" data:[post.status dataUsingEncoding:NSUTF8StringEncoding]];
     BodyPart *mediaPart = [[BodyPart alloc] initWithName:@"media[]" url:media.localURL fileName:filename mimeType:type];
     
     [self.wordPressComRestApi multipartPOST:requestUrl
--- a/WordPressKit/PostServiceRemoteREST.m
+++ b/WordPressKit/PostServiceRemoteREST.m
@@ -158,14 +158,15 @@ static NSString * const RemoteOptionValueOrderByPostID = @"ID";
     NSString *requestUrl = [self pathForEndpoint:path
                                      withVersion:ServiceRemoteWordPressComRESTApiVersion_1_2];
 
-    NSMutableDictionary *parameters = [NSMutableDictionary dictionaryWithDictionary:@{}];
-    parameters[@"content"] = post.content;
-    parameters[@"title"] = post.title;
-    parameters[@"status"] = post.status;
-    FilePart *filePart = [[FilePart alloc] initWithParameterName:@"media[]" url:media.localURL filename:filename mimeType:type];
+    
+    BodyPart *contentPart =[[BodyPart alloc] initWithName:@"content" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
+    BodyPart *titlePart = [[BodyPart alloc] initWithName:@"title" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
+    BodyPart *statusPart = [[BodyPart alloc] initWithName:@"status" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
+    BodyPart *mediaPart = [[BodyPart alloc] initWithName:@"media[]" url:media.localURL fileName:filename mimeType:type];
+    
     [self.wordPressComRestApi multipartPOST:requestUrl
-                                 parameters:parameters
-                                  fileParts:@[filePart]
+                                 parameters:nil
+                                  bodyParts:@[contentPart, titlePart, statusPart, mediaPart]
                             requestEnqueued:^(NSNumber *taskID) {
                                 if (requestEnqueued) {
                                     requestEnqueued(taskID);
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -290,11 +290,11 @@ open class WordPressComRestApi: NSObject {
     }
 
     /**
-     Executes a multipart POST using the current serializer, the parameters defined and the fileParts defined in the request
+     Executes a multipart POST to the specified endpoint defined on URLString, including the parameters and the fileParts defined in the request.
      This request will be streamed from disk, so it's ideally to be used for large media post uploads.
 
      - parameter URLString:  the endpoint to connect
-     - parameter parameters: the parameters to use on the request
+     - parameter parameters: the parameters to use on the request (only String and Int parameters will be included)
      - parameter fileParts:  the file parameters that are added to the multipart request
      - parameter requestEnqueued: callback to be called when the fileparts are serialized and request is added to the background session. Defaults to nil
      - parameter success:    callback to be called on successful request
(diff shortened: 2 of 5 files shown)


WordPressKit/WordPressComRestApi.swift
```diff
@@ -327,6 +327,15 @@ open class WordPressComRestApi: NSObject {
             for filePart in fileParts {
                 multipartFormData.append(filePart.url, withName: filePart.parameterName, fileName: filePart.filename, mimeType: filePart.mimeType)
             }
+            
+            if let parameters = parameters {
+                for (key, value) in parameters {
+                    if value is String || value is Int {
```
ceyhun: Checking for one of the use cases for https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/media/new/,
I think treating `parameters` as `Query Parameters` and adding another function parameter specifically for `Request Parameters` could be beneficial in case we also want to send `Query Parameters` next to `Request Parameters` in the future.
One other option could be to change the type of `parameters` to `[String: String]?` instead of checking for types here.
Yet another option could be adding another parameter to the method that takes a function like `multipartFormData -> Void` and the caller can deal with how to append those?

fluiddot: Thanks @ceyhun for the feedback!

> I think treating parameters as Query Parameters and adding another function parameter specifically for Request Parameters could be beneficial in case we also want to send Query Parameters next to Request Parameters in the future.

I though about this option but I wasn't sure if having both could be confusing. In regular POST requests we only have `parameters` and they're considered as the `Request Parameters`, in that case `Query Parameters` should be previously added to the URL.

> One other option could be to change the type of parameters to [String: String]? instead of checking for types here.

Yeah that would prevent having to check the types as I did, I included `Int` type because I saw a couple of cases where there were arguments of that type. In any case since we're converting them into `String` when parsing the value to `Data` we could use `[String: String]?` and let the callers explicitly do the conversion.

> Yet another option could be adding another parameter to the method that takes a function like multipartFormData -> Void and the caller can deal with how to append those?

I think this one is probably the best since it will let the caller to include more complex values than primitives. The downside is that callers would have to do the parsing to `Data` but I think that's expected since it's a multipart POST request.

fluiddot: > Yet another option could be adding another parameter to the method that takes a function like multipartFormData -> Void and the caller can deal with how to append those?

I tried to apply option 3 but unfortunately the class `MultipartFormData` that Alamofire provides is not compatible with `ObjC` and the places where this type of request is being used are `ObjC`, however this gave me an idea.

The idea was to unify the `Request Parameters` into one argument, including the file parts which in the end are also body parts. Following this I created a new argument named `bodyParts` that is an array of `BodyPart`, a new class I added (based on the previous one `FilePart`) but more generic that represents a body part of the multipartPOST requests.

I'd appreciate feedback of this [change](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/338/commits/84775760becd4ff34febb9b6c55be4342574e037) since I'm unaware of the implications of such a refactor, thanks!

WordPressKit/PostServiceRemoteREST.m
```diff
@@ -158,14 +158,15 @@ - (void)createPost:(RemotePost *)post
...
 
-    NSMutableDictionary *parameters = [NSMutableDictionary dictionaryWithDictionary:@{}];
-    parameters[@"content"] = post.content;
-    parameters[@"title"] = post.title;
-    parameters[@"status"] = post.status;
-    FilePart *filePart = [[FilePart alloc] initWithParameterName:@"media[]" url:media.localURL filename:filename mimeType:type];
+    
+    BodyPart *contentPart =[[BodyPart alloc] initWithName:@"content" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
+    BodyPart *titlePart = [[BodyPart alloc] initWithName:@"title" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
+    BodyPart *statusPart = [[BodyPart alloc] initWithName:@"status" data:[post.content dataUsingEncoding:NSUTF8StringEncoding]];
```
ceyhun: A small mistake here:
```suggestion
    BodyPart *titlePart = [[BodyPart alloc] initWithName:@"title" data:[post.title dataUsingEncoding:NSUTF8StringEncoding]];
    BodyPart *statusPart = [[BodyPart alloc] initWithName:@"status" data:[post.status dataUsingEncoding:NSUTF8StringEncoding]];
```

WordPressKit/WordPressComRestApi.swift
ceyhun: That's a great idea! But there appears to be a `BodyPart` class in Alamofire as well. Should we maybe rename it?

ceyhun (commented):
Overall the changes look good. Thanks a lot for coming up with this and implementing. I would like to test all the affected paths inside WPiOS to be confident. I saw you mentioned under `Testing Details` to upload a photo from the `Free Photo Library` which should cover the `MediaServiceRemoteREST` parts and I guess testing to upload a local image would be necessary as well. What would be the best way to test `PostServiceRemoteREST` parts?

WordPressKit/WordPressComRestApi.swift
ceyhun: That's a great idea! But there appears to be a `BodyPart` class in Alamofire as well. Should we maybe rename this?

WordPressKit/PostServiceRemoteREST.m
fluiddot: Oh good catch! My bad when doing copy/paste 😄 .

WordPressKit/WordPressComRestApi.swift
fluiddot: Thanks! I'm not familiar of class name conflicts on iOS but yeah if it could produce it let's rename it. I was thinking to use `FormPart` or `RequestPart` 🤔 .

fluiddot:
> Overall the changes look good. Thanks a lot for coming up with this and implementing. I would like to test all the affected paths inside WPiOS to be confident. I saw you mentioned under `Testing Details` to upload a photo from the `Free Photo Library` which should cover the `MediaServiceRemoteREST` parts and I guess testing to upload a local image would be necessary as well. What would be the best way to test `PostServiceRemoteREST` parts?

Yeah sorry, due the latest changes I should update the `Testing Details` because we should cover more cases. For `MediaServiceRemoteREST` we should try to upload different types of media, for `PostServiceRemoteREST` it's a bit tricky because the function that invokes the upload media is not being used anywhere 😄 . So in this case unless we explicitly connect it somewhere in the app we can't test it.

WordPressKit/WordPressComRestApi.swift
ceyhun: I think `FormPart` could be better since it's used in `MultipartFormData`.

fluiddot: Yeah makes sense, I'll rename it then, thanks!

PR #340, opened 2021-01-29
Add sortingOption parameter for fetching cards

bozidarsevo:
### Description

Ref [#14765](https://github.com/wordpress-mobile/WordPress-iOS/issues/14765)

Added `sortingOption` parameter for fetching the cards. It is used for adding a sorting option in the "Discover" tab of the "Reader", giving the possibility of sorting posts either by popularity or by most recent.

### Testing Details

The easiest way to test it is to pass a different value for the `sortingOption` parameter. You can use `date` or `popularity`(which is used by default if nil is passed).
This [PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/15754) is currently using these changes.


WordPressKit/ReaderPostServiceRemote+Cards.swift
```diff
@@ -7,14 +7,16 @@ extension ReaderPostServiceRemote {
     ///
     /// - Parameter topics: an array of String representing the topics
     /// - Parameter page: a String that represents a page handle
+    /// - Parameter sortingOption: a String that represents a sorting option
     /// - Parameter success: Called when the request succeeds and the data returned is valid
     /// - Parameter failure: Called if the request fails for any reason, or the response data is invalid
     public func fetchCards(for topics: [String],
                            page: String? = nil,
+                           sortingOption: String? = nil,
```
Gio2018: Maybe we should use an `enum` (with a `String rawValue`) instead of just passing a string to make it less error prone? One of the cases could also be a `noSorting` option that should then be the default.

WordPressKitTests/ReaderPostServiceRemote+CardsTests.swift
```diff
@@ -93,4 +93,47 @@ class ReaderPostServiceRemoteCardTests: RemoteTestCase, RESTTestable {
...
+        stubRemoteResponse("read/tags/cards?tags%5B%5D=cats&sort=date", filename: "reader-cards-success.json", contentType: .ApplicationJSON)
+        
+        readerPostServiceRemote.fetchCards(for: ["cats"], sortingOption: "date", success: { cards, _ in
+            let posts = cards.filter { $0.type == .post }
+            for i in 1..<posts.count {
+                guard let firstPostDate = posts[i-1].post?.sortDate,
+                      let secondPostDate = posts[i].post?.sortDate,
+                      firstPostDate > secondPostDate else {
+                    XCTFail("Posts should be sorted by date, starting with most recent post")
+                    expect.fulfill()
```
Gio2018: should this expectation be fulfilled here?

WordPressKitTests/ReaderPostServiceRemote+CardsTests.swift
```diff
@@ -93,4 +93,47 @@ class ReaderPostServiceRemoteCardTests: RemoteTestCase, RESTTestable {
...
+    }
+    
+    // Calls the API without the given sorting option
+    //
+    func testCallAPIWithoutTheGivenSortingOption() {
+        let readerPostServiceRemote = ReaderPostServiceRemote(wordPressComRestApi: mockRemoteApi)
+
+        readerPostServiceRemote.fetchCards(for: [], sortingOption: nil, success: { _, _ in }, failure: { _ in })
+        
+        XCTAssertFalse(mockRemoteApi.URLStringPassedIn?.contains("sort=") ?? false)
```
Gio2018: is this condition enough to guarantee that the test actually succeeds? What happens if `mockRemoteApi.URLStringPassedIn` is `nil`? Should we actually check that the `GET` method is called (maybe in other tests too)?

Gio2018 (changes requested):
Hey @bozidarsevo , this looks like a good start! Left a few comments for you to consider.

WordPressKit/ReaderPostServiceRemote+Cards.swift
bozidarsevo: Sure! I made it to be `String?` so that in the future we can pass new values without the need to update `WordPressKit`. But I can use the enum here 👍

WordPressKitTests/ReaderPostServiceRemote+CardsTests.swift
bozidarsevo: I see in the project that some test call the `fullfill()` after `XCTFail()` and some before. I thought that it is usual to fulfill the expectation at the end of checking the logic. What should be the correct way?

WordPressKitTests/ReaderPostServiceRemote+CardsTests.swift
```diff
@@ -93,4 +93,47 @@ class ReaderPostServiceRemoteCardTests: RemoteTestCase, RESTTestable {
...
+                guard let firstPostDate = posts[i-1].post?.sortDate,
+                      let secondPostDate = posts[i].post?.sortDate,
+                      firstPostDate > secondPostDate else {
+                    XCTFail("Posts should be sorted by date, starting with most recent post")
+                    expect.fulfill()
+                    return
+                }
+            }
+            expect.fulfill()
+        }, failure: { _ in })
```
bozidarsevo, line 135: Now I see that I need to also add `XCTFail` in the `failure` block.

WordPressKitTests/ReaderPostServiceRemote+CardsTests.swift
bozidarsevo: If the `mockRemoteApi.URLStringPassedIn` is nil the test will fail.
By checking if the `GET` method is called you mean to fail the test if the `failure` block of the `fetchCards` is called, check the `getMethodCalled` (`XCTAssertTrue(mockRemoteApi.getMethodCalled)`) or some other way?

Gio2018: > If the mockRemoteApi.URLStringPassedIn is nil the test will fail.

it seems to me that if `mockRemoteApi.URLStringPassedIn` is `nil`, then `mockRemoteApi.URLStringPassedIn?.contains("sort=") ?? false` would fall back to `false` and then `XCTAssertFalse(mockRemoteApi.URLStringPassedIn?.contains("sort=") ?? false)` would succeed, right?

Gio2018: > By checking if the GET method

I believe the Mock API has a boolean to check that the method has actually been called (although an `expectation` would probably be preferable, but we can use that boolean too)

WordPressKitTests/ReaderPostServiceRemote+CardsTests.swift
Gio2018: My thought here was that, since the expectation is for "Get cards sorted by date", if this fails you would not get the cards sorted, and so it should not be fulfilled. Would you agree?

WordPressKitTests/ReaderPostServiceRemote+CardsTests.swift
bozidarsevo: > > If the mockRemoteApi.URLStringPassedIn is nil the test will fail.
> 
> it seems to me that if `mockRemoteApi.URLStringPassedIn` is `nil`, then `mockRemoteApi.URLStringPassedIn?.contains("sort=") ?? false` would fall back to `false` and then `XCTAssertFalse(mockRemoteApi.URLStringPassedIn?.contains("sort=") ?? false)` would succeed, right?

Oh, now I understand. Since I copy/pasted those `XCTAssertTrue` cases from other tests I forgot to change the fallback bool value. Thank you for catching it!

WordPressKitTests/ReaderPostServiceRemote+CardsTests.swift
bozidarsevo: I would agree. 👍
Looking at the other tests in the project we should remove the `fullfill()` call in a lot of tests after `XCTFail()`, maybe that can be a thing good to do in the future. 👌

WordPressKit/ReaderPostServiceRemote+Cards.swift
```diff
@@ -1,3 +1,15 @@
+public enum ReaderSortingOption: String, CaseIterable {
+    case popularity, date
+    case noSorting
+    
+    var queryValue: String? {
```
bozidarsevo, line 6: Query value used for creating the endpoint.

WordPressKit/ReaderPostServiceRemote+Cards.swift
```diff
@@ -1,3 +1,15 @@
+public enum ReaderSortingOption: String, CaseIterable {
+    case popularity, date
```
Gio2018: we usually don't put more than one case per line, could you please adjust it?

Gio2018 (changes requested):
Almost there, just one minor change, thank you @bozidarsevo!

WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.26.0-beta.10"
+  s.version       = "4.26.0-beta.11"
```
Gio2018: since 4.26 has been published, can we bump this to 4.27.0-beta-1?

Gio2018 (commented):
I think this looks good now, just one more comment since the release changed and then we are good!!

Gio2018 (approved):

PR #347, opened 2021-02-08
Implementation for post likes endpoint

bitdc:
### Description

This PR adds an endpoint for fetching post likes based on [this API documentation](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/%24post_ID/likes/). Since the likes are contextual to the Post object, the implementation is added in `PostServiceRemoteREST`. The response is then mapped to the existing `RemoteUser` object.

References:
- Related issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/15662
- Design: [Likes Enhancements i2](https://wp.me/pbArwn-1CQ)

### Testing Details

Added test that ensures the object returned from the API are correctly parsed into `RemoteUser`s, and a negative test that ensures the failure block is called.

### DIFF
--- a/WordPressKit/PostServiceRemoteREST.h
+++ b/WordPressKit/PostServiceRemoteREST.h
@@ -3,6 +3,8 @@
 #import <WordPressKit/SiteServiceRemoteWordPressComREST.h>
 #import <WordPressKit/RemoteMedia.h>
 
+@class RemoteUser;
+
 @interface PostServiceRemoteREST : SiteServiceRemoteWordPressComREST <PostServiceRemote>
 
 /**
@@ -52,5 +54,19 @@
 - (void)getAutoSaveForPost:(RemotePost *)post
                    success:(void (^)(RemotePost *))success
                    failure:(void (^)(NSError *error))failure;
+
+/**
+ *  @brief      Requests a list of users that liked the post with the specified ID.
+ *
+ *  @discussion Due to the API limitation, up to 90 users will be returned from the
+ *              endpoint.
+ *
+ *  @param      postID      The ID for the post. Cannot be nil.
+ *  @param      success     The block that will be executed on success. Can be nil.
+ *  @param      failure     The block that will be executed on failure. Can be nil.
+ */
+- (void)getLikesForID:(NSNumber *)postID
+              success:(void (^)(NSArray<RemoteUser *> *))success
+              failure:(void (^)(NSError *))failure;
 
 @end
--- a/WordPressKit/PostServiceRemoteREST.h
+++ b/WordPressKit/PostServiceRemoteREST.h
@@ -65,8 +65,8 @@
  *  @param      success     The block that will be executed on success. Can be nil.
  *  @param      failure     The block that will be executed on failure. Can be nil.
  */
-- (void)getLikesForID:(NSNumber *)postID
-              success:(void (^)(NSArray<RemoteUser *> *))success
-              failure:(void (^)(NSError *))failure;
+- (void)getLikesForPostID:(NSNumber *)postID
+                  success:(void (^)(NSArray<RemoteUser *> *))success
+                  failure:(void (^)(NSError *))failure;
 
 @end
--- a/WordPressKit/PostServiceRemoteREST.m
+++ b/WordPressKit/PostServiceRemoteREST.m
@@ -1,6 +1,7 @@
 #import "PostServiceRemoteREST.h"
 #import "RemotePost.h"
 #import "RemotePostCategory.h"
+#import "RemoteUser.h"
 #import "WPKit-Swift.h"
 @import WordPressShared;
 @import NSObject_SafeExpectations;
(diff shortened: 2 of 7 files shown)


WordPressKit/PostServiceRemoteREST.h
```diff
@@ -53,4 +55,18 @@
...
+ *  @brief      Requests a list of users that liked the post with the specified ID.
+ *
+ *  @discussion Due to the API limitation, up to 90 users will be returned from the
+ *              endpoint.
+ *
+ *  @param      postID      The ID for the post. Cannot be nil.
+ *  @param      success     The block that will be executed on success. Can be nil.
+ *  @param      failure     The block that will be executed on failure. Can be nil.
+ */
+- (void)getLikesForID:(NSNumber *)postID
```
ScoutHarris: Can we make the method name a bit clearer? `ID` is a bit vague. Maybe something like `getLikesforPost` instead?

WordPressKit/PostServiceRemoteREST.m
```diff
@@ -306,6 +307,30 @@ - (void)restorePost:(RemotePost *)post
...
+    
+    NSString *path = [NSString stringWithFormat:@"sites/%@/posts/%@/likes", self.siteID, postID];
+    NSString *requestUrl = [self pathForEndpoint:path
+                                     withVersion:ServiceRemoteWordPressComRESTApiVersion_1_1];
+    
+    [self.wordPressComRestApi GET:requestUrl
+                       parameters:nil
+                          success:^(id responseObject, NSHTTPURLResponse *httpResponse) {
+        NSArray *jsonUsers = responseObject[@"likes"];
+        if (success && jsonUsers) {
```
ScoutHarris: It looks like nothing is returned if `jsonUsers` is not present. Shouldn't we just check for `success` and return an empty array of there are no likes?

WordPressKit/PostServiceRemoteREST.m
```diff
@@ -564,4 +589,42 @@ - (NSArray *)tagNamesFromJSONDictionary:(NSDictionary *)jsonTags {
...
+ *              RemoteUser object):
+ *              - ID -> userID
+ *              - login -> username
+ *              - name -> displayName
+ *              - site_ID -> primaryBlogID
+ *              - avatar_URL -> avatarURL
+ *
+ *  @param  jsonUser    The dictionary that represents a RemoteUser.
+ */
+- (RemoteUser *)remoteUserFromJSONDictionary:(NSDictionary *)jsonUser
```
ScoutHarris, line 618: From your success test, it looks like `email` is included in the response. Not sure if it is ever actually provided, but just in case, can we set `user.email` if it's present and default to `""` if not?

ScoutHarris (changes requested):

ScoutHarris:
Hey David. Just a few minor questions/suggestions. Otherwise, this is looking good! 🎉 

(Sorry, I forgot to comment when I submitted my review.)

WordPressKit/PostServiceRemoteREST.h
bitdc: Yeah, sure! Would `getLikesForPostID` be better here? Because there are other methods that require a `RemotePost` type as the first parameter, such as `-getAutoSaveForPost:(RemotePost *)post`.

WordPressKit/PostServiceRemoteREST.m
bitdc: Nice catch! Will fix this in the next commit.

WordPressKit/PostServiceRemoteREST.m
bitdc: Ah yes, I initially thought that too, but the problem is the `email` parameter returned from the API is a boolean. Here's an example of a complete user JSON from post likes endpoint:
```json
{
    "found": 1,
    "i_like": true,
    "site_ID": 0,
    "post_ID": 1,
    "likes": [
        {
            "ID": 106059135,
            "login": "dvdchr",
            "email": false,
            "name": "dvdchr",
            "first_name": "David",
            "last_name": "Christiandy",
            "nice_name": "dvdchr",
            "URL": "",
            "avatar_URL": "https://0.gravatar.com/avatar/9d5db177ff5c19b2c4806093ec6d87f7?s=96&d=identicon&r=G",
            "profile_URL": "https://en.gravatar.com/dvdchr",
            "ip_address": false,
            "site_ID": 174074163,
            "site_visible": true,
            "default_avatar": false
        }
    ]
}

```
In fact, I don't know what the boolean value in the `email` field represents, as I find that all of the JSON objects have `false` value for every entry. 😅

WordPressKit/PostServiceRemoteREST.h
ScoutHarris: Yep, that'll work too!

WordPressKit/PostServiceRemoteREST.m
ScoutHarris: Ah, I see. Yea, that's a little odd. 🤷 OK. We'll leave it for now. I'm not sure if we'll ever need it in this scenario, but if so we can come back to it. Thanks for checking!

ScoutHarris (approved):
Looks good! Nice job!

PR #352, opened 2021-02-19
Update site design models and service for new API endpoint

### LINKED ISSUE #3125: Initial design picker screen: add categories
### Why are we doing this?

We want to make it easier for our users to find the design that they are looking for by presenting them in a categorized list. Since we are planning to add a lot more designs in the future.

### Goals

- Update the current design picker to behave in the same way as the current categorized page picker (Starter Page Template Picker). 
- The user should be able to select a category. Once selected only the selected categories design should be available to be chosen.
- Once a design is selected the user should be presented with the previous “Preview” and “Choose” buttons. 
- If the user selects preview they should be able to preview the designs. The same way they were able to before.
- If the user selects the  “Choose” button they should be presented with the choose the domain screen. (The same as before)

### Changes that will need to happen to the apps.

1. Update the copy of the description on the screen from "Pick your favourite homepage layout. You can edit and customize it later." to
"Get started with a pre-made layout and customize it to fit your needs"
2. The new categorized design picker will be very similar to the Page picker design. 
3. Our new layouts will be grouped horizontally in the same way as the Starter Page Picker.
4. Note that the aspect ratio of the preview image is staying the same. 

#### Endpoint Changes
We are also implementing a new endpoint. The endpoint is being implemented in D56541-code.

The endpoint will be very similar to the common-block-layouts and will return a list of design + categories. The URL that we are planning to use will be. We have decided that for Android we will return the new "Single page designs + the original designs. While for iOS we will just return the original designs for now. 
https://public-api.wordpress.com/wpcom/v2/common-starter-site-designs/

- Query Parameters:
    - `type=mobile` - to signify this is the mobile apps
    - `group` - You will be able to pass in a group through as a query parameter in order to return the correct group of layouts.
        - By default the endpoint will ____?____ 
        - **iOS** should set this to equal `original` to exclude the old layouts.
        - **Android** should set this to equal `original,single-page`
    - `scale` - This works the same as it does for the common-block-layouts
    - `preview_width` - This works the same as it does for the common-block-layouts

<details>
<summary>Screenshot of the new designs.</summary>

image

</details>

### Additional context 

This Issue is part of the single page layout project. pb3aDo-IP-p2
Internal proposal and discussion.  pb3aDo-Lz-p2

mkevins:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/3125

#### Related PR:

`WordPress-iOS`: https://github.com/wordpress-mobile/WordPress-iOS/pull/15933

### Description

This updates the models and service for the new API endpoint for common starter site designs. It also adds a `withGroups` parameter to `SiteDesignRequest.init` to specify the template groups to be fetched.

### Testing Details

Unit tests: I have updated the unit tests for the site design service to use the new endpoint.

Additional testing steps in the related PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15933


antonis (commented):
Great work @mkevins 👍 
At this stage this LGTM 🎉

WordPressKit/SiteDesignServiceRemote.swift
```diff
@@ -1,35 +1,22 @@
...
-        }
-
-        return parameters
+public struct SiteDesignRequest {
+    public let parameters: [String: AnyObject]
+
+    public init(withThumbnailSize thumbnailSize: CGSize) {
+        parameters = [
+            "preview_width": "\(thumbnailSize.width)" as AnyObject,
+            "scale": UIScreen.main.nativeScale as AnyObject
```
chipsnyder, line 17: Just adding a note here to capture the requirement shift. 
we'll want to expose a way to add a way to set `"group": "original,single-page"` to this. 

https://github.com/wordpress-mobile/gutenberg-mobile/issues/3125#issuecomment-784230338

mkevins: Added via https://github.com/wordpress-mobile/WordPressKit-iOS/pull/352/commits/00700a6d53e54a320333f4ea7b0a0929869bb431, and made available via interface to WordPress-iOS.

WordPressKit/SiteDesignServiceRemote.swift
```diff
@@ -1,35 +1,33 @@
...
-    public let previewSize: CGSize?
-    public let scale: CGFloat?
-
-    public init(previewSize: CGSize? = nil, scale: CGFloat? = nil) {
-        self.previewSize = previewSize
-        self.scale = scale
+public struct SiteDesignRequest {
+    public enum TemplateGroup: String {
+        case original
+        case singlePage = "single-page"
```
antonis, line 7: I like your approach of having the available groups in an `enum` 👍 
I followed a different approach [on Android using just a string array](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1890/files#diff-3be4f77b2c2e670769618b7d0e91b4c7f77115c4cbfa763a5041e63d2ec1395bR94) but using an `enum` seems more bullet proof and I may correct this on my side. Tbh I'm not sure how often there would be a [need to add new groups](https://github.com/wordpress-mobile/gutenberg-mobile/issues/3125#issuecomment-785522260).

antonis (approved):
Great work @mkevins 👍 
Tested this with the [iOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/15933) and everything works as expected. The code changes are consistent and LGTM 🎉

WordPressKit/SiteDesignServiceRemote.swift
chipsnyder: I like the enum approach as well 👍

chipsnyder: From the [discussion](https://github.com/wordpress-mobile/gutenberg-mobile/issues/3125#issuecomment-786106299) in the other issue, we can now remove `original`

WordPressKitTests/SiteDesignServiceRemoteTests.swift
```diff
@@ -2,14 +2,18 @@ import XCTest
 @testable import WordPressKit
 
 class SiteDesignServiceRemoteTests: RemoteTestCase, RESTTestable {
-    let successMockFilename = "nux-starter-designs-success.json"
+    let successMockFilename = "common-starter-site-designs-success.json"
 
     /// The data in this file is missing required values.
-    let malformedMockFilename = "nux-starter-designs-malformed.json"
-    let endpoint = "rest/v1.1/nux/starter-designs"
```
chipsnyder, line 9: We should be able to delete these old files now that the endpoint has changed.

WordPressKitTests/SiteDesignServiceRemoteTests.swift
```diff
@@ -2,14 +2,18 @@ import XCTest
...
+    let malformedMockFilename = "common-starter-site-designs-malformed.json"
```
chipsnyder, line 8: The previous file had an invalid JSON format to test what would happen if the API for some reason returned bad data. I ran the contents of this file through a JSON validator and it came back as valid. Was there a more subtle change in there that I might have missed like a miss alignment on types?

chipsnyder (approved):
Nice work @mkevins overall looks really good left a couple of clarifying questions but nothing blocking.

WordPressKitTests/SiteDesignServiceRemoteTests.swift
mkevins: At first I actually used an invalid JSON file, but I checked the original `nux-starter-designs-malformed.json`, and it seems it was valid as well, from a JSON spec point of view. So, it seems it instead has missing key-value pairs in certain places: e.g. `diff <(cat nux-starter-designs-success.json | jq .) <(cat nux-starter-designs-malformed.json | jq .)` gives:

```
4,5d3
<     "segment_id": 2,
<     "slug": "alves",
7,8d4
<     "site_url": "https://alvesstartermobile.wordpress.com",
<     "demo_url": "https://public-api.wordpress.com/rest/v1/template/demo/alves/alvesstartermobile.wordpress.com/?language=en",
```

Which makes sense after re-reading the original code comment. So I made similar changes to the success response (i.e. deleted a couple of required key-value pairs) to create the malformed response, in order to keep aligned with the intent of the original test.

WordPressKitTests/SiteDesignServiceRemoteTests.swift
mkevins: I've removed these. :+1:

chipsnyder (approved):
I tested this with the latest changes everything looks good to me 👍

WordPressKit/SiteDesignServiceRemote.swift
mkevins: This has been updated to use `stable`. :+1:

PR #354, opened 2021-02-22
Following a conversation: handle success: false

mindgraffiti:
### Description

Ref. https://github.com/wordpress-mobile/WordPress-iOS/issues/15642

**Following** a conversation can return a response object with `success: false`, because the API response can successfully return, but the attempted task to follow a conversation can fail. The task can fail because a user can enable [Block emails](https://wordpress.com/me/notifications/subscriptions) in their Reader Subscription notification settings. **Unfollowing** a conversation should return `success: true` regardless of whether or not the user has enabled "Block emails".

Pass the bool from the response to the callback block.

### Testing Details

Check that all unit tests in `ReaderPostServiceRemote+SubscriptionTests` pass.

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -21,7 +21,7 @@ Pod::Spec.new do |s|
 
   s.dependency 'Alamofire', '~> 4.8.0'
   s.dependency 'CocoaLumberjack', '~> 3.4'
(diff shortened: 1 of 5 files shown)


ScoutHarris (approved):
LGTM!

:shipit:

mindgraffiti:
Thanks @ScoutHarris!

PR #367, opened 2021-03-08
Merge/4.28.0 into trunk

jkmassel:
### Description

Fixes #

### Testing Details


WordPressKit.podspec
```diff
@@ -21,7 +21,7 @@ Pod::Spec.new do |s|
 
   s.dependency 'Alamofire', '~> 4.8.0'
   s.dependency 'CocoaLumberjack', '~> 3.4'
-  s.dependency 'WordPressShared', '~> 1.15.0-beta.1'
+  s.dependency 'WordPressShared', '~> 1.15.0'
```
AliSoftware, line 24: Don't we usually keep the transitive dependencies of pods to point to `-beta`, so that it's more flexible during development? (And then the host app is the one responsible for pinning the transitive dependency to a non-beta version by explicitly listing it in its `Podfile`)

PR #368, opened 2021-03-08
Fix dependency issue

jkmassel:
Fixes an issue that could come up with WordPressShared is a transitive dependency.

Instead of locking this library to a specific WP version, we'll just specify the _minimum_ version that is supported (along with any future betas)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,4 +1,7 @@
 GEM
+  specs:
+
+GEM
   remote: https://rubygems.org/
   specs:
     CFPropertyList (3.0.2)
@@ -87,4 +90,4 @@ DEPENDENCIES
   cocoapods (~> 1.8)!
 
 BUNDLED WITH
-   2.1.4
+   2.2.10
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -21,8 +21,11 @@ Pod::Spec.new do |s|
 
   s.dependency 'Alamofire', '~> 4.8.0'
   s.dependency 'CocoaLumberjack', '~> 3.4'
-  s.dependency 'WordPressShared', '~> 1.15.0'
   s.dependency 'NSObject-SafeExpectations', '0.0.4'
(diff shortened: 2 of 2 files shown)


WordPressKit.podspec
```diff
@@ -21,8 +21,11 @@ Pod::Spec.new do |s|
 
   s.dependency 'Alamofire', '~> 4.8.0'
   s.dependency 'CocoaLumberjack', '~> 3.4'
-  s.dependency 'WordPressShared', '~> 1.15.0'
   s.dependency 'NSObject-SafeExpectations', '0.0.4'
   s.dependency 'wpxmlrpc', '~> 0.9'
   s.dependency 'UIDeviceIdentifier', '~> 1.4'
+
+    # Use a loose restriction that allows both production and beta versions, up to the next major version.
```
AliSoftware: Indentation :triggered:

Gemfile.lock
```diff
@@ -1,3 +1,6 @@
+GEM
+  specs:
+
```
AliSoftware, line 3: 🤔 wat?

WordPressKit.podspec
mindgraffiti:
```suggestion
# Use a loose restriction that allows both production and beta versions, up to the next major version.
```

WordPressKit.podspec
```diff
@@ -21,8 +21,11 @@ Pod::Spec.new do |s|
...
+  # If you want to update which of these is used, specify it in the host app.
+  s.dependency 'WordPressShared', '~> 1.15-beta' # Don't change this until we hit 2.0 in WPShared
```
mindgraffiti, line 30: We can definitely try this, but I foresee people just updating the shared version in WPKit when this doesn't work (instead of asking questions) 🤷‍♀️

mindgraffiti (approved):
- left a whitespace nitpick
- don't forget to bump WPKit version before you merge :)

PR #370, opened 2021-03-11
Changes the PluginDirectoryEntry.rating property to a Double

emilylaguna:
### Description

WPiOS Related PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16070

### Testing Details

Follow the steps on the WPiOS Related PR

### DIFF
--- a/WordPressKit/PluginDirectoryEntry.swift
+++ b/WordPressKit/PluginDirectoryEntry.swift
@@ -30,9 +30,9 @@ public struct PluginDirectoryEntry {
         return extractHTMLText(self.changelogHTML)
     }
 
-    let rating: Int
+    let rating: Double
     public var starRating: Double {
-        return (Double(rating) / 10).rounded() / 2
+        return (rating / 10).rounded() / 2
         // rounded to nearest half.
     }
 }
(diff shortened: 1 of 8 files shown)


WordPressKit/PluginDirectoryEntry.swift
```diff
@@ -81,7 +81,7 @@ extension PluginDirectoryEntry: Codable {
         slug = try container.decode(String.self, forKey: .slug)
         version = try? container.decode(String.self, forKey: .version)
         lastUpdated = try? container.decode(Date.self, forKey: .lastUpdated)
-        rating = try container.decode(Int.self, forKey: .rating)
+        rating = try container.decode(Double.self, forKey: .rating)
```
leandroalonso, line 84: Can you add a unit test for this?

emilylaguna: There were already some existing unit tests, so I went ahead and updated the test JSON to also include float rating values: b27516c0360533e4d65e38cae7a004fc2ecee853

mokagio: Hey @emilylaguna, thanks for fixing this and updating the tests, too! 

Great idea leaving some integers and not updating everything to doubles in the JSON fixtures. Not that one should really worry that Swift will stop converting a JSON number that doesn't have a `.` from integer to double, but doesn't hurt to have the check there since we have the big JSON data set. 👍

leandroalonso (approved):

PR #374, opened 2021-03-16
Push podspec to internal spec repo too

AliSoftware:
This will make CI push this podspec to our new internal spec repo on tag (in addition to still pushing it to trunk as usual).

Similar to:
 - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/580
 - https://github.com/wordpress-mobile/WordPressUI-iOS/pull/83
 - https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/289

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,12 +4,22 @@ orbs:
   # Using 1.0 of our Orbs means it will use the latest 1.0.x version from https://github.com/wordpress-mobile/circleci-orbs
   ios: wordpress-mobile/ios@1.0
 
+# YAML anchors for some common/repeated values
+x-common-params:
+  - &xcode-version "11.2.1"
+  - &podspec "WordPressKit.podspec"
+  - &on-tags-only
+      tags:
+        only: /.*/
+      branches:
+        ignore: /.*/
+
 workflows:
   test_and_validate:
     jobs:
       - ios/test:
           name: Test
-          xcode-version: "11.2.1"
+          xcode-version: *xcode-version
           workspace: WordPressKit.xcworkspace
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -4,12 +4,22 @@ orbs:
   # Using 1.0 of our Orbs means it will use the latest 1.0.x version from https://github.com/wordpress-mobile/circleci-orbs
   ios: wordpress-mobile/ios@1.0
 
+# YAML anchors for some common/repeated values
+x-common-params:
+  - &xcode-version "11.2.1"
```
jkmassel, line 9: Weird question – could this go in the Orb?

So we could update the whole CI fleet at once? 😅

jkmassel (approved):

.circleci/config.yml
AliSoftware: You mean as a default value?

Because I still think it's useful for the orb to allow you to configure it per project.
 - Especially if you consider the orb as being OSS and not just restricted to a8c only, where others could use the orb's commands for their own project but use a different Xcode version
 - But also because maybe some of our pods might not migrate as the same rate depending on the projects. For example imagine a pod that is only used by Simplenote but not any other app (`WordPress-Ratings-iOS` is such an example) and Simplenote not yet being ready to migrate to the latest Xcode while other apps and pods have…

I am not against providing a default value via the orb, so that for most pods we don't override it in the pods' own config and rely on that default, and so we can update them all at once. But that could also have unintended consequences like upgrading pods that haven't yet been tested with a new version of Xcode but would still transparently be bumped without us realizing the extent of the propagation at first.

PR #375, opened 2021-03-22
Filter Jetpack sites for an account

momo-ozawa:
## Description

- Added a new getBlogs method that includes a `filterJetpackSites` flag 

## How to test

- WP PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16140

### DIFF
--- a/WordPressKit/AccountServiceRemote.h
+++ b/WordPressKit/AccountServiceRemote.h
@@ -14,13 +14,15 @@ typedef NS_ERROR_ENUM(AccountServiceRemoteErrorDomain, AccountServiceRemoteError
 @protocol AccountServiceRemote <NSObject>
 
 /**
- *  @brief      Gets all blogs for an account.
+ *  @brief      Gets blogs for an account.
  *
- *  @param      success     The block that will be executed on success.  Can be nil.
- *  @param      failure     The block that will be executed on failure.  Can be nil.
+ *  @param      filterJetpackSites    Whether we're fetching only Jetpack blogs.
+ *  @param      success      The block that will be executed on success.  Can be nil.
+ *  @param      failure      The block that will be executed on failure.  Can be nil.
  */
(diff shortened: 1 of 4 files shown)


WordPressKit/AccountServiceRemote.h
```diff
@@ -14,13 +14,15 @@ typedef NS_ERROR_ENUM(AccountServiceRemoteErrorDomain, AccountServiceRemoteError
...
  *
- *  @param      success     The block that will be executed on success.  Can be nil.
- *  @param      failure     The block that will be executed on failure.  Can be nil.
+ *  @param      filterJetpackSites    Whether we're fetching only Jetpack blogs.
+ *  @param      success      The block that will be executed on success.  Can be nil.
+ *  @param      failure      The block that will be executed on failure.  Can be nil.
  */
-- (void)getBlogsWithSuccess:(void (^)(NSArray *blogs))success
-                    failure:(void (^)(NSError *error))failure;
+- (void)getBlogs:(BOOL)filterJetpackSites
```
emilylaguna, line 23: What are your thoughts on adding a new method with this flag instead of modifying the existing one?

That way we don't need to worry about regressions of the other places uses this method.

emilylaguna (changes requested):
Nice, works well. Left a comment about adding a new method instead of renaming.

WordPressKit/AccountServiceRemote.h
momo-ozawa: @emilylaguna Fixed! 937795a 

I was going back and forth between adding a `getJetpackBlogsWithSuccess:failure:` method OR modifying the existing method... I think your suggestion of keeping the existing method AND adding a new method with the filter flag seems like the best way to avoid regressions 👍

emilylaguna (approved):
LGTM!

PR #378, opened 2021-04-06
Merge 4.30.0 into trunk

jkmassel:
Merges changes from the `4.30.0` series into `trunk`.


jkmassel:
Admin-merging this to keep the WPiOS 17.1 code freeze moving

PR #382, opened 2021-04-12
Validate XMLRPC on HTTPS only if configured by hosting app

shiki:
## Description

Fixes part of https://github.com/woocommerce/woocommerce-ios/issues/3891. 

WooCommerce has been having complaints about unclear error messages when trying to log in using a site address. We discovered that the app  just shows “Not a WordPress site” for these scenarios:

- Offline
- The site only has HTTP
- Has a basic auth 
- Probably more

It turns out that we're incorrectly using the [`NSURLErrorAppTransportSecurityRequiresSecureConnection` `-1022`](https://developer.apple.com/documentation/foundation/1508628-url_loading_system_error_codes/nsurlerrorapptransportsecurityrequiressecureconnection) error code in WooCommerce's [`AuthenticationManager`](https://github.com/woocommerce/woocommerce-ios/blob/b5aa7a8b9c03078c3af97b6b29c2c5cee05c47e6/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L400) to determine if the site address is invalid. 

The source of the `-1022` error code is from `WordPressOrgXMLRPCValidator` which tries to find `xmlrpc.php` on both the HTTPS and HTTP version of the site:

https://github.com/wordpress-mobile/WordPressKit-iOS/blob/eedca91128930eb77f64cab180e822c02519ebcc/WordPressKit/WordPressOrgXMLRPCValidator.swift#L68-L74

This is a problem for the WooCommerce app because: 

- The WooCommerce app's [configuration (`Info.plist`)](https://github.com/woocommerce/woocommerce-ios/blob/2626b663dc01f8dec63fdcb7b24078bf2483a2de/WooCommerce/Resources/Info.plist#L5-L9) does not allow HTTP-only sites. Hence, if the site only has HTTP, then we would unnecessarily receive `NSURLErrorAppTransportSecurityRequiresSecureConnection`. 
- When the device is offline, iOS seems to prioritize returning `-1022` instead of [`NSURLErrorNotConnectedToInternet (-1009)`](https://developer.apple.com/documentation/foundation/1508628-url_loading_system_error_codes/nsurlerrornotconnectedtointernet). This prevents us from detecting offline errors and presenting a more accurate error message to the customer instead of the unfortunate “Not a WordPress site”. 

We discussed whether we should allow HTTP on WooCommerce but we decided against that (p91TBi-4Pz-p2). 

## Changes

This changes `WordPressOrgXMLRPCValidator` so that it will only validate HTTP sites if the hosting app's `Info.plist`'s [`NSAppTransportSecurity`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity) setting is configured to allow unsecured access. For WordPress, both HTTP and HTTPS will be validated. For WooCommerce, only HTTPS. 

https://github.com/wordpress-mobile/WordPressKit-iOS/blob/1f23174d1e9b14f7eb7748b435504257d392e346/WordPressKit/WordPressOrgXMLRPCValidator.swift#L76-L96

This change allows the WooCommerce app to detect the correct error codes. For example, we will be able to detect offline states and present the error to the customer:

Please see [`AppTransportSecuritySettings.swift`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/issue/woo/3891-check-for-secure-access/WordPressKit/AppTransportSecuritySettings.swift) for how this is implemented. 

### Side Effects

There should be no side effects on the WordPress app. For the WooCommerce app, offline error messages will now be displayed when upgrading to this WordPressKit version. However, the “Need Help?” dialogs shown point to a WordPress help page, which is incorrect. 🙈 I will fix those later if this PR is approved. 

### WordPress Regression Test

I prepared a branch that uses this WPKit version: https://github.com/wordpress-mobile/WordPress-iOS/pull/16287. 

1. Log out. 
2. Tap “Enter your existing site address”. 
3. Enter a site that does not have HTTPS. If you don't have one, you can use `http://httponly.peregrinator.net`. The password is available in the Secret Store. 
4. Confirm that you were able to log in with the site. 
5. Remove the site. 
6. Go offline. 
7. Enter the site URL again. 
8. Confirm that the “The Internet connection appears to be offline.” is displayed. 
9. Go online. 
10. Enter a self-hosted site that only has HTTPS. Any Jurassic site will do. 
11. Confirm that you can log in. 

### WooCommerce Test

I prepared a branch that uses this WPKit version: https://github.com/woocommerce/woocommerce-ios/pull/3957 

1. Log out. 
2. Tap “Enter Your Store Address”. 
3. Enter a site that does not have HTTPS. If you don't have one, you can use `http://httponly.peregrinator.net`. The password is available in the Secret Store. 
4. Confirm that an error dialog is shown: “Unable to read the WordPress site at...”. This is the default error and I will fix this in a subsequent PR. 
6. Go offline. 
7. Enter the site URL again. 
8. Confirm that the “The Internet connection appears to be offline.” is displayed. This is also the default error and I'd probably change this too. The good news is that the app no longer says “Not a WordPress site”. 
9. Go online. 
10. Enter a self-hosted WooCommerce site that only has HTTPS. Any Jurassic site will do. Jurassic can also auto-install WooCommerce. 
11. Tap Continue. 
12. Confirm that an error is shown: “To use this app, ... you'll need to have the Jetpack plugin installed...”. 
13. Or if you _activated Jetpack_ on the site, you should be asked for your WordPress.com password. 

## Author Checklist

### DIFF
--- a/WordPressKit/WordPressOrgXMLRPCValidator.swift
+++ b/WordPressKit/WordPressOrgXMLRPCValidator.swift
@@ -44,6 +44,8 @@ open class WordPressOrgXMLRPCValidator: NSObject {
     // is the default threshold for allowable redirects.
     private let redirectLimit = 16
 
+    private let appTransportSecuritySettings = AppTransportSecuritySettings()
+
     override public init() {
         super.init()
     }
@@ -65,12 +67,24 @@ open class WordPressOrgXMLRPCValidator: NSObject {
 
         var sitesToTry = [String]()
 
+        let secureAccessOnly: Bool = {
+            if let url = URL(string: site) {
+                return appTransportSecuritySettings.secureAccessOnly(for: url)
+            } else {
+                return true
+            }
+        }()
+
         if site.hasPrefix("http://") {
-            sitesToTry.append(site)
+            if !secureAccessOnly {
+                sitesToTry.append(site)
+            }
             sitesToTry.append(site.replacingOccurrences(of: "http://", with: "https://"))
         } else if site.hasPrefix("https://") {
             sitesToTry.append(site)
-            sitesToTry.append(site.replacingOccurrences(of: "https://", with: "http://"))
+            if !secureAccessOnly {
+                sitesToTry.append(site.replacingOccurrences(of: "https://", with: "http://"))
+            }
         } else {
             failure(WordPressOrgXMLRPCValidatorError.invalidScheme as NSError)
         }
--- a/WordPressKit/WordPressOrgXMLRPCValidator.swift
+++ b/WordPressKit/WordPressOrgXMLRPCValidator.swift
@@ -44,9 +44,15 @@ open class WordPressOrgXMLRPCValidator: NSObject {
     // is the default threshold for allowable redirects.
     private let redirectLimit = 16
 
-    private let appTransportSecuritySettings = AppTransportSecuritySettings()
+    private let appTransportSecuritySettings: AppTransportSecuritySettings
 
     override public init() {
+        appTransportSecuritySettings = AppTransportSecuritySettings()
+        super.init()
+    }
+
+    init(_ appTransportSecuritySettings: AppTransportSecuritySettings) {
+        self.appTransportSecuritySettings = appTransportSecuritySettings
         super.init()
     }
 
--- a/WordPressKit/WordPressOrgXMLRPCValidator.swift
+++ b/WordPressKit/WordPressOrgXMLRPCValidator.swift
@@ -76,9 +76,8 @@ open class WordPressOrgXMLRPCValidator: NSObject {
         let secureAccessOnly: Bool = {
             if let url = URL(string: site) {
                 return appTransportSecuritySettings.secureAccessOnly(for: url)
-            } else {
-                return true
             }
+            return true
         }()
 
         if site.hasPrefix("http://") {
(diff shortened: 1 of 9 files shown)


shiki:
@aerych or @ScoutHarris Would you have time to review this? 🙂 🙏

WordPressKit/WordPressOrgXMLRPCValidator.swift
```diff
@@ -65,12 +73,24 @@ open class WordPressOrgXMLRPCValidator: NSObject {
 
         var sitesToTry = [String]()
 
+        let secureAccessOnly: Bool = {
+            if let url = URL(string: site) {
+                return appTransportSecuritySettings.secureAccessOnly(for: url)
+            } else {
```
ScoutHarris: Nitpick - this `else` isn't really necessary.

ScoutHarris (approved):
Hey @shiki . Thanks for providing the HTTP only site! The tests passed and I was able to log into WP successfully in various ways. 👍 from me!

WordPressKit/WordPressOrgXMLRPCValidator.swift
shiki: Thanks! Removed in 23ec5a1e95bb4deae94e349f0eaaf72db12f7472.

shiki:
Thank you, @ScoutHarris!

PR #381, opened 2021-04-12
Include Author ID in Post Service parameters.

twstokes:
### Description

Fixes #380 

### Testing Details

An authorID provided by a `RemotePost` should translate to a `POST` parameter for REST and XML-RPC calls.

- [ ] Please check here if your pull request includes additional test coverage.

PR #385, opened 2021-04-19
Add edit context to restore post call

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/3342
**Related PR** https://github.com/wordpress-mobile/WordPress-iOS/pull/16342

### Description

When restoring a post with basic content such as a single paragraph block the server will sometimes treat that as a Classic Post if you don't include the `context=edit` query parameter. 

You can validate this by deleting a post then making a `POST` request to:
`https://public-api.wordpress.com/rest/v1.1/sites/{{site_id}}/posts/{{post_id}}/restore`
`https://public-api.wordpress.com/rest/v1.1/sites/{{site_id}}/posts/{{post_id}}/restore?context=edit`

This is 100% repeatable for posts that have been created as part of Headstart content.

### Testing Details

Using the build from PR https://github.com/wordpress-mobile/WordPress-iOS/pull/16342

1. Create a site that generates posts such as Cassel under the Blog category. 
    * This is an important step because for some reason that I wasn't able to track down logging in on a different device restores the content as a Gutenberg Post without then added parameter.
2. Navigate to a Post and ensure it's a Gutenberg Post.
3. Delete it. 
4. Restore it. 
5. Validate that the restored post (now in drafts) uses the Gutenberg Editor. 

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.31.0"
+  s.version       = "4.32.0-beta.2"
 
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
   s.description   = <<-DESC
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.32.0-beta.2"
+  s.version       = "4.32.0-beta.3"
 
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
   s.description   = <<-DESC
--- a/WordPressKit/PostServiceRemoteREST.m
+++ b/WordPressKit/PostServiceRemoteREST.m
@@ -290,6 +290,7 @@ static NSString * const RemoteOptionValueOrderByPostID = @"ID";
     NSParameterAssert([post isKindOfClass:[RemotePost class]]);
 
     // The parameters are passed as part of the string here because AlamoFire doesn't encode parameters on POST requests.
+    // https://github.com/wordpress-mobile/WordPressKit-iOS/pull/385
     NSString *path = [NSString stringWithFormat:@"sites/%@/posts/%@/restore?context=edit", self.siteID, post.postID];
     NSString *requestUrl = [self pathForEndpoint:path
                                      withVersion:ServiceRemoteWordPressComRESTApiVersion_1_1];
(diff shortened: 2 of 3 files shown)


WordPressKit/PostServiceRemoteREST.m
```diff
@@ -289,7 +289,9 @@ - (void)restorePost:(RemotePost *)post
 {
     NSParameterAssert([post isKindOfClass:[RemotePost class]]);
 
-    NSString *path = [NSString stringWithFormat:@"sites/%@/posts/%@/restore", self.siteID, post.postID];
+    // The parameters are passed as part of the string here because AlamoFire doesn't encode parameters on POST requests.
+    // https://github.com/wordpress-mobile/WordPressKit-iOS/pull/385
+    NSString *path = [NSString stringWithFormat:@"sites/%@/posts/%@/restore?context=edit", self.siteID, post.postID];
```
chipsnyder, line 294: 📓 I added the query parameters as part of the string here because AlamoFire [doesn't parse parameters](https://github.com/Alamofire/Alamofire/blob/32573d05b91d8b7349ca31b4726e80212483f02a/Source/ParameterEncoding.swift#L298-L315) when sent to a POST request.

guarani:
> This is 100% repeatable for posts that have been created as part of Headstart content.

Curious, what is Headstart content?

WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.31.0"
+  s.version       = "4.32.0-beta.2"
```
guarani: When I saw this is jumping from `4.31.0` to `4.32.0-beta.2`, I realized this wasn't updated in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/381.
Since I made a release for `4.32.0-beta.2`, I'm going to update this on develop and edit the release, otherwise WPiOS might run into issues.

WordPressKit/PostServiceRemoteREST.m
guarani: I think I've run into this before, should be fine here hard-coded I think 👍. These parameters are static so there's no risk of issues regarding URL encoding.

WordPressKit.podspec
guarani: On second thought, it's probably better to let this be and just wait to this new beta is merged and WPiOS is updated.

guarani (approved):
I think these changes are solid and I've tested the API change in the WP.com API console here: https://developer.wordpress.com/docs/api/console/

enejb:
> Curious, what is Headstart content?

Headstart content is the default content that is generated on-site creation for .com sites.
Headstart is .com specific name for this feature.

WordPressKit.podspec
chipsnyder: Just leaving a note here that `4.32.0-beta.2` was published so before merging this I'll bump this to `4.32.0-beta.3` or the appropriate version

guarani:
Thanks for explaining that, @enejb!

PR #387, opened 2021-04-20
Parse a remote user's email response.

aerych:
### Description

Refs https://github.com/wordpress-mobile/WordPress-iOS/pull/16215

This updates `BlogServiceRemoteXMLRPC.remoteUserFromXMLRPCDictionary` to include the remote user's email address as part of the response.

@ScoutHarris Game for a review?

### Testing Details

This can be tested via https://github.com/wordpress-mobile/WordPress-iOS/pull/16215. Set a break point and confirm that the email address is included in the response.

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.31.0"
+  s.version       = "4.31.1-beta.1"
 
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
   s.description   = <<-DESC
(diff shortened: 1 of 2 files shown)


WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.31.0"
+  s.version       = "4.31.1-beta.1"
```
ScoutHarris: `4.31` has already been cut; we're now on `4.32`. Also, the previous person didn't update the version correctly so it's hard to tell, but there is [already](https://github.com/wordpress-mobile/WordPressKit-iOS/releases) a `4.32.0-beta.1` so this should be `4.32.0-beta.2`.

ScoutHarris (approved):
After the version is corrected, :shipit: !

WordPressKit.podspec
aerych: Making a note to check the releases in the future rather than looking at the podspec. :+1:

aerych:
Thanks @ScoutHarris !

PR #389, opened 2021-04-21
Likes: update endpoint version and parse new data

ScoutHarris:
### Description

Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/15662
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16353

This uses the new v1.2 versions of the Post and Comment likes endpoints to obtain:
- `preferred_blog`
- `bio`
- `date_liked`

New classes:
- `RemoteLikeUser` that inherits from `RemoteUser` and adds specific Like info.
- `RemoteLikeUserPreferredBlog` used by `RemoteLikeUser` to store `preferred_blog` info.

Note:
Something to consider when reviewing the structure I implemented. It is my intent to create `LikeUser` and `LikeUserPreferredBlog` entities in WPiOS core data (names TBD if you have a preference). The `LikeUser` will have a relationship to `LikeUserPreferredBlog`. 

### Testing Details

Can be smoke tested with the referenced WPiOS PR. These changes aren't used in WPiOS yet, but since `RemoteLikeUser` inherits from `RemoteUser`, likers should still display as before.

### DIFF
--- a/WordPressKit/CommentServiceRemoteREST.m
+++ b/WordPressKit/CommentServiceRemoteREST.m
@@ -390,14 +390,14 @@
 }
 
 - (void)getLikesForCommentID:(NSNumber *)commentID
-                     success:(void (^)(NSArray<RemoteUser *> *))success
+                     success:(void (^)(NSArray<RemoteLikeUser *> *))success
                      failure:(void (^)(NSError *))failure
 {
     NSParameterAssert(commentID);
 
     NSString *path = [NSString stringWithFormat:@"sites/%@/comments/%@/likes", self.siteID, commentID];
     NSString *requestUrl = [self pathForEndpoint:path
-                                     withVersion:ServiceRemoteWordPressComRESTApiVersion_1_1];
+                                     withVersion:ServiceRemoteWordPressComRESTApiVersion_1_2];
 
     [self.wordPressComRestApi GET:requestUrl
                        parameters:nil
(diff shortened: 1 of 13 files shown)


WordPressKit/CommentServiceRemoteREST.m
```diff
@@ -473,34 +473,11 @@ - (NSString *)remoteStatusWithStatus:(NSString *)status
  
  @param jsonUsers An array containing JSON representations of users.
  */
-- (NSArray<RemoteUser *> *)remoteUsersFromJSONArray:(NSArray *)jsonUsers
+- (NSArray<RemoteLikeUser *> *)remoteUsersFromJSONArray:(NSArray *)jsonUsers
 {
     return [jsonUsers wp_map:^id(NSDictionary *jsonUser) {
-        return [self remoteUserFromJSONDictionary:jsonUser];
+        return [[RemoteLikeUser alloc] initWithDictionary: jsonUser];
```
aerych, line 479: Nice

WordPressKit/RemoteUser+Likes.swift
```diff
@@ -0,0 +1,48 @@
...
+        }()
+    }
+
+}
+
+public class RemoteLikeUserPreferredBlog: NSObject {
+    let blogUrl: String
+    let blogName: String
+    let iconUrl: String
+    let blogID: Int?
```
aerych: Should these be marked `@objc`?  Looks like they might not be accessible to the CommentService class if they aren't.

aerych (approved):
Howdy @ScoutHarris 👋 
Looks good. Re-ran unit tests and tested with WPiOS.  
I think there's one thing we probably want to adjust, but other than that :shipit: :)

WordPressKit/RemoteUser+Likes.swift
ScoutHarris: Well you are correct sir. Thank you for spotting that.

PR #397, opened 2021-05-10
getBlogs: return atomic sites when Jetpack filter is applied

leandroalonso:
Test in: https://github.com/wordpress-mobile/WordPress-iOS/pull/16463


momo-ozawa (approved):
🚀

PR #400, opened 2021-05-17
Add pagination support to Post and Comment Likes endpoints.

ScoutHarris:
### Description

Ref: n/a
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16518

This adds pagination support to Post and Comment Likes endpoints. Both `getLikesFor` methods now accept two new params:
- `count`: Number of records to retrieve. Used for the `number` endpoint param.
- `before`: String version of a date/time to fetch likes before. Used for the `before` endpoint param.

### Testing Details

Can be tested with referenced WPiOS PR.

### DIFF
--- a/WordPressKit/CommentServiceRemoteREST.m
+++ b/WordPressKit/CommentServiceRemoteREST.m
@@ -390,7 +390,9 @@
 }
 
 - (void)getLikesForCommentID:(NSNumber *)commentID
-                     success:(void (^)(NSArray<RemoteLikeUser *> *))success
+                       count:(NSNumber *)count
+                      before:(NSString *)before
+                     success:(void (^)(NSArray<RemoteLikeUser *> * _Nonnull users, NSNumber *found))success
                      failure:(void (^)(NSError *))failure
 {
     NSParameterAssert(commentID);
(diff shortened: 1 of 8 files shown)


WordPressKit/CommentServiceRemoteREST.m
```diff
@@ -400,12 +402,25 @@ - (void)getLikesForCommentID:(NSNumber *)commentID
...
     NSNumber *siteID = self.siteID;
 
+    // If no count provided, default to endpoint max.
+    if (count == 0) {
+        count = @90;
+    }
+    
+    NSMutableDictionary *parameters = [NSMutableDictionary dictionaryWithDictionary:@{ @"number": count }];
+    
+    if (before != nil) {
```
aerych, line 412: I like the explicit equality check. :)  `if (before) {` would suffice for a `nil` check but with fewer of us using objc with any frequency I think this makes the code more clear.

aerych (approved):
Howdy @ScoutHarris 👋 
I tested with the companion wpios patch and everything looked good. 
:shipit: from me :)

PR #403, opened 2021-05-28
pass on the statusCode if one exists

### LINKED ISSUE #16337: Non-descriptive error message when logging in with incorrect username and/or password
### Expected behavior

I would expect to display a meaningful error message on the login screen when the user entered the incorrect username and/or password.

### Actual behavior

The error message shown on the screen is "The Operation couldn't be completed. (WordPressKit.WordPressOrgXMLRPCApiError Code 0.)

IMG_32B29730A99C-1

1. Select "Enter your existing site address" on the app's login screen
2. Enter a self-hosted WordPress site address
3. On the next screen, enter an incorrect username and/or password

enejb:
### Description

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/16337

This Pr makes sure that we pass on the statusCode to the error object. 
This helps code down the line know what the response code is and act accordingly. 

Before:

After:

### Testing Details
Apply this patch and build the WordPress App. 

Try to add a new .org site.  
Type in the wrong username and password. Notice that the error makes sense.

### DIFF
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -301,7 +301,6 @@ open class WordPressOrgXMLRPCApi: NSObject {
 
     private func convertError(_ error: NSError, data: Data?, statusCode: Int? = nil) -> NSError {
         let responseCode = statusCode ?? error.code
-
         if let data = data {
             var userInfo: [AnyHashable: Any] = error.userInfo
             userInfo[type(of: self).WordPressOrgXMLRPCApiErrorKeyData] = data
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -300,7 +300,7 @@ open class WordPressOrgXMLRPCApi: NSObject {
     @objc public static let WordPressOrgXMLRPCApiErrorKeyStatusCode: NSError.UserInfoKey = "WordPressOrgXMLRPCApiErrorKeyStatusCode"
 
     private func convertError(_ error: NSError, data: Data?, statusCode: Int? = nil) -> NSError {
-        let responseCode = statusCode ?? error.code
+        let responseCode = statusCode == 403 ? 403 : error.code
         if let data = data {
             var userInfo: [AnyHashable: Any] = error.userInfo
             userInfo[type(of: self).WordPressOrgXMLRPCApiErrorKeyData] = data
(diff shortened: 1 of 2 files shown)


WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -300,12 +300,14 @@ open class WordPressOrgXMLRPCApi: NSObject {
     @objc public static let WordPressOrgXMLRPCApiErrorKeyStatusCode: NSError.UserInfoKey = "WordPressOrgXMLRPCApiErrorKeyStatusCode"
 
     private func convertError(_ error: NSError, data: Data?, statusCode: Int? = nil) -> NSError {
+        let responseCode = ( (statusCode != nil) ? statusCode! : error.code );
```
enejb:
```suggestion
        let responseCode = ( (statusCode != nil) ? statusCode! : error.code )
```

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -300,12 +300,14 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
+
```
enejb:
```suggestion

```
fix trailing whitespace

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -300,12 +300,13 @@ open class WordPressOrgXMLRPCApi: NSObject {
     @objc public static let WordPressOrgXMLRPCApiErrorKeyStatusCode: NSError.UserInfoKey = "WordPressOrgXMLRPCApiErrorKeyStatusCode"
 
     private func convertError(_ error: NSError, data: Data?, statusCode: Int? = nil) -> NSError {
+        let responseCode = ( (statusCode != nil) ? statusCode! : error.code )
```
frosty: Here you can just do `let responseCode = statusCode ?? error.code`

frosty (commented):
Thanks for tackling this! The approach seems like it works, I'm just slightly concerned whether any of our existing code will be relying on the error code that we currently pass out. As we're already including the status code in the `userInfo` dict, I wonder if we should instead be checking this in the place where display the code. For example, the code in `displayRemoteError(_:)` of `LoginUsernamePasswordViewController` in WordPressAuthenticator might become something like this:

```
let err = error as NSError
        let statusCode = err.userInfo[WordPressOrgXMLRPCApi.WordPressOrgXMLRPCApiErrorKeyStatusCode as String] as? Int
        if err.code == 403 || statusCode == 403 {
            displayError(message: NSLocalizedString("It looks like this username/password isn't associated with this site."...
```

But I'd be very interested to get @jleandroperez's thoughts on this!

enejb:
Thanks for the suggestion @frosty! 

I created a new PR. https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/593 works the same as this but doesn't overwrite the error.code. 

I did notice that the error code of 403 is expected in quite a few different places. so I think this Approach here might still make sense since it might fix things in multiple places? Should we just limit it to the 403 error code?

frosty:
Interesting, do you think there are other places we're probably not handling it correctly already?  I guess if that's the case, limiting it just to 403 for now might be a safer option 🤔 I've asked Jorge on the other PR which approach he'd go for.

jleandroperez (approved):
:shipit:

PR #405, opened 2021-06-02
Revert: incouding atomic sites with Jetpack

leandroalonso:
Test on: https://github.com/wordpress-mobile/WordPress-iOS/pull/16613

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.34.0"
+  s.version       = "4.34.1"
 
   s.summary       = "WordPressKit offers a clean and simple WordPress.com and WordPress.org API."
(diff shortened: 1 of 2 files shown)


WordPressKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressKit"
-  s.version       = "4.34.0"
+  s.version       = "4.34.1"
```
momo-ozawa, line 3: Should this be 4.34.1-beta.1?

momo-ozawa (approved):
LGTM w/ a question about the version.

WordPressKit.podspec
leandroalonso: Actually no. Since I'll be targeting the `release` branch on WPiOS this should be a hotfix. :)

momo-ozawa: Gotcha. Thanks for the clarification!

PR #407, opened 2021-06-13
Issue/14055 - Mark referrer as spam

### LINKED ISSUE #14055: Stats: Referrers: Not Possible to Mark a Referrer as Spam
In Calypso, referrers which are not whitelisted can be marked spam by tapping the three dots to the right of the referrer. This is not the case in the WordPress iOS app (or Android).

Expected:
I expected to be able to mark referrers as spam from within the app.

Actual: 
The option to mark referrers as spam doesn’t exist within the app.

Stats referrers in WPiOS and Calypso web (same site) on iPhone SE:
iosreferrers

##### Tested on iPhone SE, iOS 13.1.2, WPiOS 14.6
This came up in #2934822-zen when a user wanted to mark some referrers as spam but couldn't.

nikola-milicevic:
### Mark referrer as spam

Part of [#14055](https://github.com/wordpress-mobile/WordPress-iOS/issues/14055)

This PR includes needed changes to the `StatsTopReferrersTimeIntervalData` and `StatsReferrer`, in order to be able to mark referrers as spam. It does so by adding `isSpam: Bool` to the `StatsReferrer`, and by making `children` and `referrers` properties mutable.
Also, a method is added to the `StatsServiceRemoteV2` for marking / unmarking referrers as spam.

### DIFF
--- a/WordPressKit/StatsServiceRemoteV2.swift
+++ b/WordPressKit/StatsServiceRemoteV2.swift
@@ -59,6 +59,28 @@ public class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
         })
     }
 
+    /// Used to mark or unmark referrer as spam, depending of the current value.
+    /// - parameters:
+    ///   - referrerDomain: A referrer's domain.
+    ///   - currentValue: Current value of the `isSpam` referrer's property.
+    public func toggleSpamState(for referrerDomain: String,
+                                currentValue: Bool,
+                                success: @escaping () -> Void,
+                                failure: @escaping (Error) -> Void) {
+        let path: String
+        if currentValue {
+            path = self.path(forEndpoint: "sites/\(siteID)/stats/referrers/spam/delete?domain=\(referrerDomain)", withVersion: ._1_1)
+        } else {
+            path = self.path(forEndpoint: "sites/\(siteID)/stats/referrers/spam/new?domain=\(referrerDomain)", withVersion: ._1_1)
+        }
+
+        wordPressComRestApi.POST(path, parameters: nil) { object, _ in
+            success()
+        } failure: { error, _ in
+            failure(error)
+        }
+    }
+
     /// Used to fetch data about site over a specific timeframe.
     /// - parameters:
     ///   - period: An enum representing whether either a day, a week, a month or a year worth's of data.
--- a/WordPressKit/StatsServiceRemoteV2.swift
+++ b/WordPressKit/StatsServiceRemoteV2.swift
@@ -4,7 +4,7 @@ import Foundation
 // one and rename this to not have "V2" in it, but we want to keep the old one around
 // for a while still.
 
-public class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
+open class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
 
     public enum ResponseError: Error {
         case decodingFailure
@@ -67,7 +67,7 @@ public class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
     /// - parameters:
     ///   - referrerDomain: A referrer's domain.
     ///   - currentValue: Current value of the `isSpam` referrer's property.
-    public func toggleSpamState(for referrerDomain: String,
+    open func toggleSpamState(for referrerDomain: String,
                                 currentValue: Bool,
                                 success: @escaping () -> Void,
                                 failure: @escaping (Error) -> Void) {
--- a/WordPressKit/StatsServiceRemoteV2.swift
+++ b/WordPressKit/StatsServiceRemoteV2.swift
@@ -71,14 +71,20 @@ public class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
                                 currentValue: Bool,
                                 success: @escaping () -> Void,
                                 failure: @escaping (Error) -> Void) {
-        let path: String
-        if currentValue {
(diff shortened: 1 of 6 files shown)


WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -59,6 +59,28 @@ public class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
...
+    public func toggleSpamState(for referrerDomain: String,
+                                currentValue: Bool,
+                                success: @escaping () -> Void,
+                                failure: @escaping (Error) -> Void) {
+        let path: String
+        if currentValue {
+            path = self.path(forEndpoint: "sites/\(siteID)/stats/referrers/spam/delete?domain=\(referrerDomain)", withVersion: ._1_1)
+        } else {
+            path = self.path(forEndpoint: "sites/\(siteID)/stats/referrers/spam/new?domain=\(referrerDomain)", withVersion: ._1_1)
+        }
```
emilylaguna: Since the only difference between the 2 strings is the `delete` and `new`, what do you think about condensing this a bit?

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -59,6 +59,28 @@ public class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
...
+
+        wordPressComRestApi.POST(path, parameters: nil) { object, _ in
+            success()
```
emilylaguna, line 88: Since the API also returns a `success` value in the response what do you think about also validating that before triggering the `success()` callback?

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -59,6 +59,28 @@ public class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
...
     }
 
+    /// Used to mark or unmark referrer as spam, depending of the current value.
+    /// - parameters:
+    ///   - referrerDomain: A referrer's domain.
+    ///   - currentValue: Current value of the `isSpam` referrer's property.
+    public func toggleSpamState(for referrerDomain: String,
+                                currentValue: Bool,
+                                success: @escaping () -> Void,
+                                failure: @escaping (Error) -> Void) {
```
emilylaguna: I think being more explicit with the methods for the network is better, perhaps 2 methods for mark as spam and unmark as spam would be better. Thoughts?

emilylaguna (changes requested):
👋 @nikola-milicevic! Left some comments for you!

We should also update the unit tests for the `StatsServiceRemoteV2` class to include the new spam functionality.

WordPressKit/StatsServiceRemoteV2.swift
nikola-milicevic: I agree.

WordPressKit/StatsServiceRemoteV2.swift
nikola-milicevic: I thought API is designed to return success if the operation was successful, so double checking seemed redundant to me. But if that's not the case (if the success block can be reached even if the operation was unsuccessful) then double checking is a must.

WordPressKit/StatsServiceRemoteV2.swift
nikola-milicevic: I agree with the more explicit approach with API calls + deciding logic shouldn't be responsibility of this class, but in this state-changing logic it seems to me it only increases verbosity at the call site, with no little to no upsides.

nikola-milicevic:
> 👋 @nikola-milicevic! Left some comments for you!
> 
> We should also update the unit tests for the `StatsServiceRemoteV2` class to include the new spam functionality.

Thanks for the feedback, if you agree, my plan was to add tests at the very end, when we have everything else in place.

nikola-milicevic:
👋 @emilylaguna I updated the code. I added additional error checking as well as tests. Also, I didn't split toggle functionality into two methods, if you think that's better approach - I will refactor it.

WordPressKit/StatsServiceRemoteV2.swift
```diff
@@ -279,6 +289,26 @@ extension StatsServiceRemoteV2 {
...
 
+// MARK: - Mark referrer as spam helpers
+
+private extension StatsServiceRemoteV2 {
+    func pathForToggleSpamStateEndpoint(referrerDomain: String, markAsSpam: Bool) -> String {
+        let action = markAsSpam ? "new" : "delete"
+        return self.path(forEndpoint: "sites/\(siteID)/stats/referrers/spam/\(action)?domain=\(referrerDomain)", withVersion: ._1_1)
+    }
+
+    struct MarkAsSpamResponse {
```
emilylaguna, line 300: Nice approach, How do you feel about reducing the boiler plate by adhering this to Decodable?

emilylaguna (changes requested):
Nice work! I left 1 comment about reducing some boilerplate by using Decodable

WordPressKit/StatsServiceRemoteV2.swift
nikola-milicevic: That was my intention, though `POST` method doesn't return `Data`, it already returns a dictionary, so I think that would just introduce unneeded double conversion, if I'm not mistaken :) (from dictionary to `Data`, and then decode to struct).

emilylaguna: Makes sense. For larger objects I think this conversion makes sense but since it's small I don't think it's required.

emilylaguna (approved):
LGTM.

PR #418, opened 2021-07-12
Merge version 4.37.0 into trunk

mokagio:
Merges the 4.37.0 changes into `trunk`.

No test required other than verifying the `.podspec` changes and that CI passes.


WordPressKit/DomainsServiceRemote.swift
```diff
@@ -103,7 +103,7 @@ public class DomainsServiceRemote: ServiceRemoteWordPressComREST {
     }
 
     @objc public func getStates(for countryCode: String,
-                                success: @escaping ([State]) -> Void,
+                                success: @escaping ([WPState]) -> Void,
```
AliSoftware, line 106: Once could argue that this is a breaking change and thus would warrant a bump to `5.0.0` instead of `4.37.0`…

WordPressKit/WPCountry.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-@objc public class State: NSObject, Codable {
+@objc public class WPCountry: NSObject, Codable {
```
AliSoftware, line 3: lol at `git` considering that `State.swift` got renamed into `WPCountry.swift` and class changed to `State` -> `WPCountry`… and vice-versa for `Country` -> `WPState` on the other file… thanks `git` 🤦  (aka why do it simple when you can make it complicated 😆 )

AliSoftware (commented):
I'm not sure how strict we are with semver on our own internal pods, but since CocoaPods is pretty solid on this and dependency graph resolution highly depend on respecting semver to pull the correct versions, I think we should make sure that this breaking change makes it into a new major version instead of a minor bump. wdyt?

mokagio:
Closing in favor of #420. Sorry for dropping the ball here and not following up.

PR #425, opened 2021-07-29
Add Buildkite CI

jkmassel:
Adds Buildkite CI to this project. CircleCI was still relying on manual `xcodebuild` invocations, so this PR adopts `fastlane` to run the tests.

Note – this PR doesn't yet import the "publish" action (it's commented out) due to a lack of secrets support – this is deliberate, I wanted to have a pod implementation done first in order to figure out the secrets automation.

### Testing Details
- Ensure that the Buildkite CI passes


.buildkite/build-and-test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
```
mokagio, line 3: FYI, in [another script](https://github.com/wordpress-mobile/WordPress-iOS/pull/16806#discussion_r666077522), @AliSoftware and I have been discussing using `#!/bin/sh -euo pipefail`. The context is different from here, because that script is meant to run as a build phase from Xcode and uses `sh` instead of `bash` (for no particular reason) but might be worth considering.

Although, I did a quick google around and have found mixed opinions on whether it is a best practice or not. That's not to say that it definitely isn't, only that I don't have much experience in the matter and I'm still trying to figure it out 😄

mokagio: By the way, is there a particular reason why you chose `bash` instead of `sh` this and the other scripts?

.buildkite/lint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+echo "--- :rubygems: Setting up Gems"
+bundle install
+
+echo "--- :rubocop: Running Rubocop"
```
mokagio, line 15: Off topic, but how cool is it that Buildkite has support for all sorts of tooling emojis? 🤩

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,55 @@
...
+    env: *common_env
+    plugins: *common_plugins
+
+  #################
+  # Lint
+  #################
+  - label: "🧹 Lint"
+    command: .buildkite/lint.sh
+    plugins:
+      - automattic/bash-cache#add/cocoapods-support: ~
```
mokagio, line 36:
```suggestion
    plugins:
      - *common_plugins
```

.buildkite/validate-podspec.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e
+
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+echo "--- :cocoapods: Setting up Pods"
+install_cocoapods
+
+echo "--- :microscope: Validate Podspec"
+xcrun simctl list >> /dev/null # For some reason this fixes a failure in `lib lint`
```
mokagio: 😖 ❓ 

Do we happen to have a GitHub issue to link here? Just in case it's a problem upstream (CocoaPods?) and it'll be fixed eventually?

.gitignore
```diff
@@ -1,8 +1,21 @@
-.DS_Store
-
 # Xcode
 #
+# gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
```
mokagio: Nitpick. What do you think of this?

```suggestion
# Taken from
# https://raw.githubusercontent.com/github/gitignore/991e760c1c6d50fdda246e0178b9c58b06770b90/Swift.gitignore

# Xcode
```

Mainly, I find it confusing to read mentions to the other `.gitignore` files that are part of the [gitgnore](https://github.com/github/gitignore/) repo but not of this.

fastlane/Fastfile
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+default_platform(:ios)
+
+platform :ios do
+  desc 'Builds the project and runs tests'
+  lane :test do
+    run_tests(
```
mokagio, line 8: +1 for using `run_tests` instead of `scan`; much clearer.

mokagio (changes requested):
CircleCI noticed that the `Podfile.lock` was not regenerated after adding the frozen string literal comment to `Podfile` (`rubocop -A` I'm guessing?).

The check is not required by GitHub to merge, but I feel we should address the failure in the context of this PR.

I think we all agree that check should be added to the Buildkite pipeline. I'm not sure whether you'd want to add it as part of this PR or as a followup, because of what you wrote in the description:

> I wanted to have a pod implementation done first in order to figure out the secrets automation.

.buildkite/build-and-test.sh
oguzkocer: > By the way, is there a particular reason why you chose bash instead of sh this and the other scripts?

We have had this discussion before in Slack - this might have been before you and @AliSoftware  joined the team. Basically, I can't think of a good reason to pick `sh` over `bash` (assuming `bash` is available) as we'd just be limiting ourselves with it.

.buildkite/pipeline.yml
jkmassel: I tried this (and a few other approaches) and couldn't get it to work – Buildkite doesn't want to upload this: https://buildkite.com/wordpress-mobile/wordpresskit-ios/builds/26#2bbf8117-25b2-4c1d-88b1-c7b0c8967cc1

.gitignore
jkmassel: Oh ya this is super nice!!

jkmassel:
> CircleCI noticed that the `Podfile.lock` was not regenerated after adding the frozen string literal comment to `Podfile` (`rubocop -A` I'm guessing?).

Should be fixed by 1e1b8cbd6087ec31bbe3af0ead0fb2adc49ce05e and https://github.com/wordpress-mobile/WordPressKit-iOS/pull/425/commits/7cf664850864d3b37cfc15905377d92d1ede5439

.buildkite/validate-podspec.sh
jkmassel: This _may_ be something to do with the VM? I've opened an issue in https://github.com/Automattic/buildkite-ci/issues/7 to track it!

.buildkite/build-and-test.sh
mokagio: Fine by me 😄 I'll adopt this a standard moving forward 👍 Thanks for the extra context.

.buildkite/validate-podspec.sh
mokagio: Thanks! Then I would suggest doing this:

```suggestion
# For some reason this fixes a failure in `lib lint`
# https://github.com/Automattic/buildkite-ci/issues/7
xcrun simctl list >> /dev/null
```

But no big deal.

mokagio (commented):
Sorry about my suggestion to DRY the YAML not working 😞 . I might try to followup and sort it our later on...

mokagio (approved):
_I meant to approve but `Cmd Enter` instead of using the mouse._

jkmassel:
> Sorry about my suggestion to DRY the YAML not working

Hah no worries – I was sure it would work every time I tried something, and I tried like 5 different things _unsuccessfully_... 

Someday I'll understand YAML, but...not today 😅

.buildkite/build-and-test.sh
AliSoftware: > FYI, in another script, @AliSoftware and I have been discussing using #!/bin/sh -euo pipefail.

To be clear, the thing about `set -euo pipefail` vs setting `-euo pipefail` on the shebang… is a bit like tabs vs spaces: I don't think there's a strong rationale for one over the other, and ultimately it doesn't matter and it's ok to go with whatever you like (I certainly don't want to give the wrong impression I wanted to start a debate over this 😅). Both are equally fine and it isn't a big deal which one we choose to use.

PR #426, opened 2021-08-04
Add release build publishing

jkmassel:
### Description

Adds release build support to Buildkite (removing it from CircleCI).

### Testing Details

- Note that CI passes
- Note that [the production release succeeds ](https://buildkite.com/wordpress-mobile/wordpresskit-ios/builds/68#_)

### DIFF
--- a/.buildkite/lint.sh
+++ b/.buildkite/lint.sh
@@ -1,30 +1,15 @@
-#!/bin/bash
-
-set -e
+#!/bin/bash -eu
 
 echo "--- :rubygems: Setting up Gems"
 bundle install
 
 echo "--- :rubygems: Checking Gemfile.lock"
-# This will only work if it goes after `bundle install` – don't try to run it first
-if [ -n "$(git status | grep modified | grep Gemfile.lock)" ]; then 
-	echo "Error: Gemfile.lock is not in sync – please run \`bundle install\` and commit your changes"
-	exit 1
-fi 
+validate_gemfile_lock
 
 echo "--- :rubocop: Running Rubocop"
 bundle exec rubocop
 
-echo "Gemfile.lock is in sync"
-
 echo "--- :cocoapods: Checking Podfile.lock"
-PODFILE_SHA1=$(ruby -e "require 'yaml';puts YAML.load_file('Podfile.lock')['PODFILE CHECKSUM']")
-RESULT=$(echo "$PODFILE_SHA1 *Podfile" | shasum -c)
-
-# This will only work if it goes after `pod install` – don't try to run it first
-if [[ $RESULT != "Podfile: OK" ]]; then 
-	echo "Error: Podfile.lock is not in sync – please run \`bundle exec pod install\` and commit your changes"
-	exit 1
-fi 
+validate_podfile_lock
 
 # TODO: Add swiftlint
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,7 +2,7 @@
 common_params:
   # Common plugin settings to use with the `plugins` key.
   - &common_plugins
-    - automattic/bash-cache#add/cocoapods-support: ~
+    - automattic/bash-cache#add/pod-publishing: ~
   # Common environment values to use with the `env` key.
   - &common_env
     IMAGE_ID: xcode-12.5.1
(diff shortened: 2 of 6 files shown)


.buildkite/pipeline.yml
```diff
@@ -42,14 +43,15 @@ steps:
...
-  #  env: *common_env
-  #  plugins: *common_plugins
-  #  depends_on:
-  #    - "test"
-  #    - "validate"
-  #    - "lint"
-  #  if: build.tag != null
+  - label: "⬆️ Publish Podspec"
+    key: "publish"
+    command: .buildkite/publish-pod.sh
```
AliSoftware, line 47: ~If we enable pod publishing from BuildKite, we should probably disable it from CircleCI's config at the same time / in the same PR, so that we don't end up in a state where both pipelines are active and we try to push the podspec to trunk both from BuildKite and CircleCI~
[EDIT] This comment was written when I reviewed the PR while it was still in draft, but newer commits have since been pushed to fixed that, so we're good now 👍

.buildkite/pipeline.yml
```diff
@@ -31,9 +31,10 @@ steps:
   # Lint
   #################
   - label: "🧹 Lint"
+    key: "lint"
     command: .buildkite/lint.sh
     plugins:
-      - automattic/bash-cache#add/cocoapods-support: ~
+      - automattic/bash-cache#v1.1.0: ~
       - docker#v3.8.0:
           image: "public.ecr.aws/automattic/multilint:latest"
```
AliSoftware: I see that you don't reuse the `*common_plugins` here – and indeed you can't, because `&common_plugins` is an array (of one item), and here you need to add more items to it (and YAML doesn't have a syntax to merge an anchor representing an array with other array items like it has with `<<:` for dictionaries).

In order to avoid repeating the plugin (and more importantly avoid risking to update one plugin call site but forget to update the other… like you actually just did, because you use 1.1.0 here but 1.2.1 in the anchor), I suggest instead that you make your anchor not reference the array (containing only one plugin), but instead just the `bash-cache` plugin entry on its own:

```yaml
common_params:
  plugins:
    - &bash_cache_plugin
      automattic/bash-cache#v1.1.0: ~
  env: &common_env
    IMAGE_ID: xcode-12.5.1
```

Then create the array at "call site" instead, referencing the bash-cache entry… and others when needed.

So when you only need that plugin and nothing else:
```yaml
  - label: "🔬 Validating Podspec"
    key: "validate"
    command: .buildkite/validate-podspec.sh
    env: *common_env
    plugins:
      - *bash_cache_plugin
```

And for cases like here where you need more than one plugin:

```yaml
  - label: "🧹 Lint"
    key: "lint"
    command: .buildkite/lint.sh
    plugins:
      - *bash_cache_plugin
      - docker#v3.8.0:
          image: "public.ecr.aws/automattic/multilint:latest"
```

.buildkite/lint.sh
```diff
@@ -12,11 +12,11 @@ if [ -n "$(git status | grep modified | grep Gemfile.lock)" ]; then
 	exit 1
 fi 
 
+echo "Gemfile.lock is in sync"
```
jkmassel, line 15: This was showing up in the wrong place in the output

.buildkite/pipeline.yml
```diff
@@ -31,9 +31,10 @@ steps:
   # Lint
   #################
   - label: "🧹 Lint"
+    key: "lint"
     command: .buildkite/lint.sh
     plugins:
-      - automattic/bash-cache#add/cocoapods-support: ~
+      - automattic/bash-cache#v1.1.0: ~
```
AliSoftware: version mismatch with other steps here – because you're not using the anchor to DRY it 😜 . See comment below for suggestions (+ Anchor crash course Slack thread).

AliSoftware (changes requested):

jkmassel:
@AliSoftware – reworked the config for DRY based on your excellent tutorial!

Also [re-ran the tag workflow](https://buildkite.com/wordpress-mobile/wordpresskit-ios/builds/80#f4357e65-a61c-42bb-8cb9-1c758e7f232c) just in case and it worked great.

Ready for another review when you're able, thanks!

AliSoftware (approved):
✨

mokagio:
This run for [4.39.0](https://github.com/wordpress-mobile/WordPressKit-iOS/releases/tag/4.39.0) nice and smoothly.

PR #430, opened 2021-08-06
Update deployment target to iOS 13

ealeksandrov:
## Description

This PR updates deployment target to iOS 13 in line with other libraries, WP-iOS and WC-iOS.
The main reason for immediate upgrade is CocoaPods issue here - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/612#issuecomment-893763927

It also fixes all new deprecation warnings.

Due to issues with "boot simulator" CI step I had to also update Xcode and iOS versions in CircleCI config.

### Testing Details

Build and run unit tests. Please check code for deprecation updates closely since they are **not** covered by unit tests.


WordPressKit.xcodeproj/project.pbxproj
```diff
@@ -3519,7 +3516,6 @@
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				INFOPLIST_FILE = WordPressKit/Info.plist;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
```
mokagio, line 3522: It's not clear in this diff, but by opening the project in Xcode I saw you decided to remove the target-level deployment setting, making them all use the project-level one.

Great idea 👍

WordPressKit/Scanner+extensions.swift
```diff
@@ -3,13 +3,13 @@ import Foundation
...
         let quoteString = "\""
         while self.isAtEnd == false {
-            scanUpTo(quoteString, into: nil)
-            scanString(quoteString, into: nil)
-            scanUpTo(quoteString, into: &quote)
-            scanUpTo(quoteString, into: nil)
+            _ = scanUpToString(quoteString)
+            _ = scanString(quoteString)
+            quote = scanUpToString(quoteString)
+            _ = scanUpToString(quoteString)
```
mokagio, line 12: This change looks good.

It's understandably out of scope for this PR, I wish there had been already a test to verify it 😞

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -330,8 +330,10 @@ extension WordPressOrgXMLRPCApi {
                 return
             }
 
+            var evaluationError: CFError?
+            _ = SecTrustEvaluateWithError(serverTrust, &evaluationError)
             var result = SecTrustResultType.invalid
-            let certificateStatus = SecTrustEvaluate(serverTrust, &result)
+            let certificateStatus = SecTrustGetTrustResult(serverTrust, &result)
```
mokagio: Again, I don't think improving this code is necessarily in scope for this PR, but I noticed two things.

- `evaluationError` is unused. I guess we don't get a warning from the compiler because of the assignment in line 334, but the code itself never reads it. Is there any kind of error handling that we should do here? Maybe previous authors have more insight on the kind of treatment this code deserves? cc @jleandroperez @astralbodies 
- While there is a test for the class (🎉) it might not be enough to cover this scenario. I overrode `result` with `.invalid` after this line and the test still passed.

mokagio (approved):
Thank you @ealeksandrov 🙌 

I particularly appreciate how you went ahead and updated everything that was necessary and like your 1 warning fix per commit approach 👍 

---

Sorry for not getting onto this earlier. Now you have a little conflict on the pod version because I [released 4.39.0 stable](#431) as part of the WordPress iOS 18.0 code freeze.

dvdchr (approved):
LGTM! thank you for updating this. :shipit: 

(Note: as Gio has mentioned, since `4.39.0` has been released here's a friendly reminder to update the pod version to `4.40.0-beta.1` 🙂 )

WordPressKit/Scanner+extensions.swift
ealeksandrov: I agree. I've spent some time digging in and:
- Looks like this extension is broken (or I don't understand the logic behind it in my tests).
- WP-iOS has its own very similar extensions for [NSScanner](https://github.com/wordpress-mobile/WordPress-iOS/blob/a3eae1d75066b28c215180cafd12e4947c74e556/WordPress/Classes/Categories/NSScanner+Helpers.m) and [NSAttributedString](https://github.com/wordpress-mobile/WordPress-iOS/blob/a3eae1d75066b28c215180cafd12e4947c74e556/WordPress/Classes/Categories/NSAttributedString+Util.m).
- This `Scanner` extension used only in `NSMutableAttributedString` extension which is unused in library and declared [internal](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/22a68b62053d9a268a4fd093b205b83e53b49bf6/WordPressKit/NSMutableAttributedString%2Bextensions.swift#L3).

I've decided to decouple this issue into separate branch (and remove duplicated code from WP there):
https://github.com/wordpress-mobile/WordPressKit-iOS/tree/update/quotes-attributes-extension

WordPressKit/WordPressOrgXMLRPCApi.swift
ealeksandrov: Right, I was also experimenting with it, but decided to keep exact same logic as before. Any additional input on this would be nice!
Removed unused error var in 4669e9f671bede92c2be24cb8c2e9940d83c315e.

ealeksandrov:
Thanks @mokagio and @dvdchr! I've added few updates to dev dependencies as part of general cleanup and will merge it as 4.40.1-beta.1 🎉

WordPressKit/Scanner+extensions.swift
mokagio: Great digging! Keep me posted on in please. Now I'm curious to see how the cleanup will evolve 🍿

WordPressKit/WordPressOrgXMLRPCApi.swift
jleandroperez: Hey there folks! I'm afraid I can't recall working on this code, too many years :(

That being said, if there's a Trust evaluation error, I doubt there's much we can do. Fallback to "default handling", and the URL Request should fail IMO

WordPressKit/Scanner+extensions.swift
ealeksandrov: Latest update: unused extension removed from lib in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/449 + WP part refactored with tests in https://github.com/wordpress-mobile/WordPress-iOS/pull/17259

PR #433, opened 2021-08-09
Recommend App: Add service to fetch link from remote

dvdchr:
### Description

Refs: https://github.com/wordpress-mobile/WordPress-iOS/issues/16995, `pctCYC-9r-p2`, `pctCYC-8X-p2#comment-325`

This change introduces a `ShareAppContentServiceRemote` that fetches the link and text to be used as the content to be shared.

### Testing Details

Ensure that the tests pass:
- Contents from the API should be correctly parsed into `RemoteShareAppContent`.
- When encountering unexpected structure, failure Result type should be returned.
- When encountering any other errors (e.g. 500, no connection), failure Result type should be returned.
---


WordPressKit.podspec
```diff
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '4.39.0'
+  s.version       = '4.40.0-beta.2'
```
dvdchr, line 5: Note: #430 will be assigning `4.40.0-beta.1`.

WordPressKit/ShareAppContentServiceRemote.swift
```diff
@@ -0,0 +1,38 @@
+/// Encapsulates logic for fetching content to be shared by the user.
+///
+open class ShareAppContentServiceRemote: ServiceRemoteWordPressComREST {
+    /// Fetch content to be shared by the user, based on the provided `appName`.
+    ///
+    /// - Parameters:
+    ///   - appName: A string that identifies the app to be shared.
+    ///   - completion: A closure that will be called when the fetch request completes.
+    open func getContent(for appName: String, completion: @escaping (Result<RemoteShareAppContent, Error>) -> Void) {
```
ScoutHarris: Hey @dvdchr .

I think it would be nice to have an `enum` for the `appName` instead of just a `String`. Maybe something like:

```
public enum ShareAppName: String {
    case wordpress = "wordpress"
    case jetpack = "jetpack"
}
```

That way, it would be restricted to known options, and it would be clear what those options are. What do you think?

dvdchr: I previously considered whether it's better to use a String enum, or to leave it stringly-typed to leave the context out of the kit repo – it was kind of a 50:50 for me. But I've no problem adjusting this to use enum instead. 🙂

dvdchr: Addressed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/433/commits/08deac22ecf1aee9219619c9839707a02a0a1c8a !

ScoutHarris (approved):
Looks good!

:shipit:

PR #434, opened 2021-08-10
Use plugin actions

jkmassel:
### Description

Improves the DRY-ness between repos for pod work. The `Publish Podspec` task is still per-repo, because some repos use the private specs repo and others don't – in this case trying to DRY it might just turn it into spaghetti.

### Testing Details

Note that CI passes, and that the deleted code has been moved to https://github.com/Automattic/bash-cache-buildkite-plugin/commit/f8b36025681c77f4b5d119e99662f1fd2aa4143b


.buildkite/pipeline.yml
```diff
@@ -31,13 +31,9 @@ steps:
...
-    command: .buildkite/lint.sh
-    plugins:
-      - *bash_cache
-      - docker#v3.8.0:
-          image: "public.ecr.aws/automattic/multilint:latest"
-    agents:
-      queue: "default"
+    command: "lint_pod"
+    env: *common_env
+    plugins: *common_plugins
```
jkmassel, line 36: I moved this back to the macOS hosts because docker images and library plugins don't play nicely together – I have a workaround, but it'd require a new docker image, so this was a faster way to prove that this works (also it might've been an over-optimization...)

.buildkite/pipeline.yml
```diff
@@ -1,7 +1,7 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   plugins: &common_plugins
-  - &bash_cache automattic/bash-cache#v1.2.1: ~
+  - &bash_cache automattic/bash-cache#add/pod-actions: ~
```
mokagio: This branch doesn't exist anymore, does it?

I created a branch on top of this, see below, and CI [failed](https://buildkite.com/wordpress-mobile/wordpresskit-ios/builds/101#604d01ca-1cea-450d-975f-fcfbc5df7173/4-20) with:

> error: pathspec 'add/pod-actions' did not match any file(s) known to git

.buildkite/pipeline.yml
```diff
@@ -13,7 +13,7 @@ steps:
   #################
   - label: "🧪 Build and Test"
     key: "test"
-    command: .buildkite/build-and-test.sh
+    command: "build_and_test_pod"
```
mokagio, line 16: Is there a style reason for using the quotes?

I [verified](https://github.com/wordpress-mobile/WordPressKit-iOS/compare/dry/buildkite-actions...mokagio/434-test) that the commands run fine without quotes, too.

.buildkite/pipeline.yml
mokagio: Maybe this?

```suggestion
  - &bash_cache automattic/bash-cache#v1.3.2: ~
```

I can see the "Add pods actions" [commit](https://github.com/Automattic/bash-cache-buildkite-plugin/commit/6adb942a93282d2bf35aa6dc39eb7b613b08de9d) is part of that tag.

mokagio (changes requested):
See bash cache plugin version comment. Requesting changes to prevent a merge that could break CI.

.buildkite/pipeline.yml
jkmassel: Great catch, thanks!!

.buildkite/pipeline.yml
jkmassel: Good question – without a linter for these, I sort of just "do what feels good" – I was basically doing "no quotes for paths, quotes for commands", but that might not be correct/necessary 🤔

Perhaps something to add to our YAML style guide? 😅

mokagio: > a linter for these

🤩 thank you for giving me even more ideas for low end-user value-generating things that I know will want to work on but won't do because there's more important stuff to do first! 🤣

mokagio (approved):
Now that the plugin points to a tag, everything seems good 👍

PR #437, opened 2021-08-24
Merge release 4.40.0 into trunk

mokagio:
Merges the 4.40.0 changes into `trunk`.

No test required other than verifying the `.podspec` changes and that CI passes.


WordPressKitTests/Mock Data/me-settings-close-account-failure.json
```diff
@@ -0,0 +1,4 @@
+{
+    "error": "active-subscriptions"
+    "message": "This user account cannot be closed while it has active subscriptions."
+}
```
AliSoftware, line 4: As the GitHub syntax highlighting suggests, there's an error in this JSON, with a missing comma on line 2.

Not sure if that's intentional (e.g. to use it as mock data to test invalid JSON returned by the server) but I don't think so? So might be worth investigating and fixing in a future PR.

AliSoftware (approved):
Version bump and commit list make sense.

Left a comment about an invalid JSON in mocks that I think needs addressing in a future PR.

PR #446, opened 2021-09-15
Reader cross post parsing: fix post url format

ScoutHarris:
### Description

Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/17158
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/17159

When a cross post's `postURL` is created, there was an unnecessary "/" in the string format, causing the URL to have "//". This removes that extra "/". 

In addition, the check for `url` didn't actually do anything, so it was possible for cross post `siteURL` and `postURL` to be junk. Now, if there is no `url`, the cross post URLs will not be populated.

### Testing Details

### DIFF
--- a/WordPressKit/RemoteReaderPost.m
+++ b/WordPressKit/RemoteReaderPost.m
@@ -192,14 +192,12 @@ static const NSUInteger ReaderPostTitleLength = 30;
 
             NSString *path = [obj stringForKey:CrossPostMetaValue];
             NSURL *url = [NSURL URLWithString:path];
-            if (!url) {
-                NSLog(@"break");
-            }
-
-            meta.siteURL = [NSString stringWithFormat:@"%@://%@", url.scheme, url.host];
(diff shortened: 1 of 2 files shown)


WordPressKit/RemoteReaderPost.m
```diff
@@ -192,14 +192,12 @@ - (RemoteReaderCrossPostMeta *)crossPostMetaFromPostDictionary:(NSDictionary *)d
 
             NSString *path = [obj stringForKey:CrossPostMetaValue];
             NSURL *url = [NSURL URLWithString:path];
-            if (!url) {
-                NSLog(@"break");
```
aerych, line 196: 💥

aerych (approved):
Howdy @ScoutHarris 👋 
Tested with the companion wpios patch.  Looks good!
:shipit: !

PR #449, opened 2021-10-04
Remove unused extensions

ealeksandrov:
### Description

2 extensions methods `scanQuotesText()` and `applyAttributes(toQuotes:)` are declared internal - it means they can't be used outside of this library. I found similar code in WP-iOS and refactored it, adding tests (https://github.com/wordpress-mobile/WordPress-iOS/pull/17259).

If discussed code will be shared/reused - we can move extensions+tests here later.

I didn't add a new version since removing internal methods doesn't have any impact on consumer apps.

### Testing Details

Check that CI is green and nothing broke in tests.


dvdchr (approved):
LGTM :shipit:

PR #452, opened 2021-10-14
Reader Post: add endpoint to update notification settings for subscriptions

dvdchr:
### Description

Refs: https://github.com/wordpress-mobile/WordPress-iOS/pull/17322

- Added `receivesCommentNotifications` property to `RemoteReaderPost`.
- Added method in `ReaderPostServiceRemote+Subscriptions` to update notification settings for a post subscription.

### Testing Details

Ensure that all the tests are 🆗 . Manual testing instructions can be done following the instructions described in https://github.com/wordpress-mobile/WordPress-iOS/pull/17322.

### DIFF
--- a/WordPressKit.xcodeproj/project.pbxproj
+++ b/WordPressKit.xcodeproj/project.pbxproj
@@ -588,6 +588,7 @@
 		FAD1344525908F5F00A8FEB1 /* JetpackBackupServiceRemote.swift in Sources */ = {isa = PBXBuildFile; fileRef = FAD1344425908F5F00A8FEB1 /* JetpackBackupServiceRemote.swift */; };
 		FAD1344B259094C300A8FEB1 /* JetpackBackup.swift in Sources */ = {isa = PBXBuildFile; fileRef = FAD1344A259094C300A8FEB1 /* JetpackBackup.swift */; };
 		FAD1345125909DEA00A8FEB1 /* JetpackBackupServiceRemoteTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = FAD1345025909DEA00A8FEB1 /* JetpackBackupServiceRemoteTests.swift */; };
+		FEB7A88F271873BD00A8CF85 /* reader-post-comments-update-notification-success.json in Resources */ = {isa = PBXBuildFile; fileRef = FEB7A88E271873BD00A8CF85 /* reader-post-comments-update-notification-success.json */; };
 		FEFFD99126C1347D00F34231 /* ShareAppContentServiceRemote.swift in Sources */ = {isa = PBXBuildFile; fileRef = FEFFD99026C1347D00F34231 /* ShareAppContentServiceRemote.swift */; };
 		FEFFD99326C141A800F34231 /* RemoteShareAppContent.swift in Sources */ = {isa = PBXBuildFile; fileRef = FEFFD99226C141A800F34231 /* RemoteShareAppContent.swift */; };
 		FEFFD99726C158F400F34231 /* share-app-content-success.json in Resources */ = {isa = PBXBuildFile; fileRef = FEFFD99626C158F400F34231 /* share-app-content-success.json */; };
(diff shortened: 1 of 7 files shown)


WordPressKit.xcodeproj/project.pbxproj
```diff
@@ -1910,28 +1912,7 @@
...
-				40819774221E497C00A298E4 /* stats-published-posts.json */,
-				404057DB221C9FD70060250C /* stats-referrer-data.json */,
 				937250EB267A15060086075F /* stats-referrer-mark-as-spam.json */,
-				404057D7221C98690060250C /* stats-clicks-data.json */,
-				404057D3221C5FC30060250C /* stats-countries-data.json */,
-				404057CF221C46780060250C /* stats-videos-data.json */,
-				404057CA221B80BC0060250C /* stats-top-authors.json */,
-				40F9880D221ACFB400B7B369 /* stats-streak-result.json */,
-				404057C6221B36070060250C /* stats-search-term-result.json */,
-				984E34F322EF9464005C3F92 /* stats-file-downloads.json */,
```
dvdchr, line 1934: I did not delete these 🤔 . Must've been relocated to different lines since I added one mock data and placed it in the right slot (alphabetically).

WordPressKitTests/ReaderPostServiceRemote+SubscriptionTests.swift
```diff
@@ -101,4 +103,66 @@ class ReaderPostServiceRemoteSubscriptionTests: RemoteTestCase, RESTTestable {
...
+                                                                  siteID: siteID,
+                                                                  receiveNotifications: receiveNotifications,
+                                                                  success: {
+            // the boolean result should match the receiveNotifications property passed in the params.
+            expect.fulfill()
+        },
+                                                                  failure: { error in
+            XCTFail("This callback shouldn't get called")
+            expect.fulfill()
+        })
```
dvdchr, line 124: If you noticed about this: I'm not sure why the closures are indented this way. It's all Xcode 13. I am powerless! 😅

WordPressKit.xcodeproj/project.pbxproj
aerych: These files are still showing up in the project inspector, and `ack` shows they are still referenced in other places in the project.pbxproj.  The test suite builds and runs fine and all tests are passing.  
Looking at the [git blame for the previous history of the file](https://github.com/wordpress-mobile/WordPressKit-iOS/blame/8b531cc04e24d65fc692394aa077f277ed6f35d9/WordPressKit.xcodeproj/project.pbxproj#L1913) shows that most of these were references added years ago, and if we look through the list of items for `Mock Data` we see that these are repeated later in the list. 
Seems like this is probably Xcode cleaning up stale references?

WordPressKit/ReaderPostServiceRemote+Subscriptions.swift
```diff
@@ -100,4 +104,42 @@ extension ReaderPostServiceRemote {
...
+    /// `unsubscribeFromPost` method.
+    ///
+    /// - Parameters:
+    ///   - postID: The ID of the post.
+    ///   - siteID: The ID of the site.
+    ///   - receiveNotifications: When the value is true, subscriber will also receive a push notification for new comments on the subscribed post.
+    ///   - success: Closure called when the request has succeeded.
+    ///   - failure: Closure called when the request has failed.
+    @objc open func updateNotificationSettingsForPost(with postID: Int,
+                                                      siteID: Int,
```
aerych, line 122: I was trying to think of any scenario where we wouldn't have a site ID but would have the domain and whether a string should also be supported here. 
Can't think of a single one. :D   LGTM!

aerych (approved):
Howdy @dvdchr 👋 
Changes look good to me!  Tested with the other PR. 
:shipit: !

WordPressKit.xcodeproj/project.pbxproj
dvdchr: > Seems like this is probably Xcode cleaning up stale references?

Totally makes sense. Thanks for digging through it 🙂 . It's just scary to me how Xcode randomly deleted lines like this. Of course, I tested it locally and nothing went wrong. But usually, it works for me but not for everybody else. 😂

WordPressKit/ReaderPostServiceRemote+Subscriptions.swift
dvdchr: > I was trying to think of any scenario where we wouldn't have a site ID

Hmm, good point... I'll keep this in mind. I dug into it a little bit and looks like the only route where the reader detail view is instantiated with _only_ a post URL happens when clicking a WordPress.com blog post from a Reader detail screen.

However, I can't seem to replicate this behavior as I ended up opening a web view instead... So I dug some more, I noticed that fetching the `ReaderPost` by URL failed:

Screen Shot 2021-10-15 at 17 54 05

This could be a bug in `ReaderPostService`, and I'll log this issue in WP-iOS later on. However, my assumption is that the case isn't that high as the reproduction step is by having a link on a reader post that points to a WP.com blog post (ones that have the `year/month/date/slug` path pattern, for example: https://wordpress.com/blog/2021/10/07/behind-the-scenes-tech-stack-wordpress-com-growth-summit/). I can iterate on this later on if we think we need to handle this case properly.

Appreciate your help in reviewing this! 👍

PR #455, opened 2021-10-19
Adds product endpoint and improves domains endpoint.

diegoreymendez:
### Description

Adds a new endpoint for retrieveing products information and unifies two existing service calls in the domains endpoint.

### Testing Details

Follow the testing instructions in the WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/17336

### DIFF
--- a/WordPressKit/ProductServiceRemote.swift
+++ b/WordPressKit/ProductServiceRemote.swift
@@ -1,0 +1,30 @@
+import Combine
+import Foundation
+
+/// Provides information about available products for user purchases, such as plans, domains, etc.
+///
+open class ProductServiceRemote {
+    let serviceRemote: ServiceRemoteWordPressComREST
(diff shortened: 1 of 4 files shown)


WordPressKit/ProductServiceRemote.swift
```diff
@@ -0,0 +1,83 @@
+import Combine
```
Gio2018: Are we planning to use combine in the future?

Gio2018 (approved):
Just one question, for the rest :shipit:

WordPressKit/ProductServiceRemote.swift
diegoreymendez: Ah, some remaining code from an early attempt to use Combine.  I'll remove it.

PR #456, opened 2021-10-20
Domains: Add renewal and expiry properties

frosty:
This PR adds properties for domain name renewal and expiry that we receive via a domains API request.

To test, please see the associated WPiOS PR.

### DIFF
--- a/WordPressKit/DomainsServiceRemote.swift
+++ b/WordPressKit/DomainsServiceRemote.swift
@@ -269,7 +269,20 @@ private func mapDomainsResponse(_ response: AnyObject) throws -> [RemoteDomain] 
                 throw DomainsServiceRemote.ResponseError.decodingFailed
         }
 
-        return RemoteDomain(domainName: domainName, isPrimaryDomain: isPrimary, domainType: domainTypeFromDomainJSON(domainJson))
+        let autoRenewing = domainJson["auto_renewing"] as? Bool
+        let autoRenewalDate = domainJson["auto_renewal_date"] as? String
+        let expirySoon = domainJson["expiry_soon"] as? Bool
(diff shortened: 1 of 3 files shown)


WordPressKit/DomainsServiceRemote.swift
```diff
@@ -269,7 +269,20 @@ private func mapDomainsResponse(_ response: AnyObject) throws -> [RemoteDomain]
                 throw DomainsServiceRemote.ResponseError.decodingFailed
         }
 
-        return RemoteDomain(domainName: domainName, isPrimaryDomain: isPrimary, domainType: domainTypeFromDomainJSON(domainJson))
+        let autoRenewing = domainJson["auto_renewing"] as? Bool
+        let autoRenewalDate = domainJson["auto_renewal_date"] as? String
```
diegoreymendez, line 273: Storing this string as a date does give us a simple immediate solution, but we lose localization and formatting options for the dates when they're shown on screen.

I was just checking the server response and strangely enough it seems that dates are handled using a format like so:

```"auto_renewal_date": "October 22, 2021",```

Any chance we can store those as dates anyway?

diegoreymendez: Actually, I just found out from the endpoint that this is not the case as the endpoint call uses a locale param which changes the language of the dates.

So this is a non-issue.

diegoreymendez (approved):

WordPressKit/DomainsServiceRemote.swift
frosty: 👍  Thanks, I went down a similar route and should've explained why I just went with the string value!

PR #472, opened 2021-12-10
WordPressComRestApi: add GET and POST `async` methods

leandroalonso:
### Description

This PR adds GET and POST methods that use the new `async` API — thus only available for iOS 15.0+.

It was added:

* GET request that supports `async`
* GET requests that convert the response to a `Decodable` property also using `async`
* POST request that supports `async`
* POST requests that convert the response to a `Decodable` property also using `async`

The unit tests offer a good explanation of the API and the function.

**Important**: we can't merge this PR without using the Xcode 13 CI image, since this is not part of this task I didn't added in this PR.


WordPressKitTests/WordPressComRestApiTests+AsyncAwait.swift
```diff
@@ -0,0 +1,94 @@
...
+class WordPressComRestApiAsyncAwaitTests: XCTestCase {
+
+    private let url = "https://public-api.wordpress.com/rest/v1.1/sites/0/media/"
+
+    override func setUp() {
+        super.setUp()
+    }
+
+    override func tearDown() {
+        super.tearDown()
```
mokagio: Nitpick: In WordPress iOS, we [decided to remove the calls to `super`](https://github.com/wordpress-mobile/WordPress-iOS/issues/17569) for setup and tear down as they don't do anything, apparently.

```suggestion
    override func tearDown() {
```

WordPressKitTests/WordPressComRestApiTests+AsyncAwait.swift
```diff
@@ -0,0 +1,94 @@
...
+            return HTTPStubsResponse(error: notConnectedError)
+        }
+
+        do {
+            let api = WordPressComRestApi(oAuthToken: "fakeToken")
+            let (_, _) = try await api.get(url)
+            XCTFail("Expected to throw while awaiting, but succeeded")
+        } catch {
+            XCTAssertEqual(error.localizedDescription, "The operation couldn’t be completed. (NSURLErrorDomain error -1011.)")
+        }
```
mokagio: I was hoping we could use `XCTAssertThrowsError` here, but it doesn't compile because of "'async' call in an autoclosure that does not support concurrency". 😢

WordPressKitTests/WordPressComRestApiTests+AsyncAwait.swift
```diff
@@ -0,0 +1,94 @@
...
+
+    func testBadServerResponseGetCall() async throws {
+        stub(condition: isAbsoluteURLString(url)) { _ in
+            let notConnectedError = NSError(domain: NSURLErrorDomain, code: URLError.badServerResponse.rawValue)
+            return HTTPStubsResponse(error: notConnectedError)
+        }
+
+        do {
+            let api = WordPressComRestApi(oAuthToken: "fakeToken")
+            let (_, _) = try await api.get(url)
```
mokagio: Nitpick. Since this is ignored, we can streamline to:
```suggestion
            _ = try await api.get(url)
```

WordPressKitTests/WordPressComRestApiTests+AsyncAwait.swift
```diff
@@ -0,0 +1,94 @@
...
+
+        do {
+            let api = WordPressComRestApi(oAuthToken: "fakeToken")
+            let _ = try await api.get(url, ofType: AnyModel.self)
+        } catch {
+            XCTAssertNotNil((error as? DecodingError))
+        }
+    }
+
+    func testSuccessfullGetCodableCalssl() async throws {
```
mokagio: I think there's a typo here

```suggestion
    func testSuccessfullPostCodableCall() async throws {
```

WordPressKit/WordPressComRestApi.swift
```diff
@@ -421,6 +422,100 @@ open class WordPressComRestApi: NSObject {
...
+    /**
+     Executes a GET request to the specified endpoint
+
+     - parameter url: the url `String` to be requested
+     - parameter Type: an entity that conforms with `Codable`
+
+     - returns: a `Decodable` object of the `Type` given
+     */
+    @available(iOS 15.0.0, *)
+    public func get<T: Decodable>(_ urlString: String, ofType: T.Type) async throws -> T {
```
mokagio: Have you considered omitting the `ofType` parameter and having the caller help the compiler understand what `T` is?

What I'm suggesting is to modify the function signature to:

```suggestion
    public func get<T: Decodable>(_ urlString: String) async throws -> T {
```

And then call it by explicitly defining the type in the assignments, e.g. in the tests like:

```swift
let anyModel: AnyModel = try await api.get(url)
```

Personally, I find this option more appealing because it doesn't require an extra parameter.

WordPressKit/WordPressComRestApi.swift
```diff
@@ -421,6 +422,100 @@ open class WordPressComRestApi: NSObject {
         let fileURL = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent(fileName)
         return fileURL
     }
+
+    // MARK: - Async
+
+    private lazy var urlSession: URLSession! = {
```
mokagio: The implicitly unwrapped optional type definition shouldn't be necessary (compiles on my end). Is there a particular reason why we need it?

```suggestion
    private lazy var urlSession: URLSession = {
```

WordPressKit/WordPressComRestApi.swift
```diff
@@ -421,6 +422,100 @@ open class WordPressComRestApi: NSObject {
...
+        let configuration = URLSessionConfiguration.default
+
+        var additionalHeaders: [String: AnyObject] = [:]
+        if let oAuthToken = self.oAuthToken {
+            additionalHeaders["Authorization"] = "Bearer \(oAuthToken)" as AnyObject?
+        }
+        if let userAgent = self.userAgent {
+            additionalHeaders["User-Agent"] = userAgent as AnyObject?
```
mokagio: Since we know from the `if let` succeeding into this code path that the two values are not nil, do we need to cast to `AnyObject?` or can we use drop nullability with `AnyObject`?

```suggestion
            additionalHeaders["Authorization"] = "Bearer \(oAuthToken)" as AnyObject
        }
        if let userAgent = self.userAgent {
            additionalHeaders["User-Agent"] = userAgent as AnyObject
```

WordPressKit/WordPressComRestApi.swift
```diff
@@ -592,3 +687,28 @@ private extension WordPressComRestApi {
...
+
+// MARK: - POST encoding
+
+private extension Dictionary {
+    func percentEncoded() -> Data? {
+        return map { key, value in
+            let escapedKey = "\(key)".addingPercentEncoding(withAllowedCharacters: .urlQueryValueAllowed) ?? ""
+            let escapedValue = "\(value)".addingPercentEncoding(withAllowedCharacters: .urlQueryValueAllowed) ?? ""
+            return escapedKey + "=" + escapedValue
+        }
```
mokagio: Nitpick. In the case where either percent encoding fails, we may end up with something inconsistent (e.g. a `=` by itself, or `foo=` or `=bar`.

What do you think of this version that uses `compactMap` to filter out those occurrences?

```suggestion
return compactMap { key, value in
            guard
                let escapedKey = "\(key)".addingPercentEncoding(withAllowedCharacters: .urlQueryValueAllowed),
                let escapedValue = "\(value)".addingPercentEncoding(withAllowedCharacters: .urlQueryValueAllowed)
            else {
                return .none
            }

            return escapedKey + "=" + escapedValue
        }
```

It compiles and the tests pass locally for me, but I don't see an explicit test for different behavior of this new method, so there could be something I lost in translation.

mokagio (approved):
Looking good! Love to see some async/await. 😄 💙 

Left only non blocking suggestions and nitpicks.

WordPressKitTests/WordPressComRestApiTests+AsyncAwait.swift
leandroalonso: TIL! Fixed. :)

WordPressKit/WordPressComRestApi.swift
leandroalonso: Weird! I swear I tried that and the compiler was giving me errors... I 100% prefer that too, changed.

leandroalonso:
@mokagio thanks for the review! I addressed all the comments.

However, I'll have to wait for the Xcode 13 image before merging it.

mokagio:
> However, I'll have to wait for the Xcode 13 image before merging it.

This should help: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/473

PR #473, opened 2022-01-10
Use Xcode 13.0 image in Buildkite

mokagio:
### Description

Long overdue CI update to use the latest (major version of) Xcode. Prompted by [this comment](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/472#issuecomment-1004157146).

**Update**: This turned out to be a more involved update than I expected. I run into [a Bundler-related CI issue which I had already experienced](https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16) and had to fiddler a bit with the pipeline settings to get out of it. The result: now the repo also runs on the latest Bundler, Ruby 2.x, and bash cache Buildkite plugin.

### Testing Details

If CI is green, then we're good.

- [x] Please check here if your pull request includes additional test coverage. N.A.


.gitignore
```diff
@@ -85,6 +85,7 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots/**/*.png
 fastlane/test_output
+fastlane/README.md
```
mokagio, line 88: This mustn't have been done before and no one ever run the tests locally via Fastlane, which is something I did investigating a CI failure.

Sort of out of scope, but small enough to fit in here anyway.

leandroalonso (approved):
:shipit: 

Thanks for tackling that @mokagio!

PR #474, opened 2022-01-16
Adds support for marking multiple notifications as read/unread

alpavanoglu:
### Description

Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/17135

This PR adds support for marking multiple notifications as read or unread. Instead of providing a single NotificationID,
we can now also provide an array of IDs. And Bumps the version to `4.46.0-beta.1`.

### Testing Details
* As the already existing `updateReadStatus` function now utilizes the newly added function to avoid code duplication, marking a single notification as read/unread needs to be validated.

* The proper way to test it will be to interact with the Checkmark button that is added with [this PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/17772#issuecomment-1013704458). **This will only be possible with the follow up PR's** that make use of this WPKit change. Though, unit tests are already in place.

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,13 +1,13 @@
 PODS:
   - Alamofire (4.8.2)
-  - CocoaLumberjack (3.7.4):
-    - CocoaLumberjack/Core (= 3.7.4)
-  - CocoaLumberjack/Core (3.7.4)
+  - CocoaLumberjack (3.7.2):
+    - CocoaLumberjack/Core (= 3.7.2)
+  - CocoaLumberjack/Core (3.7.2)
   - FormatterKit/Resources (1.9.0)
   - FormatterKit/TimeIntervalFormatter (1.9.0):
     - FormatterKit/Resources
   - NSObject-SafeExpectations (0.0.4)
-  - OCMock (3.9.1)
+  - OCMock (3.8.1)
   - OHHTTPStubs (9.1.0):
     - OHHTTPStubs/Default (= 9.1.0)
   - OHHTTPStubs/Core (9.1.0)
@@ -23,8 +23,8 @@ PODS:
   - OHHTTPStubs/OHPathHelpers (9.1.0)
   - OHHTTPStubs/Swift (9.1.0):
     - OHHTTPStubs/Default
-  - UIDeviceIdentifier (1.7.1)
-  - WordPressShared (1.17.0):
+  - UIDeviceIdentifier (1.4.0)
+  - WordPressShared (1.16.1):
     - CocoaLumberjack (~> 3.4)
     - FormatterKit/TimeIntervalFormatter (~> 1.8)
   - wpxmlrpc (0.9.0)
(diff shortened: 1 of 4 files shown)


Podfile.lock
```diff
@@ -54,15 +54,15 @@ SPEC REPOS:
...
-  UIDeviceIdentifier: 44f805037d21b94394821828f4fcaba34b38c2d0
-  WordPressShared: 5477f179c7fe03b5d574f91adda66f67d131827e
+  UIDeviceIdentifier: d0fee204f467dacf8808b7ac14ce43affeb271a5
+  WordPressShared: a4b0308a6345d4dda20c8f7ad9317df4246b4a00
   wpxmlrpc: bf55a43a7e710bd2a4fb8c02dfe83b1246f14f13
 
 PODFILE CHECKSUM: 41e0ac544d2de6e9267760f9f501e765675b1fa4
 
-COCOAPODS: 1.10.2
+COCOAPODS: 1.11.2
```
ScoutHarris: I assume you are using a more recent version of Cocoapods that caused this change. This is actually maintained by our platform team, so we don't want to do this now. The same goes for 3rd party pods, i.e. the four that were updated here. As such, we shouldn't include any of these changes in this PR.

WordPressKit/NotificationSyncServiceRemote.swift
```diff
@@ -54,15 +58,35 @@ public class NotificationSyncServiceRemote: ServiceRemoteWordPressComREST {
...
+    }
+
+    /// Updates an array of Notifications' Read Status as specified.
+    ///
+    /// - Parameters:
+    ///     - notificationIDs: ID's of Notifications to Mark as Read.
+    ///     - read: The new Read Status to set.
+    ///     - completion: Closure to be executed on completion, indicating whether the OP was successful or not.
+    ///
+    @objc public func updateReadStatusOfNotifications(_ notificationIDs: [String], read: Bool, completion: @escaping ((Error?) -> Void)) {
```
ScoutHarris: Just a nitpick - I think the method name `updateReadStatusForNotifications` would fit better with our naming conventions.

WordPressKit/NotificationSyncServiceRemote.swift
```diff
@@ -13,6 +13,10 @@ public class NotificationSyncServiceRemote: ServiceRemoteWordPressComREST {
         case failed
     }
 
+    public enum InputError: Int, Error {
```
ScoutHarris, line 16: I wonder if we should just use an `Error` type here? I don't see the purpose of the `Int`. Granted I've not added many of these myself, but a search reveals we use `Error` much more often than `Int, Error`.

ScoutHarris (changes requested):
I confirmed the existing mark notification read/unread still works. 🎉  I did leave some feedback, but the Request Changes is mainly due to the `Podfile.lock` changes.

WordPressKit/NotificationSyncServiceRemote.swift
alpavanoglu: Because the completion closure accepts a "generic" `Error` type, I needed the `InputError` to be some sort of `RawRepresentable` or `Equatable` to make testable. Indeed from the classes perspective this does not add any value. So it is added for the unit tests. I think we can compare the description in the tests but that didn't seem very clean. Can you think of a neat way of testing this while keeping the type as `InputError: Error`?

Podfile.lock
alpavanoglu: Got it. I will revert these changes right away 👍🏽

WordPressKit/NotificationSyncServiceRemote.swift
ScoutHarris: Ah, I see. That's fine then. It wasn't a big deal anyway, we can leave it as-is. Thanks for the explanation!

alpavanoglu: Cool! I can also build an overarching Error type over `InputError` & `SyncError` and change the closure argument types from Error to that. But might be an overkill at this case

ScoutHarris: > But might be an overkill at this case

Agreed. 😄

ScoutHarris (approved):
Thanks for the changes @alpavanoglu . All looks good!

:shipit:

PR #476, opened 2022-01-21
Notifications: add support for marking multiple notifications as read/unread

ScoutHarris:
### Description

Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/17135
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/17801

This PR adds support for marking multiple notifications as read or unread. Instead of providing a single NotificationID,
we can now also provide an array of IDs.

This is to merge Alp's trial project PR #474 to `trunk`.

### Testing Details

### DIFF
--- a/WordPressKit/NotificationSyncServiceRemote.swift
+++ b/WordPressKit/NotificationSyncServiceRemote.swift
@@ -58,7 +58,7 @@ public class NotificationSyncServiceRemote: ServiceRemoteWordPressComREST {
     ///     - completion: Closure to be executed on completion, indicating whether the OP was successful or not.
     ///
     @objc public func updateReadStatus(_ notificationID: String, read: Bool, completion: @escaping ((Error?) -> Void)) {
-        updateReadStatusOfNotifications([notificationID], read: read, completion: completion)
+        updateReadStatusForNotifications([notificationID], read: read, completion: completion)
     }
 
     /// Updates an array of Notifications' Read Status as specified.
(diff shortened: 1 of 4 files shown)


WordPressKit/NotificationSyncServiceRemote.swift
```diff
@@ -13,6 +13,10 @@ public class NotificationSyncServiceRemote: ServiceRemoteWordPressComREST {
         case failed
     }
 
+    public enum InputError: Int, Error {
+        case arrayEmpty
```
frosty: Nitpick really, but `arrayEmpty` sounds a bit tied to the implementation details to me. I'd prefer something like `noNotificationIDsProvided` or `nothingToUpdate` or something. (although this also has me thinking if this is really an error?)

frosty (approved):
Works well, just one small nitpick.

WordPressKit/NotificationSyncServiceRemote.swift
alpavanoglu: Fair point. My reasoning was that passing an empty array is somewhat an api misuse therefore I wanted to highlight the issue for the caller. Would an assertion fit here better? Or an improved name would do the trick? What would you suggest?

frosty: I totally understand 🙂  I guess I have often just seen an assertion for something like this, if it should always be called with an array of notifications. I'd be happy with that or a slightly different name!

PR #477, opened 2022-01-22
Add `getEmailFollowers` and `deleteEmailFollower` API calls

wargcm:
## Description

See wordpress-mobile/WordPress-iOS#6969

Adds the following API calls:
- [`/sites/$site/stats/followers`](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/stats/followers/)
- `/sites/$site/email-followers/$user_id/delete` (No API documentation)

These are done in the same manner as [`/sites/$site/follows`](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/follows/) and `/sites/$site/followers/$user_id/delete` to allow for better compatibility. It reuses `StatsFollower` from the insights screen and converts it to a `RemotePerson` to allow for easier use.

## Testing Details

Unit tests were added for every change. Additionally, I verified integration with [WordPress-iOS](https://github.com/wordpress-mobile/WordPress-iOS) after making my changes for the email UI filter.

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '4.46.0'
+  s.version       = '4.46.1'
 
   s.summary       = 'WordPressKit offers a clean and simple WordPress.com and WordPress.org API.'
   s.description   = <<-DESC
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '4.46.1'
+  s.version       = '4.46.0-beta.3'
 
   s.summary       = 'WordPressKit offers a clean and simple WordPress.com and WordPress.org API.'
   s.description   = <<-DESC
--- a/WordPressKit/PeopleServiceRemote.swift
+++ b/WordPressKit/PeopleServiceRemote.swift
@@ -97,6 +97,45 @@ public class PeopleServiceRemote: ServiceRemoteWordPressComREST {
         })
     }
 
+    /// Retrieves the collection of email followers associated to a site.
+    ///
+    /// - Parameters:
+    ///     - siteID: The target site's ID.
+    ///     - page: The page to fetch.
+    ///     - max: The max number of followers to fetch.
+    ///     - success: Closure to be executed on success.
+    ///     - failure: Closure to be executed on error.
+    ///
+    /// - Returns: An array of EmailFollower.
+    ///
+    public func getEmailFollowers(_ siteID: Int,
+                                  page: Int = 1,
+                                  max: Int = 20,
+                                  success: @escaping ((_ followers: [EmailFollower], _ hasMore: Bool) -> Void),
+                                  failure: @escaping (Error) -> Void) {
+        let endpoint = "sites/\(siteID)/stats/followers"
+        let path = self.path(forEndpoint: endpoint, withVersion: ._1_1)
+        let parameters: [String: AnyObject] = [
+            "page": page as AnyObject,
+            "max": max as AnyObject,
+            "type": "email" as AnyObject
+        ]
+
+        wordPressComRestApi.GET(path, parameters: parameters, success: { (responseObject, _) in
+            guard let response = responseObject as? [String: AnyObject],
+                  let subscribers = response["subscribers"] as? [[String: AnyObject]],
+                  let totalPages = response["pages"] as? Int else {
+                      failure(ResponseError.decodingFailure)
+                      return
+                  }
+            let followers = subscribers.compactMap { EmailFollower(siteID: siteID, statsFollower: StatsFollower(jsonDictionary: $0)) }
+            let hasMore = totalPages > page
+            success(followers, hasMore)
+        }, failure: { (error, _) in
+            failure(error)
+        })
+    }
+
     /// Retrieves the collection of Viewers associated to a site.
     ///
     /// - Parameters:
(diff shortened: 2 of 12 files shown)


WordPressKit/PeopleServiceRemote.swift
```diff
@@ -97,6 +97,45 @@ public class PeopleServiceRemote: ServiceRemoteWordPressComREST {
         })
     }
 
+    /// Retrieves the collection of email followers associated to a site.
+    ///
+    /// - Parameters:
+    ///     - siteID: The target site's ID.
+    ///     - page: The page to fetch.
+    ///     - max: The max number of followers to fetch.
+    ///     - success: Closure to be executed on success.
```
leandroalonso: What do you think about changing the success description to:

```suggestion
    ///     - success: Closure to be executed on success with an array of EmailFollower.
```

WordPressKit/PeopleServiceRemote.swift
```diff
@@ -97,6 +97,45 @@ public class PeopleServiceRemote: ServiceRemoteWordPressComREST {
...
+    ///     - failure: Closure to be executed on error.
+    ///
+    /// - Returns: An array of EmailFollower.
```
leandroalonso: I think `returns` stands for the actual return of the function, which in this case is `Void` and we can omit it. Wdyt?

WordPressKit/PeopleServiceRemote.swift
```diff
@@ -225,6 +264,28 @@ public class PeopleServiceRemote: ServiceRemoteWordPressComREST {
...
+    ///     - failure: Optional closure to be executed on error.
+    ///
+    @objc public func deleteEmailFollower(_ siteID: Int,
+                                          userID: Int,
+                                          success: (() -> Void)? = nil,
+                                          failure: ((Error) -> Void)? = nil) {
+        let endpoint = "sites/\(siteID)/email-followers/\(userID)/delete"
+        let path = self.path(forEndpoint: endpoint, withVersion: ._1_1)
+
+        wordPressComRestApi.POST(path, parameters: nil, success: { (_, _) in
```
leandroalonso: Small nitpick:

```suggestion
        wordPressComRestApi.POST(path, parameters: nil, success: { _, _ in
```

WordPressKit/PeopleServiceRemote.swift
```diff
@@ -225,6 +264,28 @@ public class PeopleServiceRemote: ServiceRemoteWordPressComREST {
...
+            success?()
+        }, failure: { (error, _) in
```
leandroalonso:
```suggestion
        }, failure: { error, _ in
```

WordPressKitTests/Models/Stats/V2/Insights/StatsDotComFollowersInsightTests.swift
```diff
@@ -0,0 +1,129 @@
...
+        let jsonDictionary = getFollowerDictionary(id: id, name: name, subscribedDate: date, avatarUrl: avatarUrl)
+
+        // When
+        let follower = StatsFollower(jsonDictionary: jsonDictionary)
+
+        // Then
+        XCTAssertEqual(follower?.id, id)
+        XCTAssertEqual(follower?.name, name)
+        XCTAssertEqual(follower?.subscribedDate, date)
+        XCTAssertEqual(follower?.avatarURL, expectedUrl)
```
leandroalonso: We could probably remove that? This is already tested in the first test and it would be easier to see what this test is about (only checking the `id`)

WordPressKitTests/Models/Stats/V2/Insights/StatsDotComFollowersInsightTests.swift
```diff
@@ -0,0 +1,129 @@
+import XCTest
+@testable import WordPressKit
+
+class StatsDotComFollowersInsightTests: XCTestCase {
+
+    private enum Constants {
```
leandroalonso: In tests, my main priority is always legibility, so while I don't see this extraction as a problem I think we can keep this string where it's needed?

leandroalonso (commented):
@wargcm this is looking good! Left some small suggestions.

Another thing you should do in this PR is to bump the `.podspec` version. An example: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/475/files#diff-4a47fbf5cfb74b949ceff23c93ef33c8b3b92d3a37eb0e846f5d0251d64a82edR5

WordPressKit/PeopleServiceRemote.swift
wargcm: Good call, that would make it more accurate. I updated it include both the `EmailFollower` array and the bool passed to the closure.

WordPressKit/PeopleServiceRemote.swift
wargcm: Makes sense, I removed the `returns`.

WordPressKitTests/Models/Stats/V2/Insights/StatsDotComFollowersInsightTests.swift
wargcm: I updated it to be more focused on just the `id` property.

WordPressKitTests/Models/Stats/V2/Insights/StatsDotComFollowersInsightTests.swift
wargcm: Fair point. I was thinking from the angle of "I don't want to scroll all the way to the bottom to change this value 😆 ". I agree that legibility should take priority over that though.

WordPressKit.podspec
```diff
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '4.46.0'
+  s.version       = '4.46.1'
```
wargcm: @leandroalonso Is this the correct version you wanted me to bump it to? Or did you want me to make it something like `4.47.0-beta.1`?

leandroalonso: Let's take a look at the list of latest releases: https://github.com/wordpress-mobile/WordPressKit-iOS/releases

It should be `4.46.0-beta.3`

leandroalonso (approved):

WordPressKit.podspec
mokagio: @wargcm @leandroalonso FYI: I forgot to publish the [`4.46.0` stable release](https://github.com/wordpress-mobile/WordPressKit-iOS/releases/tag/4.46.0) for the `4.46.0` tag on Monday, so `4.46.0-beta.3`, while the correct guess based on what the Releases page of this repo showed at the time Leandro left the previous comment, is not the right value.

In the meantime, @ScoutHarris kindly shipped [`4.47.0-beta.1`](https://github.com/wordpress-mobile/WordPressKit-iOS/releases/tag/4.47.0-beta.1). If you need to update your `Podfile` to use the changes from this PR, please use `4.47.0-beta.1`.

Apologies for the confusion 😞

wargcm: No worries, thanks for letting me know!

PR #479, opened 2022-01-24
Update UIDeviceIdentifier dependency to version 2.0.0

mokagio:
The major version breaking change is due to the library explicitly supporting iOS 9.0 and above, thus dropping support for all prior versions. That doesn't affect us, of course, because we are targeting way above 9.0 already.

See https://github.com/squarefrog/UIDeviceIdentifier/issues/47#issuecomment-998592243

I think we ought to upgrade to 2.0.0 to avoid `pod update` not fetching newest versions because it would stop to the latest 1.x release of the library.

### Testing Details

There are no functionality changes between 2.0.0 and the previous version, 1.4.0, only new models added to the list: https://github.com/squarefrog/UIDeviceIdentifier/compare/1.4.0...2.0.0 If CI is green, we can merge this.


AliSoftware (approved):
Major bump makes sense given the circumstances and breaking changes not affecting us indeed. :shipit:

PR #494, opened 2022-04-14
Add new fields to site creation request

Gio2018:
### This PR adds two new fields to the site creation request, to handle site creation without an existing domain suggestions: `findAvailableUrl` and `siteCreationFlow`.

Fixes #NA

### Testing Details

Can be tested on the related WordPRess-iOS [PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/18378)

### DIFF
--- a/WordPressKit/WordPressComServiceRemote+SiteCreation.swift
+++ b/WordPressKit/WordPressComServiceRemote+SiteCreation.swift
@@ -82,15 +82,15 @@ public struct SiteCreationRequest: Encodable {
     }
 
     private enum CodingKeys: String, CodingKey {
-        case clientIdentifier   = "client_id"
-        case clientSecret       = "client_secret"
+        case clientIdentifier = "client_id"
+        case clientSecret = "client_secret"
         case languageIdentifier = "lang_id"
-        case isPublic           = "public"
-        case shouldValidate     = "validate"
-        case siteURLString      = "blog_name"
-        case title              = "blog_title"
-        case options            = "options"
-        case findAvailableUrl  = "find_available_url"
+        case isPublic = "public"
+        case shouldValidate = "validate"
+        case siteURLString = "blog_name"
+        case title = "blog_title"
+        case options = "options"
+        case findAvailableURL = "find_available_url"
     }
 }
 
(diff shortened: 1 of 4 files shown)


WordPressKit/WordPressComServiceRemote+SiteCreation.swift
```diff
@@ -87,13 +100,15 @@ private struct SiteCreationOptions: Encodable {
...
     let timezoneIdentifier: String?
+    let siteCreationFlow: String?
 
     enum CodingKeys: String, CodingKey {
         case segmentIdentifier  = "site_segment"
         case verticalIdentifier = "site_vertical"
         case siteInformation    = "site_information"
         case siteDesign         = "template"
         case timezoneIdentifier = "timezone_string"
+        case siteCreationFlow  = "site_creation_flow"
```
twstokes: Nit, alignment:

```suggestion
        case siteCreationFlow   = "site_creation_flow"
```

WordPressKit/WordPressComServiceRemote+SiteCreation.swift
```diff
@@ -78,6 +90,7 @@ public struct SiteCreationRequest: Encodable {
         case siteURLString      = "blog_name"
         case title              = "blog_title"
         case options            = "options"
+        case findAvailableUrl  = "find_available_url"
```
twstokes: Nittiest of nits, totally optional: `siteURLString` has `URL` and this one has `Url`.

WordPressKitTests/SiteCreationRequestEncodingTests.swift
```diff
@@ -14,6 +14,8 @@ final class SiteCreationRequestEncodingTests: XCTestCase {
     static let expectedClientId = "TEST-ID"
     static let expectedClientSecret = "TEST-SECRET"
     static let expectedTimezoneIdentifier = "Pacific/Samoa"
+    static let siteCreationFlow: String? = nil
```
twstokes, line 17: Non-blocker, but it may provide a _little_ bit of value to populate this string and then check for it in `validate` (as well as `false` for `findAvailableUrl`). Up to you, though.

twstokes (approved):
LGTM - only nits! 🎉

WordPressKit/WordPressComServiceRemote+SiteCreation.swift
Gio2018: This looked different on my Xcode... (even worse than this 🤣 ). Maybe we should just ditch the `=` aligned on the same column

PR #497, opened 2022-04-22
Update CartResponse CartKey return type to fix parse failure

frosty:
### Description

While testing the Domains purchase flow in the Jetpack app, I noticed that when trying to create a cart the response wasn't being parsed correctly. The model was expecting the `CartKey` property to have a `String` value but it needs to be parsed as an `Int`. This PR updates the cart response to handle this correctly.

### Testing Details

Please test with the steps in the associated WordPress for iOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/18426


guarani (approved):
Hi @frosty, I see this is a draft but I'm leaving an approving review because it appears that nothing else is needed here code-wise and [I've tested this via the Jetpack app](https://github.com/wordpress-mobile/WordPress-iOS/pull/18426#pullrequestreview-970047018).

PR #502, opened 2022-05-09
Blogging Prompts: fix prompt date

ScoutHarris:
### Description

Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/18375

The date string for each fetched prompt is now converted to a Date.

### Testing Details

Can be tested with WPiOS PR https://github.com/wordpress-mobile/WordPress-iOS/pull/18545.

I added some temporary logging to confirm the prompt dates. I will remove it before merging (hence the `DO NOT MERGE` label).

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '4.52.0-beta.1'
+  s.version       = '4.52.0-beta.2'
 
   s.summary       = 'WordPressKit offers a clean and simple WordPress.com and WordPress.org API.'
   s.description   = <<-DESC
--- a/WordPressKit/RemoteBloggingPrompt.swift
+++ b/WordPressKit/RemoteBloggingPrompt.swift
@@ -29,6 +29,7 @@ extension RemoteBloggingPrompt: Decodable {
     private static var dateFormatter: DateFormatter = {
         let formatter = DateFormatter()
         formatter.locale = .init(identifier: "en_US_POSIX")
+        formatter.timeZone = .init(secondsFromGMT: 0)
         formatter.dateFormat = "yyyy-MM-dd"
         return formatter
     }()
--- a/WordPressKit/RemoteBloggingPrompt.swift
+++ b/WordPressKit/RemoteBloggingPrompt.swift
@@ -27,7 +27,10 @@ extension RemoteBloggingPrompt: Decodable {
 
     /// Used to format the fetched object's date string to a date.
     private static var dateFormatter: DateFormatter = {
-        JSONDecoder.DateDecodingStrategy.DateFormat.noTime.formatter
+        let formatter = DateFormatter()
+        formatter.locale = .init(identifier: "en_US_POSIX")
+        formatter.dateFormat = "yyyy-MM-dd"
+        return formatter
     }()
 
     public init(from decoder: Decoder) throws {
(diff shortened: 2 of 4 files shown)


dvdchr:
Hey @ScoutHarris 👋🏼 . Thank you for noticing the bug and opening a PR to fix this! Changes look good to me, but apart from one issue. I've pushed two commits: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/502/commits/008447f66061c6b091cb7f8486122b734f01df07 to address a timezone issue, and https://github.com/wordpress-mobile/WordPressKit-iOS/pull/502/commits/e5cb3769be64b22720e2b698108aa2f8fce61101 to fortify the unit tests. Feel free to revert or let me know if you have concerns with the changes! 🙇🏼 

In https://github.com/wordpress-mobile/WordPressKit-iOS/pull/502/commits/008447f66061c6b091cb7f8486122b734f01df07, I've set the timezone information of the `DateFormatter` to GMT. This prevents Swift from assuming that the date string is in the machine's local timezone. Here's the output when running from my local machine (before the timezone is set):
```
🔴 endpoint object date:  2022-05-03
🔴 RemoteBloggingPrompt date:  2022-05-02 17:00:00 +0000
🔴 endpoint object date:  2021-09-12
🔴 RemoteBloggingPrompt date:  2021-09-11 17:00:00 +0000
```

Here's how it looks after https://github.com/wordpress-mobile/WordPressKit-iOS/pull/502/commits/008447f66061c6b091cb7f8486122b734f01df07:
```
🔴 endpoint object date:  2022-05-03
🔴 RemoteBloggingPrompt date:  2022-05-03 00:00:00 +0000
🔴 endpoint object date:  2021-09-12
🔴 RemoteBloggingPrompt date:  2021-09-12 00:00:00 +0000
```

---

Additionally, this should have been caught via unit tests. There are two causes that led to this sneaky little bug escaping our perimeter:

1. In the `JSONDecoderExtension` class, the `noTime` enum case uses the date format string `yyyy-mm-dd`. The lowercase `mm` here translates to the minutes of the hour, instead of the month. I'm pretty sure this is an unintended mistake, and the format should've been `yyyy-MM-dd`.
2. In the tests, I've used `2022-01-01` and `2022-01-02` as the dates for the mock data. This led to a false positive since the phony parsing happened to match the month in the mock data. In https://github.com/wordpress-mobile/WordPressKit-iOS/pull/502/commits/e5cb3769be64b22720e2b698108aa2f8fce61101, I've updated the dates in the mock data so all the date components are different — and I'm taking note to avoid using the value `1` 🙃 . 

I was thinking to suggest that the date format needs to be updated to `yyyy-MM-dd` to prevent more victims in the future, but then I realized that it's been there since ~2018. I'll probably need to make sure that changing this doesn't cause any unwanted effect, and I'll create a separate PR for this later!

Just to share my findings, there are several classes that depend on `JSONDecoder.apiDecoder`: `EditorSettings`, `PlanServiceRemote_ApiVersion1_3`, `JetpackBackupServiceRemote`, `DomainsServiceRemote`, `JetpackScanServiceRemote`, etc. However, we only need to care about models with date fields populated with `yyyy-MM-dd` date strings. The list narrows down to `RemotePlan_ApiVersion1_3` model where the `autoRenewDate` field is populated with yyyy-MM-dd format, based on [the mock data](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/trunk/WordPressKitTests/Mock%20Data/site-plans-v3-success.json#L33).

dvdchr (approved):
LGTM!

ScoutHarris:
Hey @dvdchr . Thanks for fixing the fix. 😆 You're awesome!

PR #517, opened 2022-06-06
SwiftLint setup – Part 1 – Local only SwiftLint, with local configuration

mokagio:
### Description

This is the most basic SwiftLint setup we could have:

1. Installed locally, with version managed via CocoaPods. _Note that there are alternatives, but the one with the best straightforwardness to level of control I could find in the context of HACK Week._
2. Use a config file tracked locally.
3. Run only as an Xcode Build Phase Run Script

### Testing Details

Checkout this branch, and update the code to trigger a SwiftLint violation. The simplest thing to do is add a `;` at the end of a line in the production and test code. Build the app and verify that Xcode shows a warning.

### Next Steps

- Use a remote configuration, so that we can share it across various project and maintain a consistent style
- Experiment: Use Danger instead of Hound for inline comments. This is slower but more configurable.

---

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,85 @@
+# Project configuration
+excluded:
+  - Pods
+  - Scripts
+  - vendor
+
+# Rules
+whitelist_rules:
+  # Colons should be next to the identifier when specifying a type.
+  - colon
+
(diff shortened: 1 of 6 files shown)


mokagio:
Closed in favor of #693 and #735

PR #518, opened 2022-06-06
Run SwiftLint in CI

mokagio:
### Description

Sets up [Danger](https://danger.systems) and [`danger-swiftlint-ruby`](https://github.com/ashfurrow/danger-ruby-swiftlint) to run SwiftLint in Buildkite and leave inline comments with the failures.

Screen Shot 2022-06-08 at 1 11 36 pm

### Danger vs Hound

Danger via Buildkite is slower than Hound. I think we could Dockerize and do other things to make it faster, but I doubt it'll ever be as fast as Hound.

However, with Danger we have full control and we can adopt the latest version of the linters. With Hound, we depend on what the Hound team does on their end, and they haven't been very fast at adopting new SwiftLint versions historically.

### Testing Details

See #525 for the setup in action finding a failure.

### See Also

- [SwiftLint setup – Part 1 – Local only SwiftLint, with local configuration](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/517)
- [SwiftLint setup – Part 2 – Use remote configuration for SwiftLint](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/519)

---


mokagio:
See also #693 and #735

PR #519, opened 2022-06-06
SwiftLint setup – Part 2 – Use remote configuration for SwiftLint

mokagio:
### Description

This PR simply switches to using a remote configuration for SwiftLint.

The advantage is that we can now have a centralized source of truth for the style to use across our Swift codebases. SwiftLint also lets you override the remote configuration locally, so projects with special needs can implement them without duplicating what could be a large configuration file.

### Testing Details

Checkout this branch, and update the code to trigger a SwiftLint violation. The simplest thing to do is add a ; at the end of a line in the production and test code. Build the app and verify that Xcode shows a warning.

You can then enable a new rule or disable an existing one locally, and verify SwiftLint picks up the related violations.

### See Also

- [SwiftLint setup – Part 1 – Local only SwiftLint, with local configuration](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/517)

### Next Steps

- [Experiment: Use Danger instead of Hound for inline comments](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/518). This is slower but more configurable.

---


mokagio:
Closed in favor of #693 and #735

PR #530, opened 2022-07-20
Add Test Analytics

mokagio:
### Description

Adds Buildkite Test Analytics reporting. See also paaHJt-3wn-p2.

### Testing Details

Nothing to test really, you can see Test Analytics works by checking the latest build on this branch and following the test analytics link:

image

- [x] Please check here if your pull request includes additional test coverage. – N.A.


WordPressKitTests/RemoteTestCase.swift
```diff
@@ -129,8 +130,10 @@ extension RemoteTestCase {
     ///         https://github.com/AliSoftware/OHHTTPStubs/wiki/Usage-Examples#stack-multiple-stubs-and-remove-installed-stubs
     ///
     func stubAllNetworkRequestsWithNotConnectedError() {
-        stub(condition: { _ in
-            return true
+        stub(condition: { request in
+            // Stub all requests other than those to the Buildkite Test Analytics API.
+            // We need those to go through for Test Analytics reporting.
+            request.url != TestCollector.apiURL
```
crazytonyli: I see there are a couple places doing similar check, I'm wondering is there anyway to bypass this buildkite API URL globally?

mokagio: I looked into OHHTStubs but couldn't find any global setup hooks. Luckily we have @AliSoftware himself that we can ask this to 😄 .

WordPressKitTests/TestCollector+Constants.swift
```diff
@@ -0,0 +1,7 @@
+import BuildkiteTestCollector
+import Foundation
+
+extension TestCollector {
+
+    static let apiURL = URL(string: "https://analytics-api.buildkite.com/v1/uploads")!
+}
```
mokagio, line 7: I opened https://github.com/buildkite/test-collector-swift/issues/36 suggesting to make this part of the library.

WordPressKitTests/RemoteTestCase.swift
AliSoftware: Stubs you configure in `OHHTTPStubs` are stacked (in a queue/array property of the shared instance), and when a request comes those stubs' conditions are then tested in reverse order (last added stub checked first). See [this wiki page for details](https://github.com/AliSoftware/OHHTTPStubs/wiki/Usage-Examples#stack-multiple-stubs-and-remove-installed-stubs).

I didn't check which of the different places doing similar check are located in the codebase, but you could thus consider installing that fallback stub early in every test's `setUp`/`setUpWithError`, and that would work even for test cases that might need custom stubs to be added to the stub queue on top of it.

```swift
func setUpWithError() throws {
  try super.setUpWithError()
  // Default fallback for tests that didn't have any custom stub configured that caught the request higher in the chain
  // For all requests except Buildkite Test Analytics, prevent them from going to the real world by default
  stub(!isHost("analytics-api.buildkite.com")) { … }
}
func tearDownWithError() throws {
  HTTPStubs.removeAllStubs()
  try super.tearDownWithError()
}

func test1() throws {
  // This test doesn't need custom stubs.
  // Any URLRequest that might get out accidentally will be caught by the fallback stub we set up in `setUpWithError`
  // Except for Buildkite Test Analytics requests, which will return `false` to the `condition` of the `stub` from `setUpWithError` and will thus be let thru to the real world
  callMySUT()
}

func test2() throws {
  // This test requires a custom stub for a specific set of requests
  stub(isHost("wordpress-api-stub") && pathMatches("/post/[0-9]*/")) {
    fixture(…)
  }
  // Then we make the call that will end up do that request
  callMyOtherSUT()
  // 1. If a request to "https://wordpress-api-stub/post/123/" is sent, it will be caught by the stub defined above
  // 2. If a request to Buildkite Test Analytics is sent:
  //    - It will be first tested against the condition for the stub defined in this method, which will return `false`
  //    - It will then be tested against the next stub, aka the one we set up in `setUpWithError`, and that one will return `false` as well (because the `url` will be `TestCollector.apiURL`… so the `!=` condition will be `false`)
  //    - So ultimately that request will reach the real world
  // 3. If a request to any other URL is sent:
  //    - It will be first tested against the condition for the stub defined in this method, which will return `false`
  //    - It will then be tested against the next stub, aka the one we set up in `setUpWithError`, and that one will return `true`, because the `url` will NOT be `TestCollector.apiURL`.
  //    - So ultimately such a case will emit a `XCTFail` and return that fake `NotConnectedToInternet` error, as defined in that fallback stub in `setUpWithError`
}
```

> **Note**: [See here for a list of common matchers that are built-in with OHHTTPStubs/Swift](https://github.com/AliSoftware/OHHTTPStubs/blob/master/Sources/OHHTTPStubsSwift/OHHTTPStubsSwift.swift#L118-L532), like the `isHost` and `pathMatches` ones (+ `!` matcher operator) I used in my examples above

WordPressKitTests/TestCollector+Constants.swift
AliSoftware: Might make more sense to make stubs to just test for the `analytics-api.buildkite.com` host, rather than relying on a specific full URL that might change in the future (especially if they update it in `v2` or subsequent versions of their APIs)?

You could then test it using a convenience OHHTTPStubs syntax like `stub(!isHost(TestCollector.apiHost)) { … }` — see [example usage in my other comment above](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/530/files/bccc87031632b4e5bfe012c0ce99055fd38de0b6#r941308493).

WordPressKitTests/RemoteTestCase.swift
mokagio: I don't know whether we need a fallback stub like in the setup you describe.

What I wish we could do is saying "always allow requests to `isHost(TestCollector.apiHost)` to go through". But I can see how this is not part of the library's philosophy and also not feasible in vanilla XCTest, where there is no "before the test suite runs" hook.

And for good reason!

It might be extra work to setup exceptions on a per-test-case basis, but the alternative would be to open the global state flood gates in the tests, and that would be worse, leading to all sorts of bad test design and non-deterministic failures.

mokagio:
@crazytonyli I updated the syntax a bit based on @AliSoftware's feedback, but I don't know whether I can improve the setup with because of [lack of global hooks](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/530#discussion_r942088259).

How would you feel about merging this the way it is and followup if it turns out there's a way to do it after all?

crazytonyli (approved):

PR #550, opened 2022-11-27
Delete CocoaLumberjack

crazytonyli:
### Description

Similar to https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/325. I actually copied the main changes (the new logging functions) from that PR.

Most of the changes are replacing `DDLogXxx` with `WPKitLogXxx`, which is added in WPLogging.h

**TODO**:
- [x] Update WordPressShared dependency after [2.0.0 is released](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/326).

### Testing Details


WordPressKit/Private/WPKitLoggingPrivate.h
```diff
@@ -1,2 +1 @@
-@import CocoaLumberjack;
-extern DDLogLevel ddLogLevel;
+
```
mokagio: Do we need to keep this empty file?

WordPressKit/Private/WPKitLoggingPrivate.m
```diff
@@ -1,2 +1 @@
-#import "WPKitLoggingPrivate.h"
-DDLogLevel ddLogLevel = DDLogLevelWarning;
+
```
mokagio: Do we need to keep this empty file?

WordPressKit/WPKitLogging.m
```diff
@@ -1,10 +1,47 @@
...
+        if (logger == NULL) { \
+            NSLog(@"[WordPressKit] Warning: please call `WPKitSetLoggingDelegate` to set a error logger."); \
+            return; \
+        } \
+        if (![logger respondsToSelector:@selector(logFunc)]) { \
+            NSLog(@"[WordPressKit] Warning: %@ does not implement " #logFunc, logger); \
+            return; \
+        } \
+        [logger performSelector:@selector(logFunc) withObject:[[NSString alloc] initWithFormat:str arguments:args]]; \
+    })
```
mokagio, line 27: I didn't know you could go to newline in `#define`s like this. Neat!

WordPressKitTests/Utilities/LoggingTests.swift
```diff
@@ -0,0 +1,70 @@
+import XCTest
+
+@testable import WordPressKit
+
+private class CaptureLogs: NSObject, WordPressLoggingDelegate {
+    var verboseLogs = [String]()
+    var debugLogs = [String]()
+    var infoLogs = [String]()
+    var warningLogs = [String]()
+    var errorLogs = [String]()
```
mokagio: Testing nitpick. To force all tests using `CaptureLogs` to use it as a `WordPressLoggingDelegate` implementation, we could make the probes `var`s `private(set)` so that they can only be read in the tests but not directly modified. The `private(set)` access, in my opinion, also makes it clearer that their purpose is to be used to check the object's state only.

```suggestion
    private(set) var verboseLogs = [String]()
    private(set) var debugLogs = [String]()
    private(set) var infoLogs = [String]()
    private(set) var warningLogs = [String]()
    private(set)var errorLogs = [String]()
```

Podfile
```diff
@@ -9,8 +9,8 @@ platform :ios, '13.0'
 
 def wordpresskit_pods
   pod 'Alamofire', '~> 4.8.0'
-  pod 'CocoaLumberjack', '~> 3.4'
-  pod 'WordPressShared', '~> 1.15-beta' # will use release and beta versions up to 2.0
+  # pod 'WordPressShared', '~> 1.15-beta' # will use release and beta versions up to 2.0
+  pod 'WordPressShared', :git => 'https://github.com/wordpress-mobile/WordPress-iOS-Shared.git', :branch => 'trunk'
```
mokagio: ~~We'll need to update this to use `'~> 2.0'` or `'~> 2.0', '>- 2.0.0-beta.1'` now that [2.0.0-beta.1 has been shipped](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/326).~~

This is already tracked in the PR description 😅 🤦‍♂️

mokagio (commented):
CI fails because of inconsistencies between `Podifle` and `Podifle.lock` (maybe because it's pointing to `trunk` and CI might be pulling a different commit?). That's alright because we need to update the `Podfile` again, now that there's [a new WordPressShared beta available](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/326).

However, when running CocoaPods locally, I noticed a long warning long starting with `[!] [Xcodeproj] Generated duplicate UUIDs:`. It can be seen [in CI too](https://buildkite.com/automattic/wordpresskit-ios/builds/629#0184bb3f-f961-4bf7-9816-47c25b3b1850/319-343).

WordPressKit/Private/WPKitLoggingPrivate.h
mokagio: I think we can safely remove it. https://github.com/wordpress-mobile/WordPressKit-iOS/pull/551.

WordPressKit.podspec
```diff
@@ -23,12 +23,11 @@ Pod::Spec.new do |s|
...
   s.dependency 'Alamofire', '~> 4.8.0'
-  s.dependency 'CocoaLumberjack', '~> 3.4'
   s.dependency 'NSObject-SafeExpectations', '0.0.4'
   s.dependency 'wpxmlrpc', '~> 0.9'
   s.dependency 'UIDeviceIdentifier', '~> 2.0'
 
   # Use a loose restriction that allows both production and beta versions, up to the next major version.
   # If you want to update which of these is used, specify it in the host app.
-  s.dependency 'WordPressShared', '~> 1.15-beta' # Don't change this until we hit 2.0 in WPShared
+  s.dependency 'WordPressShared', '~> 2.0.0-beta'
```
mokagio, line 32: Have you considered `2.0-beta`? Or is a stricter requirement better while we roll this out?

mokagio (approved):

WordPressKit.podspec
crazytonyli: Good catch. This is an oversight, I should've used `2.0-beta`. 👍

crazytonyli: Addressed here: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/553/commits/308f019fce1247de85e0e122801765494f8b0733

PR #556, opened 2022-12-08
Re-implement a few reader models

crazytonyli:
### Description

Almost line-by-line translation of a few reader model types. One and only breaking change is the properties are changed from implicity unwrapped optional to optional.

### Testing Details

Should be okay if CI is green.


CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- Re-implement a few reader model types in Swift.
```
mokagio:
```suggestion
- Re-implement a few reader model types in Swift. [#556]
```

WordPressKit/RemoteReaderSiteInfo.swift
```diff
@@ -0,0 +1,154 @@
...
+private let SiteDictionaryVisibleKey = "visible"
+private let SiteDictionaryPostCountKey = "post_count"
+private let SiteDictionaryIconPathKey = "icon.img"
+private let SiteDictionaryDescriptionKey = "description"
+private let SiteDictionaryIDKey = "ID"
+private let SiteDictionaryNameKey = "name"
+private let SiteDictionaryURLKey = "URL"
+private let SiteDictionarySubscriptionsKey = "subscribers_count"
+private let SiteDictionarySubscriptionKey = "subscription"
+private let SiteDictionaryUnseenCountKey = "unseen_count"
```
mokagio, line 20: Have you considered grouping these in a type?

```swift
private enum SiteDictionaryKey {
  static let feedId = "feed_ID"
  static let feedURL = "feed_URL"
  // etc...
}

// Usage
response.number(forKey: SiteDictionaryKey.feedId)
```

One could go a bit deeper and use `enum` with `case`s then add syntax sugar methods on `NSDictionary` to take away the `.rawValue` call, so we could write something like:

```swift
response.number(forSiteDictionaryKey: .feedId)
response.bool(forSiteDictionaryKey: .following)
```

...but maybe that's too much overhead and we should save the time to get to a place where we can use modern Swift with `Codable` sooner. 🤷‍♂️

WordPressKit/RemoteReaderCard.swift
```diff
@@ -45,11 +45,11 @@ public struct RemoteReaderCard: Decodable {
             sites = sitesArray.compactMap {
                 // Since RemoteReaderSiteInfo is written in Obj-C and doesn't support decoding
                 // attempt to recast the response as a dictionary value to be passed to the init below
```
mokagio: This comment is now out-of-date.

```suggestion
```

mokagio (approved):
👌

WordPressKit/RemoteReaderSiteInfo.swift
crazytonyli: That would be a nicer Swift code for sure. I intentionally tried to keep the re-implementation "boring", or as close to the original Objective-C implementation as possible, for easier review and less risk of introducing bugs. I think, once I've translated the 40 something Objective-C types into Swift, then we can do a proper "swiftification" afterwards?

PR #557, opened 2022-12-08
Re-implement RemoteMenu, RemoteMenuItem, and RemoteMenuLocation

crazytonyli:
### Description

Part of re-implementing Objective-C in Swift.


WordPressKit/RemoteMenu.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+@objcMembers public class RemoteMenu: NSObject {
```
mokagio, line 3: TIL about `@objcMembers`, to make _all_ members of a Swift type available to Objective-C. Neat.

mokagio (approved):
Neat little PR 👍 

All those `?` make me worry, but at least the compiler will make consumers deal with them...

When do you think it would be a good time to address all that nullability? Or, better question, is that something we should be worrying about at all (in the context of limited resources)

PR #558, opened 2022-12-08
Re-implement RemoteBlog and RemoteBlogOptionsHelper in Swift

crazytonyli:
### Description

Similar to https://github.com/wordpress-mobile/WordPressKit-iOS/pull/556, this PR is part of re-implementing Objective-C in Swift.


WordPressKit/RemoteBlog.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+import NSObject_SafeExpectations
+
+/// This class encapsulates all of the *remote* Blog properties
```
mokagio, line 4: Do you know what a _remote_ blog is? And what's the difference with a "non-remote (?)" blog?

WordPressKit/AccountServiceRemoteREST.m
```diff
@@ -1,6 +1,4 @@
 #import "AccountServiceRemoteREST.h"
-#import "RemoteBlog.h"
-#import "RemoteBlogOptionsHelper.h"
```
mokagio, line 3: Chasing down these two imports, I noticed their footprint is minimal in this implementation. As far as I could see, only this method:

https://github.com/wordpress-mobile/WordPressKit-iOS/blob/8ecfcca4bd594b306f8a3de8e2f0f15dcfbaad96/WordPressKit/AccountServiceRemoteREST.m#L386-L393

Might it be useful to re-write it in Swift in the context of this PR?

WordPressKit/RemoteBlog.swift
```diff
@@ -0,0 +1,79 @@
...
+    public var capabilities: NSDictionary?
+
+    /// Blog's total disk quota space.
+    public var quotaSpaceAllowed: NSNumber?
+
+    /// Blog's total disk quota space used.
+    public var quotaSpaceUsed: NSNumber?
+
+    /// Parses details from a JSON dictionary, as returned by the WordPress.com REST API.
+    public init(JSONDictionary json: NSDictionary) {
```
mokagio, line 59: I'm sure you thought of this, too, but at some point it'd be good to make this and the other types currently using `JSONDictionary` `Decodable`.

mokagio (approved):
Looking good. I tried to read the code, although I don't trust my eyes to be able to reliably pick up translation issues... 🤔 

But:

- I trust your work
- We have unit tests
- If something breaks in this process it'll be a good chance to add more tests

WordPressKit/RemoteBlog.swift
crazytonyli: I think "remote" indicates it's an HTTP API model, which is defined by the WordPress API schema (hence the "remote" name?). A "non-remote" blog (or just `Blog`) means it's a custom model type defined by the app, which may have a different structure than the "remote" one.

WordPressKit/AccountServiceRemoteREST.m
crazytonyli: I intentionally started with the easy ones to translate. This `AccountServiceRemoteREST` falls into my harder basket, and will be dealt with after I have finished with the easy ones 🥲

WordPressKit/RemoteBlog.swift
crazytonyli: Indeed. Once I've finished the Swift translation, I'll have a look at the `NSDictionary` usage across all the models. I'll probably start with specializing the argument type first, to be `[String: Any]` for example. I'm not sure about `Decodable`, as it's a more drastic change to the JSON parsing code. But I'll for sure considering adopting `Decodable` when I get to it. 👍

WordPressKit/AccountServiceRemoteREST.m
mokagio: Sure.

Just to be clear, I was suggesting converting only that method, but I can see how adding a dedicated extension just for it would be wasteful given we'd remove it once the whole class is converted soon enough.

PR #559, opened 2022-12-09
Re-implement a few remote post related types

crazytonyli:
### Description

Similar to https://github.com/wordpress-mobile/WordPressKit-iOS/pull/556, this PR is part of re-implementing Objective-C files in Swift.


WordPressKit/RemotePost.swift
```diff
@@ -0,0 +1,81 @@
...
+
+@objcMembers public class RemotePost: NSObject {
+
+    public static let statusDraft = "draft"
+    public static let statusPending = "pending"
+    public static let statusPrivate = "private"
+    public static let statusPublish = "publish"
+    public static let statusScheduled = "future"
+    public static let statusTrash = "trash"
+    public static let statusDeleted = "deleted" // Returned by wpcom REST API when a post is permanently deleted.
```
mokagio: I was going to suggest to define a `Status` `enum`, but I guess that wouldn't work because of the `@objcMembers` requirement.

WordPressKit/RemotePost.swift
```diff
@@ -0,0 +1,81 @@
...
+        super.init()
+        self.siteID = siteID
+        self.status = status
+        self.title = title
+        self.content = content
+    }
+
+    public override var debugDescription: String {
+        "\(super.description) (\(allProperties))"
+    }
```
mokagio, line 80: TIL [`debugDescription`](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418703-debugdescription).

> The debugger’s `po` command uses this property to create a textual representation of the object suitable for display in the debugger. The default implemention [sic] returns the same value as description. Override either property to provide custom object descriptions.

WordPressKit/RemotePost.swift
mokagio: Although...  I guess we could define the `enum` in an extension and use it here in the assignment?

```swift
extension RemotePost {

  enum Status: String {
     case draft
     case pending
     ...

...

public static let statusDraft = Status.draft.rawValue
public static let statusPending = Status.pending.rawValue
...
```

mokagio: What do you think of making that comment a documentation one?

```suggestion
    /// Returned by the WordPress.com REST API when a post is permanently deleted.
    public static let statusDeleted = "deleted"
```

mokagio (approved):

crazytonyli:
I've changed the base branch to the dedicated Swift migration branch, see https://github.com/wordpress-mobile/WordPressKit-iOS/pull/562 for more context.

WordPressKit/RemotePost.swift
crazytonyli: Yes, definitely. There are lots of space to make the translated code "more Swift". I'm currently looking into the model's parsing code and in turn may change their property declarations, which kind of relate to your proposal here.

PR #561, opened 2022-12-14
Re-implement `RemoteReaderPost` in Swift - Part 1

crazytonyli:
### Description

[`RemoteReaderPost`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/1baed487b45bb5cbf7c041ebd21e9a0551272d14/WordPressKit/RemoteReaderPost.m) has lots of parsing functions, it would be very hard to review and easy to make mistakes if translating the ObjC code to Swift in one go. I will break down the translation into a few parts. The first PR here translates the simpliest parsing functions, which are all covered by unit tests.

### Testing Details

We should be good if the CI jobs are green.


WordPressKit/ReaderPostsEnvelope.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
```
mokagio, line 1: This import might be unnecessary.

crazytonyli: Yes, that's true for the current project setup.  But I think it's needed for SPM support.

crazytonyli:
I've changed the base branch to the dedicated Swift migration branch, see https://github.com/wordpress-mobile/WordPressKit-iOS/pull/562 for more context.

PR #567, opened 2022-12-22
Add twoStepEnabled property to AccountSettings

salimbraksa:
Part of https://github.com/wordpress-mobile/WordPress-iOS/issues/19567

### Related PR
- https://github.com/wordpress-mobile/WordPress-iOS/pull/19807

### Description

This PR stores the `twoStepEnabled` boolean received from the Backend in `AccountSettings` model. The Jetpack app uses that property to decide whether to show `Scan QR Code` row in `Me` screen or not.

### Testing Details

See [Related PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/19807) Test Instructions.

### DIFF
--- a/WordPressKit/AccountSettings.swift
+++ b/WordPressKit/AccountSettings.swift
@@ -18,6 +18,7 @@ public struct AccountSettings {
     public let language: String    // language
     public let tracksOptOut: Bool
     public let blockEmailNotifications: Bool
+    public let twoStepEnabled: Bool // two_step_enabled
 
     public init(firstName: String,
                 lastName: String,
@@ -32,7 +33,8 @@ public struct AccountSettings {
                 webAddress: String,
                 language: String,
                 tracksOptOut: Bool,
-                blockEmailNotifications: Bool) {
+                blockEmailNotifications: Bool,
+                twoStepEnabled: Bool) {
         self.firstName = firstName
         self.lastName = lastName
         self.displayName = displayName
@@ -47,6 +49,7 @@ public struct AccountSettings {
         self.language = language
         self.tracksOptOut = tracksOptOut
         self.blockEmailNotifications = blockEmailNotifications
+        self.twoStepEnabled = twoStepEnabled
     }
 }
 
--- a/WordPressKit/AccountSettingsRemote.swift
+++ b/WordPressKit/AccountSettingsRemote.swift
@@ -157,8 +157,7 @@ public class AccountSettingsRemote: ServiceRemoteWordPressComREST {
     }
 
     private func settingsFromResponse(_ responseObject: AnyObject) throws -> AccountSettings {
-        guard let
-            response = responseObject as? [String: AnyObject],
+        guard let response = responseObject as? [String: AnyObject],
             let firstName = response["first_name"] as? String,
             let lastName = response["last_name"] as? String,
             let displayName = response["display_name"] as? String,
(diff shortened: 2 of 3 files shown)


WordPressKit/AccountSettingsRemote.swift
```diff
@@ -172,10 +171,12 @@ public class AccountSettingsRemote: ServiceRemoteWordPressComREST {
...
-            let blockEmailNotifications = response["subscription_delivery_email_blocked"] as? Bool else {
-                WPKitLogError("Error decoding me/settings response: \(responseObject)")
-                throw ResponseError.decodingFailure
-            }
+            let blockEmailNotifications = response["subscription_delivery_email_blocked"] as? Bool
+        else {
+            WPKitLogError("Error decoding me/settings response: \(responseObject)")
+            throw ResponseError.decodingFailure
+        }
+        let twoStepEnabled = response["two_step_enabled"] as? Bool ?? false
```
twstokes: 👋 Hey @salimbraksa - I was curious why this property is outside of the `guard` - are there instances where `two_step_enabled` isn't sent?

twstokes (commented):
Testing passed in https://github.com/wordpress-mobile/WordPress-iOS/pull/19807, but I did have one question. Thanks!

WordPressKit/AccountSettingsRemote.swift
salimbraksa: @twstokes Honestly I don't know if `two_step_enabled` is optional ( on the Backend side ). If the property is inside the `guard-let` and the property turns out to be `nil`, the whole request will fail and this change will cause a regression bug impacting existing features.

So to be safe, I'm handling that mapping outside the `guard` and I believe this won't cause any regression issues whether `two_step_enabled` is optional or not.

But out of curiosity, how can we know whether `two_step_enabled` is optional or not on the Backend side?

twstokes: 👋 That makes sense. From what I can tell, it won't be optional:

- We specify version 1.1 of the public API [here](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/474ea9cda1ec1b35e28f0bd967c46ec145405548/WordPressKit/AccountSettingsRemote.swift#L33).
- The [endpoint documentation](https://developer.wordpress.com/docs/api/1.1/get/me/settings/) doesn't mention `two_step_enabled` being optional or any different than the other parameters.
- You can also check out the internal PHP class for the endpoint (I'll provide the link).

Perhaps we should make it uniform with the other parameters unless we're sure it may not always be sent for v1.1 on .com. What do you think about that?

salimbraksa: Agreed. I'll make the changes. Thanks for helping with this. 🙏

salimbraksa: Resolved in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/567/commits/bbaf2dca6825c47c6a512f5cb43640604f80d224

twstokes (approved):
LGTM - thanks! 🚀

WordPressKit/AccountSettings.swift
```diff
@@ -32,7 +33,8 @@ public struct AccountSettings {
                 webAddress: String,
                 language: String,
                 tracksOptOut: Bool,
-                blockEmailNotifications: Bool) {
+                blockEmailNotifications: Bool,
+                twoStepEnabled: Bool) {
```
crazytonyli, line 37: @salimbraksa I think this is a breaking change, since a new argument is added to the public initialiser? Maybe we need a major version bump for this change.

PR #581, opened 2023-02-28
Refactor the logic to fetch metadata of VideoPress videos

fluiddot:
### Description

This PR exposes the endpoint to fetch the VideoPress token which is required to play private videos (internal discussion - p9ugOq-3sW-p2). As well as replace the previous function to fetch VideoPress video metadata (i.e. `getVideoURLFromVideoPressID`) with a new version that retrieves all metadata parameters.

Related to https://github.com/wordpress-mobile/gutenberg-mobile/issues/5512.

### Testing Details

Follow testing instructions from https://github.com/wordpress-mobile/WordPress-iOS/pull/20237.

---

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '6.3.0-beta.1'
+  s.version       = '7.0.0-beta.1'
 
   s.summary       = 'WordPressKit offers a clean and simple WordPress.com and WordPress.org API.'
   s.description   = <<-DESC
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '6.1.0'
+  s.version       = '6.2.0-beta.1'
 
   s.summary       = 'WordPressKit offers a clean and simple WordPress.com and WordPress.org API.'
   s.description   = <<-DESC
--- a/WordPressKit/RemoteVideoPressVideo.swift
+++ b/WordPressKit/RemoteVideoPressVideo.swift
@@ -40,7 +40,7 @@ import Foundation
     public var token: String?
 
     enum CodingKeys: String, CodingKey {
-        case id, title, description, width, height, duration, displayEmbed, allowDownload, rating, privacySetting, posterURL, originalURL, watermarkURL, bgColor, blogId, postId, finished, token, playURL
+        case id, title, description, width, height, duration, displayEmbed, allowDownload, rating, privacySetting, posterURL, originalURL, watermarkURL, bgColor, blogId, postId, finished, token
     }
 
     public init(dictionary metadataDict: NSDictionary, id: String) {
@@ -72,21 +72,24 @@ import Foundation
         finished = metadataDict.object(forKey: "finished") as? Bool
     }
 
-    /// Returns the URL that should be used to play the video.
+    /// Returns the specified URL adding the token as a query parameter, which is required to play private videos.
+    /// - Parameters:
+    ///   - url
     ///
-    /// The URL used is the original but adding the token as a query parameter, which is required to play private videos.
-    public func getPlayURL() -> URL? {
-        guard var videoPlayURL = self.originalURL else {
+    /// - Returns: The specified URL with the token as a query parameter.
+    @objc(getURLWithToken:)
+    public func getURLWithToken(url: URL?) -> URL? {
+        guard var urlWithToken = url else {
             return nil
         }
-        if let token = self.token, var urlComponents = URLComponents(url: videoPlayURL, resolvingAgainstBaseURL: true) {
+        if let token = self.token, var urlComponents = URLComponents(url: urlWithToken, resolvingAgainstBaseURL: true) {
             let metadataTokenParam = URLQueryItem(name: "metadata_token", value: token)
             var queryItems: [URLQueryItem] = urlComponents.queryItems ?? []
             queryItems.append(metadataTokenParam)
             urlComponents.queryItems = queryItems
-            videoPlayURL = urlComponents.url!
+            urlWithToken = urlComponents.url!
         }
-        return videoPlayURL
+        return urlWithToken
     }
 
     public func encode(to encoder: Encoder) throws {
@@ -110,8 +113,6 @@ import Foundation
         try container.encode(postId, forKey: .postId)
         try container.encode(finished, forKey: .finished)
         try container.encode(token, forKey: .token)
-        let playURL = getPlayURL()?.absoluteString ?? self.originalURL?.absoluteString ?? ""
-        try container.encode(playURL, forKey: .playURL)
     }
 
     public func asDictionary() -> [String: Any] {
--- a/WordPressKit/RemoteVideoPressVideo.swift
+++ b/WordPressKit/RemoteVideoPressVideo.swift
@@ -40,7 +40,7 @@ import Foundation
     public var token: String?
 
     enum CodingKeys: String, CodingKey {
-        case id, title, description, width, height, duration, displayEmbed, allowDownload, rating, privacySetting, posterURL, originalURL, watermarkURL, bgColor, blogId, postId, finished, token
+        case id, title, videoDescription = "description", width, height, duration, displayEmbed, allowDownload, rating, privacySetting, posterURL, originalURL, watermarkURL, bgColor, blogId, postId, finished, token
     }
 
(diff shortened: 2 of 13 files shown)


WordPressKit.podspec
```diff
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '6.2.0'
+  s.version       = '6.3.0-beta.1'
```
twstokes: 👋 Hey @fluiddot, should we bump the major here instead of the minor due to the breaking changes? ([SemVer reference](https://semver.org/spec/v2.0.0.html))

WordPressKit/RemoteVideoPressVideo.swift
```diff
@@ -0,0 +1,128 @@
...
+        finished = metadataDict.object(forKey: "finished") as? Bool
+    }
+
+    /// Returns the specified URL adding the token as a query parameter, which is required to play private videos.
+    /// - Parameters:
+    ///   - url: URL to include the token.
+    ///
+    /// - Returns: The specified URL with the token as a query parameter.
+    @objc(getURLWithToken:)
+    public func getURLWithToken(url: URL?) -> URL? {
```
twstokes: [According to the test](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/581/files#diff-a12b9fab5b497cd6f553adf161c95b8cb62337a3fb0ae518fe5791d81948c4f3R58), we're expecting this method to return the original URL if a token doesn't exist.

**Optional suggestion:** Due to the name of the function and its return type, I'd be tempted to have this return a non-nil URL only if it can make a URL with the token. I'd also consider dealing with the optional URL early ([style guide ref](https://github.com/wordpress-mobile/swift-style-guide#deal-with-optionals-early)) and put the responsibility of this check on the caller.

Tests and callers would need to be updated, but the function could then be simplified to something like:

``` swift
    public func getURLWithToken(url: URL) -> URL? {
        guard
            let token,
            var urlComponents = URLComponents(url: url, resolvingAgainstBaseURL: true)
        else {
            return nil
        }

        let metadataTokenParam = URLQueryItem(name: "metadata_token", value: token)
        urlComponents.queryItems = (urlComponents.queryItems ?? []) + [metadataTokenParam]
        return urlComponents.url
    }
```

WordPressKit/RemoteVideoPressVideo.swift
```diff
@@ -0,0 +1,128 @@
...
+        guard var urlWithToken = url else {
+            return nil
+        }
+        if let token = self.token, var urlComponents = URLComponents(url: urlWithToken, resolvingAgainstBaseURL: true) {
+            let metadataTokenParam = URLQueryItem(name: "metadata_token", value: token)
+            var queryItems: [URLQueryItem] = urlComponents.queryItems ?? []
+            queryItems.append(metadataTokenParam)
+            urlComponents.queryItems = queryItems
+            urlWithToken = urlComponents.url!
```
twstokes: See comment above on a proposal that could protect against the force unwrap. If we don't go with the proposal, we should consider returning `urlWithToken` if `urlComponents.url` is `nil` (even though we know this should be safe).

WordPressKit/RemoteVideoPressVideo.swift
```diff
@@ -0,0 +1,128 @@
...
+        }
+        return urlWithToken
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+
+        try container.encode(id, forKey: .id)
+        try container.encode(title, forKey: .title)
+        try container.encode(videoDescription, forKey: .description)
```
twstokes: If I understand correctly, we explicitly handle this encoding due to the `videoDescription` property name conflict. The good news is I _believe_ this can be avoided by adding `videoDescription = "description"` to our `CodingKeys`.

This should allow auto syntheses of the `Encodable` protocol and would remove the need for this function.

WordPressKit/RemoteVideoPressVideo.swift
```diff
@@ -0,0 +1,128 @@
...
+    public var watermarkURL: URL?
+    public var bgColor: String?
+    public var blogId: Int?
+    public var postId: Int?
+    public var finished: Bool?
+
+    public var token: String?
+
+    enum CodingKeys: String, CodingKey {
+        case id, title, description, width, height, duration, displayEmbed, allowDownload, rating, privacySetting, posterURL, originalURL, watermarkURL, bgColor, blogId, postId, finished, token
```
twstokes: We may be able to add `videoDescription = "description"` to allow auto synthesis of `Encodable`.

twstokes (commented):
This looks really good @fluiddot - thanks for taking it on!

I've left some comments in the PR and after we resolve those I can start testing. Thanks!

WordPressKit.podspec
fluiddot: Yep, good point. I'll bump the major version instead, thanks for catching this up 🙇 .

WordPressKit/RemoteVideoPressVideo.swift
fluiddot: The idea behind returning the URL when there's no token was to simplify the caller's code, but it's true that it can be misleading that the function might return the URL without the token if it's not defined. I'll update it with your suggestion, thanks!

WordPressKit/RemoteVideoPressVideo.swift
fluiddot: Ah, good to know. I'll try this option and update it accordingly if it works. Thanks for the tip!

WordPressKit.podspec
fluiddot: Suggestion applied in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/581/commits/7946aa5616737c2fac6be14d843419a2f01afba8.

WordPressKit/RemoteVideoPressVideo.swift
fluiddot: Suggestion applied in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/581/commits/d77b9479c0c1a9e13493229f03ad67aa12ba540a.

WordPressKit/RemoteVideoPressVideo.swift
fluiddot: I finally applied the above proposal so the force unwrap is no longer done.

WordPressKit/RemoteVideoPressVideo.swift
fluiddot: I tried this and worked as expected 🎊 , I applied this suggestion in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/581/commits/b901dc8f0ae251f2698a1bfa2e7c2a1c4599e6c1. Thanks!

WordPressKit/RemoteVideoPressVideo.swift
fluiddot: Yep, applied in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/581/commits/b901dc8f0ae251f2698a1bfa2e7c2a1c4599e6c1.

fluiddot:
Thanks @twstokes for the review, I really appreciate it 🙇. I applied the different suggestions so the PR is ready for another review. Let me know if you could take a look, thanks!

Due to this change https://github.com/wordpress-mobile/WordPressKit-iOS/pull/581/commits/d77b9479c0c1a9e13493229f03ad67aa12ba540a, I'll update the [parent PR](https://github.com/wordpress-mobile/gutenberg-mobile/issues/5512) accordingly to the change.

WordPressKit/RemoteVideoPressVideo.swift
fluiddot: Heads up that due to the major version bump, we also need to bump the `WordPressAuthenticator` library: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/754.

fluiddot:
> Due to this change [d77b947](https://github.com/wordpress-mobile/WordPressKit-iOS/commit/d77b9479c0c1a9e13493229f03ad67aa12ba540a), I'll update the [parent PR](https://github.com/wordpress-mobile/gutenberg-mobile/issues/5512) accordingly to the change.

Updated in https://github.com/wordpress-mobile/WordPress-iOS/pull/20237/commits/82d800cd5c89a259b136af8a7a25380982002a24.

WordPressKit/RemoteVideoPressVideo.swift
twstokes: Good catch @fluiddot!

twstokes (approved):
LGTM @fluiddot! Great work. 🚀

PR #582, opened 2023-02-28
Update feature flag platform name

guarani:
### Description

Addresses https://github.com/wordpress-mobile/WordPress-iOS/issues/20222

The feature flag name should be `ios` as mentioned in the above issue. On Android we're using the OS name `android`, so on iOS we should use `ios`..

### Testing Details

See https://github.com/wordpress-mobile/WordPress-iOS/pull/20238

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _None._
 
 ### Bug Fixes
 
-_None._
+- Changes the feature flag platform identifier to `ios`
 
 ### Internal Changes
 
@@ -58,7 +58,6 @@ _None._
 ### Bug Fixes
 
 - Fixes regression in logic to decode whether user has a free plan from JSON [#578]
-- Changes the feature flag platform identifier to `ios`
 
 ## 6.0.0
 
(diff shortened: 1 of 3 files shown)


staskus (approved):
Since no remote flag was relying on this platform flag so far, it's safe to make this change. I checked, this is the only place where the platform is passed. ✅

guarani:
Reverted `podspec` after @mokagio confirmed to me that we can simply merge to `trunk` and then point the client app to `trunk`.

<sub>internal ref paaHJt-4lp-p2#comment-7165</sub>

staskus:
@guarani 

I'm wondering if we should revert it back to `apple`. 

I'm making a change for a remote feature flag since it's still configured to use `apple` but it looks to require more changes than defining `platform_is(`ios`)` since it depends on the enum and additional checks.

fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Syvo%2Sjcpbz%2Qehyrf%2Sfep%2Srahzf%2Spynff%2Qcyngsbez.cuc%3Se%3Qs109ps1o-og
fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Syvo%2Sjcpbz%2Qehyrf%2Sfep%2Sehyrf%2Spynff%2Qcyngsbez%2Qvf.cuc%3Se%3Qs109ps1o%2326-og

----

### Update

Even if it requires more changes than anticipated, we should stick to renaming the flag since this change is shipping with 21.9.

Please check:
D104827-code

PR #588, opened 2023-03-22
Enable test repetition to work around recent CI flakiness.

mokagio:
I've been slowed down by CI flakiness in the past few releases and finally set aside some time to look into why that was the case.

Unfortunately, I haven't been able to pin down a cause yet and I am wary of focusing on this too much at the expense of other time sensitive projects.

So, for the time being, I have enabled test repetition, which has proven effective in dealing with flakiness.

To help future research, I also updated the test script to annotate the build with the tests results.

### Testing Details

See CI builds. All should be green.

---


.buildkite/build-and-test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -u
+
```
mokagio, line 2: Lifted the logic below from WordPress iOS and updated the paths and step names for how this repo is configured.

.buildkite/pipeline.yml
```diff
@@ -13,10 +13,11 @@ steps:
   #################
   - label: "🧪 Build and Test"
     key: "test"
-    command: |
-      build_and_test_pod
+    command: .buildkite/build-and-test.sh
     artifact_paths:
-      - "fastlane/test_output/*.xcresult.zip"
+      - fastlane/test_output/*.xcresult.zip
```
mokagio, line 18: (Not so) Fun fact: This file had never been uploaded because there had never been an instruction to create it from the `.xcresult`

crazytonyli (approved):
:shipit:

PR #592, opened 2023-04-11
Update blogging prompts endpoint to v3

dvdchr:
Refs https://github.com/wordpress-mobile/WordPress-iOS/issues/20496

> **Warning**
> This introduces a breaking change for the `RemoteBloggingPrompt` model due to some properties being removed. The WordPress-iOS codebase actually did not use any of the removed properties, so this is more like a clean up.
>
> Additionally, the remote endpoint for `fetchPrompts` is updated to the v3 version which returns a different set of data for the same parameters.

### Description

- This updates the Blogging Prompts service to fetch data from the v3 endpoint. 
- There are some differences between v2 and v3, and I've decided to modify the `RemoteBloggingPrompt` model directly instead of introducing a new model. 
- Lastly, the request parameters are also adjusted for v3, including the new optional `ignoresYear` option that uses the custom `--mm-dd` format.

Please note that this PR needs to be merged closely with its counterpart PR for WordPress-iOS to ensure that there are no breakages.

### Testing Details

- Added new tests & adjusted the existing ones for blogging prompts. Please run them locally and ensure that they are passing.
- Manual testing will be shared later once the counterpart PR from WordPress-iOS is up.

---


dvdchr:
Hi @mokagio 👋🏼, I would like to get your thoughts, please. This is a breaking change, so I considered bumping this to `9.0.0-beta.1`. But this would cause a conflict in WordPress-iOS since it depends on WordPressAuthenticator, which [declares a pessimistic requirement of WordPressKit](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/48637d73fe1c8f249ea7fdec1d71fdca9a366509/Podfile#L31) like so: 

```ruby
pod 'WordPressKit', '~> 8.7-beta'
```

Should I open a PR to bump the dependency on WordPressAuthenticator?

mokagio:
@dvdchr Thank you for being mindful of breaking changes 🙇‍♂️ 

This is a wild one, but from you vantage point how would it look to move the logic from this library to the app?

Realistically, this library is not meant for the general public. The only app clients are Jetpack/WordPress and WooCommerce.

WooCommerce does not, never did, and it's conceivable to assume never will need to hit the blogging prompts API.

If we folded the logic into the app, we'd bypass the need for the dependency update dance for this and every future blogging prompts iteration.

---

If that sounds too much, maybe I can help share the load.

If it still sounds too much, then updating WordPressAuthenticator in tandem would be the way to go.

dvdchr:
> ... how would it look to move the logic from this library to the app? ...
> If we folded the logic into the app, we'd bypass the need for the dependency update dance for this and every future blogging prompts iteration.

Yeah, that makes sense @mokagio. It'd be great if we could avoid the dependency wrangling, as you've said, since this feature is specific to the Jetpack app anyway. I'll port this logic to WordPress-iOS. 👍🏼

dvdchr:
Closing in favor of https://github.com/wordpress-mobile/WordPress-iOS/pull/21952

PR #598, opened 2023-05-04
Update editor settings models for block based themes

wargcm:
See: https://github.com/wordpress-mobile/WordPress-iOS/issues/20635

## Description

- Updates `isFSETheme` key on `RemoteBlockEditorSettings` to `__unstableIsBlockBasedTheme`
- Adds new key to `RemoteEditorThemeSupport` for block templates

These model changes are to support checking whether a blog has a theme that uses the full screen editor.

## Testing

Checkout the branch: https://github.com/wordpress-mobile/WordPress-iOS/compare/task/20635-update-fse-check

- Build and launch Jetpack
- Sign in to an account with multiple blogs
- Select a blog which has a theme using the full screen editor
- Set a breakpoint on these model changes
- Open the Gutenberg editor by creating a new post, it should kick off a network request fetching the editor settings
- **Verify** `isFSETheme` on the editor settings is `true`
- Modify [this function](https://github.com/wordpress-mobile/WordPress-iOS/blob/22c7f92981801a0254af4201ae3c7d2aa4d84938/WordPress/Classes/Services/BlockEditorSettingsService.swift#L47-L52)  to:
```swift
    func fetchSettings(_ completion: @escaping BlockEditorSettingsServiceCompletion) {
        // if blog.supports(.blockEditorSettings) {
        //     fetchBlockEditorSettings(completion)
        // } else {
            fetchTheme(completion)
        // }
    }
```
- Repeat the steps above except this time **verifying** `blockTemplates` is `true`
- Select a blog which does not have a theme using the full screen editor
- Repeat the above steps

---

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '8.0.0'
+  s.version       = '8.1.0-beta.1'
 
   s.summary       = 'WordPressKit offers a clean and simple WordPress.com and WordPress.org API.'
   s.description   = <<-DESC
--- a/WordPressKit/RemoteBlockEditorSettings.swift
+++ b/WordPressKit/RemoteBlockEditorSettings.swift
@@ -2,6 +2,7 @@ import Foundation
 
 public class RemoteBlockEditorSettings: Codable {
     enum CodingKeys: String, CodingKey {
+        case isFSETheme = "__unstableIsBlockBasedTheme"
         case galleryWithImageBlocks = "__unstableGalleryWithImageBlocks"
         case quoteBlockV2 = "__experimentalEnableQuoteBlockV2"
         case listBlockV2 = "__experimentalEnableListBlockV2"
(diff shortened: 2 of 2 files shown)


dvdchr (approved):
LGTM :shipit:

mokagio (commented):
@wargcm @dvdchr  thank you for the detailed manual testing instructions.

Have you considered adding unit tests for this decoding change?

wargcm:
@mokagio There were some added in WP iOS which inadvertently test these changes: https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/WordPress/WordPressTest/BlockEditorSettings_GutenbergEditorSettingsTests.swift

But I can add some to cover it in this repo if we think it's useful.

mokagio:
> But I can add some to cover it in this repo if we think it's useful.

That would be excellent. Thanks!

The tighter we can make the feedback loop on the code correctness, the better. If the tests live in the consumer app, we'd only know about a regression a change in the library introduces once the update library is integrated in the app. But if the tests are in the library itself, we would catch a regression before it has a chance to reach the consumers.

PR #601, opened 2023-05-17
Add site editor settings tests for block theme support

wargcm:
## Description

Adds some tests to cover the changes made in #598.

### Testing Details

- Run the tests
- **Verify** they pass

---


dvdchr (approved):
LGTM!

dvdchr:
I wonder if this needs a new version bump. Since there's no actual change on the lib code, I suppose not? 🤔

wargcm:
@dvdchr That was my thought process too. I figured since there wasn't any change to the library itself, it didn't need a bump.

@mokagio What are your thoughts? Should I bump the version?

mokagio (approved):
Thank you for promptly following up on this @wargcm !

It's reassuring to see there was already testing support, and it seems it was relatively straightforward to add in tests for the new behavior (I hope?) 😄 

I think we could benefit from splitting the JSON decoding tests from the ones of for the service behavior itself, but that would have been out of scope here. Just tossing it out a something to keep on the radar.

@dvdchr @wargcm +1 on not needing a version bump. If we built a new version from here, the production code that consumers would fetch would be exactly the same as the latest version.

PR #607, opened 2023-06-09
Fix flaky SharingServiceRemoteTests test

kean:
Fix flaky SharingServiceRemoteTests test

**Root cause**: `[RemotePublicizeService]` is constructed from a dictionary where the order of keys is not stable.

---

### DIFF
--- a/WordPressKitTests/SharingServiceRemoteTests.swift
+++ b/WordPressKitTests/SharingServiceRemoteTests.swift
@@ -39,17 +39,17 @@ final class SharingServiceRemoteTests: RemoteTestCase, RESTTestable {
         stubRemoteResponse(pathToStub, filename: publicizeServicesMockFilename, contentType: .ApplicationJSON)
 
         mockService.getPublicizeServices(for: mockID) { publicizeServices in
-            guard let firstService = publicizeServices.first else {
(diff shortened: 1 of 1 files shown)


momo-ozawa:
@mokagio 👋 

Just want to make sure - we don't need to make a new release for fixing tests, do we?

momo-ozawa (approved):

mokagio:
@momo-ozawa 

> we don't need to make a new release for fixing tests, do we?

Correct. No need for a release 👍

WordPressKitTests/SharingServiceRemoteTests.swift
```diff
@@ -39,17 +39,17 @@ final class SharingServiceRemoteTests: RemoteTestCase, RESTTestable {
         stubRemoteResponse(pathToStub, filename: publicizeServicesMockFilename, contentType: .ApplicationJSON)
 
         mockService.getPublicizeServices(for: mockID) { publicizeServices in
-            guard let firstService = publicizeServices.first else {
+            guard let facebookService = publicizeServices.first(where: { $0.serviceID == "facebook" }) else {
```
mokagio, line 42: @kean thank you for hunting this down! 🙇‍♂️ 

Out of curiosity, did you consider using `XCTUnwrap` here? Or does the fact that we're running the code inside the `getPublicizeServices...` closure prevent using `try`?

PR #610, opened 2023-06-20
Resolve TODOs by backporting async/await code

jkmassel:
### Description

While digging into async issues, I came across a couple of TODOs that were pretty quick to clean up. While we can't switch to async/await entirely at this point, enabling backporting is a step toward allowing it (even if just for use in our test suite).

### Testing Details
Ensure tests pass.

---

### DIFF
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -37,6 +37,10 @@ public enum ResponseType {
 
 open class WordPressComRestApi: NSObject {
 
+    enum Errors: Error {
+        case invalidResponse
+    }
+
     // MARK: Properties
 
     @objc public static let ErrorKeyErrorCode       = "WordPressComRestApiErrorCodeKey"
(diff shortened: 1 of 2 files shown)


WordPressKit/WordPressComRestApi.swift
```diff
@@ -498,10 +500,23 @@ open class WordPressComRestApi: NSObject {
             request.httpBody = parameters.percentEncoded()
         }
 
-        return try await urlSession.data(for: request, delegate: nil)
+        return try await withCheckedThrowingContinuation { continuation in
```
jkmassel, line 503: `urlSession.data(for:)` is only available in iOS15, so I had to wrap the method in async/await myself

leandroalonso:
@jkmassel it looks like the tests are hanging after `testSuccessfullGetCodableCall` 🤔

WordPressKit/WordPressComRestApi.swift
mokagio: [Jetpack/WordPress will soon support iOS 15 and above only](https://github.com/wordpress-mobile/WordPress-iOS/pull/20917) and [WooCommerce already does](https://github.com/woocommerce/woocommerce-ios/blob/trunk/WooCommerce/WooCommerce.xcodeproj/project.pbxproj#L13591).

Your diff doesn't add much complexity, but I wonder if it would be cleaner to bump the library to iOS 15 instead.

PR #613, opened 2023-06-23
[EU-US Compliance] Add `IPLocationRemote`

alpavanoglu:
### Description
This PR adds the necessary remote for the public `geo/` Endpoint. It will be used to check the location and decide whether to show the pop up or not.

https://github.com/wordpress-mobile/WordPress-iOS/issues/20711

### Testing Details
This can be tested from the main app once the remote is utilized with a following PR.

---

### DIFF
--- a/WordPressKit/IPLocationRemote.swift
+++ b/WordPressKit/IPLocationRemote.swift
@@ -24,6 +24,7 @@ public final class IPLocationRemote {
         let request = URLRequest(url: url)
         let task = urlSession.dataTask(with: request) { data, _, error in
             guard let data else {
+                completion(.failure(IPLocationError.requestFailure(error)))
                 return
             }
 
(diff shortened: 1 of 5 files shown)


WordPressKit/IPLocationRemote.swift
```diff
@@ -0,0 +1,53 @@
...
+        let path = WordPressComOAuthClient.WordPressComOAuthDefaultApiBaseUrl + "/geo/"
+        guard let url = URL(string: path) else {
+            return completion(.failure(IPLocationError.malformedURL))
+        }
+
+        let request = URLRequest(url: url)
+        let task = urlSession.dataTask(with: request) { data, _, error in
+            guard let data else {
+                return
+            }
```
salimbraksa, line 29: It would be great **not** to handle this silently, the completion handler won't be called. We can do the following:
```swift
let data = data ?? Date()
```
Then let the decoding throw an error.

salimbraksa (commented):
The new `IPLocationRemote` class looks good to me. I suggested only 1 minor improvement.

salimbraksa (approved):
LGTM! 🚀

PR #624, opened 2023-09-05
Add unit tests for the MediaServiceRemote.getMediaLibrary API

crazytonyli:
### Description

While working on the media areas on the Jetpack/WordPress app, I noticed the `-[MediaServiceRemote getMediaLibraryWithPageLoad:success:failure:]` method has somewhat unconventional behaviors. Its implementations (WP.com REST API and XML-RPC API) makes paginated API requests and returns the API call results to the caller via the `pageLoad` block. It also calls the `success` block with all the media items.

I thought it might be valuable to write tests to ensure the API behaviour is maintained in all future releases, and most importantly ensure the API behaviour is consistent between REST API and XML-RPC API implementations.

I created a `MediaLibraryTestSupport` type to stub "load Media Library" HTTP requests. The implementation is complicated enough that I feel like it's worthy adding a "meta test" for it. I'm open to that idea if you think it might be useful.

### Testing Details

It should be all good if CI passes.

---

### DIFF
--- a/WordPressKitTests/LoadMediaLibraryTests.swift
+++ b/WordPressKitTests/LoadMediaLibraryTests.swift
@@ -1,0 +1,110 @@
+import Foundation
+import XCTest
+
+@testable import WordPressKit
+
+private enum Kind {
+    case wpcom
+    case xmlrpc
+}
+
+// This test case is for the `-[MediaServiceRemote getMediaLibraryWithPageLoad:success:failure:]` method, which is
+// implemented in `MediaServiceRemoteREST` and `MediaServiceRemoteXMLRPC`. The test functions in this test case are
+// created to ensure both implementation shares the same behaviour.
+//
+// See the `-[MediaServiceRemote getMediaLibraryWithPageLoad:success:failure:]` API doc for its expected behaviours.
+class LoadMediaLibraryTests: XCTestCase {
+
+    fileprivate var kind: Kind = .wpcom
+
+    func testSmallLibrary() {
+        let mediaLibrary = MediaLibraryTestSupport(total: 50)
+        let (pageLoad, success, failure) = load(mediaLibrary: mediaLibrary, failAtPage: -1)
+        XCTAssertEqual(pageLoad.count, 0)
+        XCTAssertEqual(success?.count, 50)
+        XCTAssertNil(failure)
+    }
+
+    func testTwoPageLibrary() {
+        let mediaLibrary = MediaLibraryTestSupport(total: 120)
+        let (pageLoad, success, failure) = load(mediaLibrary: mediaLibrary, failAtPage: -1)
+        XCTAssertEqual(pageLoad.count, 1)
+        XCTAssertEqual(success?.count, 120)
+        XCTAssertNil(failure)
+    }
+
+    func testLargeLibrary() {
+        let mediaLibrary = MediaLibraryTestSupport(total: 650)
+        let (pageLoad, success, failure) = load(mediaLibrary: mediaLibrary, failAtPage: -1)
+        XCTAssertEqual(pageLoad.count, 6)
+        XCTAssertEqual(success?.count, 650)
+        XCTAssertNil(failure)
+    }
+
+    func testFailure() {
+        let mediaLibrary = MediaLibraryTestSupport(total: 550)
+        let (pageLoad, success, failure) = load(mediaLibrary: mediaLibrary, failAtPage: 3)
+        XCTAssertEqual(pageLoad.count, 2)
+        XCTAssertNil(success)
+        XCTAssertNotNil(failure)
+
+        let loaded = pageLoad.map { $0?.count ?? 0 }.reduce(0, +)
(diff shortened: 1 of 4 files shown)


WordPressKitTests/MediaLibraryTestSupport.swift
```diff
@@ -0,0 +1,196 @@
...
+    private var rpcStub: HTTPStubsDescriptor? {
+        didSet {
+            if let oldValue {
+                HTTPStubs.removeStub(oldValue)
+            }
+        }
+    }
+
+    init(total: Int) {
+        media = (1...total).map { id in
```
mokagio: What do you think of `totalMedia`?

```swift
    init(totalMedia: Int) {
        media = (1...totalMedia).map { id in
```

WordPressKitTests/MediaLibraryTestSupport.swift
```diff
@@ -0,0 +1,196 @@
...
+            return .init(error: URLError(.cannotDecodeContentData))
+        }
+
+        let delegate = RequestParser()
+        parser.delegate = delegate
+        guard parser.parse() else {
+            return .init(data: Data(), statusCode: 200, headers: nil)
+        }
+
+        assert(delegate.methodName == "wp.getMediaLibrary")
```
mokagio: Have you considered using `XCTAssertEqual` here? Just so the test would fail instead of crash?

WordPressKitTests/LoadMediaLibraryTests.swift
```diff
@@ -0,0 +1,110 @@
...
+/// - SeeAlso `-[MediaServiceRemote getMediaLibraryWithPageLoad:success:failure:]`
+private typealias MediaLibraryResult = (pageLoad: [[Any]?], success: [Any]?, failure: Error?)
+
+class LoadMediaLibraryRPCTests: LoadMediaLibraryTests {
+
+    override func setUp() {
+        kind = .xmlrpc
+    }
+
+}
```
mokagio, line 110: Have you considered duplicating the test methods and passing a `Kind` input to `load(mediaLibrary: MediaLibraryTestSupport, failAtPage: Int)`? I think it would make the test results clearer albeit at the cost of some duplication.

What I mean with clearer is that if a test fails in `LoadMediaLibraryRPCTests` we would see the failure in Xcode at the `LoadMediaLibraryTests`, which might be lead someone looking for the source of the failure into looking in the wrong implementation class.

WordPressKitTests/LoadMediaLibraryTests.swift
```diff
@@ -0,0 +1,110 @@
...
+            result.pageLoad.append($0)
+        } success: {
+            result.success = $0
+            finished.fulfill()
+        } failure: { error in
+            result.failure = error
+            finished.fulfill()
+        }
+
+        wait(for: [finished], timeout: 5000)
```
mokagio: I'm assuming the XCTest-level wait for expectation here is used as a mechanism to make the code sync, not because there is async behavior that is being tested, right? Otherwise, I would expect to see some `XCTAssert*` in this method.

If that's the case, have you considered using structured concurrency like `async throws -> (pageLoad: [[Any]?], success: [Any]?)`

mokagio (approved):
Left a few structure suggestions, but I don't consider them blockers.

I was surprised to see Hound barking here. I must have forgotten to configure standalone SwiftLint and/or disable Hound. Leave it with me.

WordPressKitTests/LoadMediaLibraryTests.swift
crazytonyli: I wanted to avoid duplication, because I wanted to use the same tests to ensure REST and XML-RPC implementations have the same behaviour. Essentially, I want [rspec's shared examples](https://rspec.info/features/3-12/rspec-core/example-groups/shared-examples/), but I don't know what's the best way to do that in XCTest.

WordPressKitTests/LoadMediaLibraryTests.swift
crazytonyli: > I'm assuming the XCTest-level wait for expectation here is used as a mechanism to make the code sync, not because there is async behavior that is being tested, right?

Yes, correct. I didn't think about using async function. I guess they are different mechanism that achieving the same thing. It's just that we happen to use XCTest which comes with a wait mechanism.

An issue with `async throws -> (pageLoad: [[Any]?], success: [Any]?)` is, it's a big tricky to test the function throws an error: I don't think `XCTestAssertThrow` supports using an async closure. That means, we'll need to add try-catch, or expectation to the test function. The current sync version seems simpler in this aspect.

mokagio: > An issue with async throws -> (pageLoad: [[Any]?], success: [Any]?) is, it's a big tricky to test the function throws an error: I don't think XCTestAssertThrow supports using an async closure. That means, we'll need to add try-catch, or expectation to the test function. The current sync version seems simpler in this aspect.

🤔 Yeah. I didn't look into that. I'm not sure.

PR #626, opened 2023-09-08
Implement All Domains endpoint

### LINKED ISSUE #21556: 🍎 Domains List Endpoint

salimbraksa:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/21556

### Related PR
This PR is blocking the following:
- https://github.com/wordpress-mobile/WordPress-iOS/pull/21805

### Description

Implement API to execute All Domains endpoint.

### Testing Details

 See https://github.com/wordpress-mobile/WordPress-iOS/pull/21805 Test Instructions.

---

- [ ] I have considered updating the `version` in the `.podspec` file.


WordPressKitTests/AllDomainsResultDomainTests.swift
```diff
@@ -0,0 +1,160 @@
...
+
+final class AllDomainsResultDomainTests: XCTestCase {
+
+    // MARK: - Properties
+
+    private let dateFormatter = ISO8601DateFormatter()
+
+    // MARK: - Tests
+
+    func testDecoding1() throws {
```
hassaanelgarem: [Suggestion] I suggest using more descriptive test names. I see each test is testing a different form of input. Maybe we can use name that reflect the differences.

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
```diff
@@ -0,0 +1,133 @@
...
+                case .failure(let error):
+                    throw error
+                }
+            } catch let error {
+                completion(.failure(error))
+            }
+        }
+    }
+
+    private func allDomainsEndpointPath(params: AllDomainsEndpointParams?) -> String {
```
hassaanelgarem: The request is missing the `no_wpcom=true` query. From the project spec:
> To retrieve the list of domains we’ll make an authenticated GET request to https://public-api.wordpress.com/rest/v1.1/all-domains?no_wpcom=true&resolve_status=true&locale={LOCALE}

hassaanelgarem (commented):
Other than the comment I added, the changes look good to me. And I tested using the linked WPiOS PR and it works as described!

I also noticed that the new `testGetAllDomainsSucceeds` is failing.

WordPressKitTests/AllDomainsResultDomainTests.swift
salimbraksa: @hassaanelgarem I agree. I spent a lot of time trying to come up with descriptive test names, but I wasn't happy with the result. Here is what I tried:

- Before `testDecoding1`, it was `testDecodingWithDefaultInput`.
- Before `testDecoding2`, it was `testDecodingWithEmptyExpiryDateAndRegistrationDate`
- Before `testDecoding3`, it was `testDecodingWithMappingTypeAndHasRegistrationIsTrue`

The difference between the test cases is the input, the idea was to test the decoding logic with different input. And it's difficult to come with test names describing the difference.

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
salimbraksa: Good catch. Resolved in f6f8461df348031f5eba645a488463f2f878daaa

salimbraksa:
> I also noticed that the new testGetAllDomainsSucceeds is failing.

Thanks for testing @hassaanelgarem. The latest commit should fix the failing test.

WordPressKitTests/AllDomainsResultDomainTests.swift
hassaanelgarem: I think having long test names is more than fine. I like the names you originally thought of 👍

hassaanelgarem (approved):
LGTM 🚀

WordPressKitTests/AllDomainsResultDomainTests.swift
salimbraksa: Resolved in 4328bad561ac715bfb377695aa2397b016a3bfef

salimbraksa:
Hey @mokagio , is this item in the checklist still needed?
> I have considered updating the version in the .podspec file.

WordPressKitTests/AllDomainsResultDomainTests.swift
mokagio: Something I've seen in the Apple docs is also to supplement test names that might be hard to read with a documentation comment (`///`, not `//`):

```swift
/// When expiry date and registration date are given,
/// Then the output is ...
func testDecodingWithEmptyExpiryDateAndRegistrationDate() {
```

**Update** I see you used that style in https://github.com/wordpress-mobile/WordPressKit-iOS/commit/4328bad561ac715bfb377695aa2397b016a3bfef 😄

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
```diff
@@ -0,0 +1,137 @@
...
+    // MARK: - Public Types
+
+    public typealias AllDomainsEndpointResult = Result<[AllDomainsListItem], Error>
+
+    public struct AllDomainsEndpointParams {
+        public var resolveStatus: Bool?
+        public var locale: String?
+        public var noWPCOM: Bool?
+        public init() {}
+    }
```
mokagio: I don't see any of the three properties being mutated in the implementation. Have you considered making them `let`? I also don't see them ever being assigned `nil` value, have you considered making them non-optional?

**Update:** After looking a the code more, I'm guessing the params are `Optional` because the third-party consumer of the method to call the API may decide not to pass them. That assume those parameters are not mandatory, but I can't find any documentation for the endpoint to learn about it 😞

WordPressKitTests/AllDomainsResultDomainTests.swift
```diff
@@ -0,0 +1,163 @@
...
+        XCTAssertEqual(domain.isDomainOnlySite, otherDomain.isDomainOnlySite)
+        XCTAssertEqual(domain.isWpcomStagingDomain, otherDomain.isWpcomStagingDomain)
+        XCTAssertEqual(domain.registrationDate, otherDomain.registrationDate)
+        XCTAssertEqual(domain.expiryDate, otherDomain.expiryDate)
+        XCTAssertEqual(domain.wpcomDomain, otherDomain.wpcomDomain)
+        XCTAssertEqual(domain.currentUserIsOwner, otherDomain.currentUserIsOwner)
+        XCTAssertEqual(domain.siteSlug, otherDomain.siteSlug)
+        XCTAssertEqual(domain.status?.value, otherDomain.status?.value)
+        XCTAssertEqual(domain.status?.type, otherDomain.status?.type)
+    }
```
mokagio, line 72: Have you considered making `Domain`, aka `DomainsServiceRemote.AllDomainsListItem`, `Equatable`?

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
```diff
@@ -238,4 +243,37 @@ class DomainsServiceRemoteRESTTests: RemoteTestCase, RESTTestable {
...
+    func testGetAllDomainsSucceeds() {
+        let expect = expectation(description: "Get All Domains Succeeds")
+
+        stub { request -> Bool in
+            let url = request.url
+            let matched = url?.path == self.allDomainsEndpoint && url?.query == "no_wpcom=true&resolve_status=true&locale=en"
+            XCTAssertTrue(matched)
+            return matched
+        } response: { request in
+            let path = OHPathForFile(self.allDomainsMockFilename, type(of: self)) ?? ""
```
mokagio: I think it would be okay to crash here, as if the expected file is not there, it's a developer error.

```suggestion
            let path = OHPathForFile(self.allDomainsMockFilename, type(of: self))!
```

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
```diff
@@ -238,4 +243,37 @@ class DomainsServiceRemoteRESTTests: RemoteTestCase, RESTTestable {
...
     }
+
+    // MARK: - Get All Domains Tests
+
+    func testGetAllDomainsSucceeds() {
+        let expect = expectation(description: "Get All Domains Succeeds")
+
+        stub { request -> Bool in
+            let url = request.url
+            let matched = url?.path == self.allDomainsEndpoint && url?.query == "no_wpcom=true&resolve_status=true&locale=en"
```
mokagio: I assume the check on the query here is to ensure that the parameters object is read and used correctly to build the request? 

If so, two notes. First, it's worth noting that we have no guarantees in how Foundation will decide to build the query string. I could end up being different, e.g. with `locale=en` as the first parameter. OHHTTPStubs offers a condition checker for query parameters that is robust against changes in the query string order:

```swift
public func containsQueryParams(_ params: [String:String?]) -> HTTPStubsTestBlock {
```

There is also a similar method for the path, 

```swift
public func isPath(_ path: String) -> HTTPStubsTestBlock {
```

Second, and more of a test style and readability thing, so partly subject to personal opinion. The query is defined up here and the parameters down there. It's not clear that they are related to each other. If they the params were defined before the stub and maybe the check read their value (`containsQueryParams(["no_wpcom": parameters.noWPCOM, ...`) the relationship would be clearer.

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
```diff
@@ -238,4 +243,37 @@ class DomainsServiceRemoteRESTTests: RemoteTestCase, RESTTestable {
...
+                let expectedCount = 10
+                XCTAssertEqual(domains.count, expectedCount, "There should be \(expectedCount) domains returned.")
+            case .failure:
+                XCTFail("Get All Domains request failed")
+            }
+            expect.fulfill()
+        }
+
+        waitForExpectations(timeout: timeout)
+    }
```
mokagio, line 326: Did you consider adding a test for the failure path?

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
```diff
@@ -0,0 +1,137 @@
+import Foundation
+
+extension DomainsServiceRemote {
+
+    // MARK: - API
+
+    public func fetchAllDomains(params: AllDomainsEndpointParams? = nil, completion: @escaping (AllDomainsEndpointResult) -> Void) {
+        let path = self.allDomainsEndpointPath(params: params)
+        self.wordPressComRestApi.GET(path, parameters: nil) { result, _ in
```
mokagio: Have you considered relying on the `parameters` argument of the `GET(_, parameters:)` API client to build the URL query, as opposed to implement the logic manually here?

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
mokagio: Also, the core of the logic here is how the parameters translate to the query string. The implementation checks the for a value and only adds it to the query string then. Have you considered testing that in isolation and in more details? With in isolation, I'm thinking a method on `AllDomainsEndpointParams` that spits out the `[URLQueryItem]` (or `[String: Any]` if we rely on the API client method).

mokagio:
> Hey @mokagio , is this item in the checklist still needed?
> 
> > I have considered updating the version in the .podspec file.

Good point. I think we can do without that check. The idea behind the changelog file was to collect changes by type, so that when someone makes a new release, they can decide which version to use based on the type of changes (breaking => major, new feature => minor, only bug fixes => patch).

I'd appreciate a PR on the GitHub template file to remove the line. Thanks!

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
salimbraksa: @mokagio Good point, I didn't pay attention to that `parameters` argument. But I agree it should be the one to use.

WordPressKitTests/AllDomainsResultDomainTests.swift
salimbraksa: @mokagio Yes I have, but that deep comparison logic is only needed in a test environment. In a non-test environment, usually these models are considered equal if their primary attributes are equal. In the context of the `Domain` type, the primary attribute is `domain`.

So my point is that implementation of `Equatable` protocol is going to be different between a test and non-test environment.

```swift
// non-test environment
extension Domain: Equatable {
   func ==(left: Domain, right: Domain) -> Bool {
         return left.domain == right.domain
   }
}

// test environment
extension Domain: Equatable {
     func ==(left: Domain, right: Domain) -> Bool {
         // compare all properties.
   }
}
```

salimbraksa: I believe that if a specific logic or behavior is solely required for testing purposes, it should be confined to the test environment.

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
salimbraksa: Resolved in 9d0a555a91d0db7a158f8bc613176a8be085180c

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
salimbraksa: Resolved in 9d0a555a91d0db7a158f8bc613176a8be085180c

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
salimbraksa: > Did you consider adding a test for the failure path?

There are 3 cases where the `fetchAllDomains` could fail:
- The `params` encoding fails.
- The request fails due to networking issue, or the server returns 400 or 500 error. 
- The json decoding fails.

I added new unit tests covering the 1st case 677fc351ab45f8e919b0c313e55269b16e2dd6ca. Not sure we can do much about the 2nd case though.

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
salimbraksa: 1703cfb8bdf526b420302e814e3276d0164f2d62 adds a short documentation for `fetchAllDomains` method.

jkmassel: Looking at the backend code, it seems `no_wpcom` is always valid on the endpoint, as is `resolve_status`. Could these use good defaults to avoid nullability?

salimbraksa: @jkmassel What do you mean by they're valid on the endpoint?

salimbraksa:
Hey @mokagio let me know if there is anything else that needs to change. 🙏

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
hassaanelgarem: @salimbraksa I think what Jeremy means here is that the endpoint will always be able to handle these parameters. So if we always send a value (the same value as the default values on the endpoint) that would save us the hassle of handling optionals.
The default values on the endpoint are:
* `no_wpcom`: false
* `resolve_status`: false
Ref: fbhepr%2Skers%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Sjcpbz%2Qwfba%2Qraqcbvagf%2Spynff.jcpbz%2Qfgber%2Qqbznvaf%2Qncv%2Qraqcbvagf.cuc%3Se%3Q28672n42%23927-og

salimbraksa: Thanks @hassaanelgarem. Resolved in 423a5f005aeb1bb82d32c69504fff9fe8c396d81.

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
```diff
@@ -0,0 +1,182 @@
...
+    public typealias AllDomainsEndpointResult = Result<[AllDomainsListItem], Error>
+
+    public struct AllDomainsEndpointParams {
+
+        public var resolveStatus: Bool = false
+        public var noWPCOM: Bool = false
+        public var locale: String?
+
+        public init() {}
+    }
```
hassaanelgarem, line 63: [Suggestion] Have you considered doing something like this?
```suggestion
    public struct AllDomainsEndpointParams {

        let resolveStatus: Bool
        let noWPCOM: Bool
        let locale: String?

        public init(resolveStatus: Bool = false, noWPCOM: Bool = false, locale: String?) {
            self.resolveStatus = resolveStatus
            self.noWPCOM = noWPCOM
            self.locale = locale
        }
    }
```

hassaanelgarem (approved):
LGTM!
Left 1 suggestion on the code

salimbraksa:
> LGTM!
> 
> Left 1 suggestion on the code

@hassaanelgarem Aoutch, sorry I merged this PR before seeing your suggestion. 🙇

WordPressKit/Domains/DomainsServiceRemote+AllDomains.swift
salimbraksa: Yes, but I wasn't sure about rewriting those properties in the init. Consider a situation where the `struct` had +10 properties, then the init would have the same +10 properties, In my opinion, that's redundant. 

In UIKit and Foundation, there are many `structs` that you can mutate their properties directly. So I think the practice of mutable struct properties is not uncommon.

If you feel strongly about this, I can open another PR and make the change.

mokagio: > In UIKit and Foundation, there are many `structs` that you can mutate their properties directly. So I think the practice of mutable struct properties is not uncommon.

Thank you for your rationale @salimbraksa 😄 

I feel strongly about this, actually. I shall P2 about this to make my case. It would be good to bubble up and have an agreed upon guideline. In the meantime, as far as I'm concerned, the implementation can stay like this. 👍

WordPressKitTests/AllDomainsResultDomainTests.swift
mokagio: > I believe that if a specific logic or behavior is solely required for testing purposes, it should be confined to the test environment.

Agreed. At the same time, if we compromised and defined `Equatable` in the production code, this code would simplify to


```swift
extension DomainsServiceRemote.AllDomainsListItem: Equatable {}
extension DomainsServiceRemote.AllDomainsListItem.Status: Equatable {}

XCTAssertEqual(output, expectedOutput)
```

With the added benefit that the compiler would update the check for us if a property is added, removed, or changes type.

Seems like a worthy trade to me.

mokagio: > that deep comparison logic is only needed in a test environment. In a non-test environment, usually these models are considered equal if their primary attributes are equal

Interesting... 🤔 Is that justified? I can see how `a.domain != b.domain` obviously means that the `Domain` are different. But what about `a.domain == b.domain; a.isWpcomStagingDomain != b.isWpcomStagingDomain`? That should count as a different domain for me. 

I can see the argument "in practice that will never happen," but that's a weak assumption to make. If the domain modeling allows it to happen, then we I think ought to handle it.

That's a broader conversation than this PR, though.

WordPressKitTests/DomainsServiceRemoteRESTTests.swift
mokagio: There are ways to test both scenarios. I'll see if I can make an example PR about it, maybe next week.

PR #630, opened 2023-10-11
Set RemotePostCategory.parent to zero when the JSON property is nil

### LINKED ISSUE #21138: Core Data error: Non-optional properties with a nil value on Category
Sentry Issue: [JETPACK-IOS-P3](https://a8c.sentry.io/issues/3948401126/?referrer=github_integration)

```
Unresolved Core Data save error (Derived Context): Multiple errors:
  1: Non-optional property with a nil value on Category.categoryID
  2: Non-optional property with a nil value on Category.categoryName
  File "ContextManager+ErrorHandling.swift", line 92, in ContextManager.handleSaveError
  File "ContextManager+ErrorHandling.swift", line 75, in ContextManager.internalSave
  File "ContextManager.swift", line 228, in ContextManager.save
  File "ContextManager.swift", line 251, in ContextManager.save
  File "ContextManager.swift", line 122, in ContextManager.performAndSave
...
(12 additional frame(s) were not displayed)
```

The issue occurred 67 times over the past 30 days. First detected on version 21.7.

crazytonyli:
### Description

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/21138.

The above issue is a site related issue and doesn't affect many users. The `parent` JSON property should be `0` if the category is at top level, and non-zero if it's at child level. But `parent` should never be nil. 

However, in very rare cases, it appears WordPress REST API does return `nil` which causes the app attempts to store an invalid nil value. I tried to look into the backend code and WordPress site settings to see if I can reproduce this behaviour. But I couldn't figure out how.

This PR assigns a default value `0` to the `RemotePostCategory.parent` property, essentially treating a `"parent": null` response same as `"parent": 0` (a.k.a a top level category).

### Testing Details

I have added an unit test to verify the new behaviour.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- `RemotePostCategory.parent` is set to zero when API returns `"parent": null` [#630]
 
 ### Bug Fixes
 
--- a/WordPressKit/TaxonomyServiceRemoteREST.m
+++ b/WordPressKit/TaxonomyServiceRemoteREST.m
@@ -293,7 +293,7 @@ static NSUInteger const TaxonomyRESTNumberMaxValue = 1000;
     RemotePostCategory *category = [RemotePostCategory new];
     category.categoryID = [jsonCategory numberForKey:TaxonomyRESTIDParameter];
     category.name = [jsonCategory stringForKey:TaxonomyRESTNameParameter];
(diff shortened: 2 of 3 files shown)


mokagio (approved):
Thank you for digging into this and for adding a test to verify the fix!

---

> However, in very rare cases, it appears WordPress REST API does return nil which causes the app attempts to store an invalid nil value. I tried to look into the backend code and WordPress site settings to see if I can reproduce this behaviour. But I couldn't figure out how.

cc @hannahtinkler

PR #629, opened 2023-10-08
Add Webauthn endpoints

Ecarrion:
Part of https://github.com/woocommerce/woocommerce-ios/issues/10854
Part of https://github.com/woocommerce/woocommerce-ios/issues/10855

# Why

This PR adds the necessary endpoints for the security key authentication.

These will be used on `WordpressAuthenticator-IOS` and later on the Woo / Jetpack and / WP apps.

# How

- Update the `oauth/token` endpoint to store the `webauthn` nonce.
- Add the `webauthn-challenge-endpoint` endpoint to request a challenge.
- Add the `webauthn-authentication-endpoint` endpoint to verify a signed challenge

This is all coded to the spec defined pe5pgL-3Nh-p2

# Demo

https://github.com/wordpress-mobile/WordPressKit-iOS/assets/562080/2f69ea2f-b680-4d0a-8840-3a76540f935d

# Testing

You can believe the video 😅 or you can test this [experimental branch](https://github.com/woocommerce/woocommerce-ios/pull/10904)

## Notes

- Compiling the experimental branch may be a bit buggy, Cocoapods doesn't work that well with unreleased versions of a pod.
- You need to have a WordPress.com account with a non-A8C email to test social logins
- Magic links do not work. pe5pgL-3QM-p2

## Steps

- Launch the woo app from the [experimental branch](https://github.com/woocommerce/woocommerce-ios/pull/10904)
- Continue with user name/password OR continue with Google or Apple with a non-A8C email.
- See that the two-factor screen presents a "Use a security key" CTA
- Select that option and select your passkey stored in the iOS keychain.
- See that you are correctly authenticated.

---

- [x] Please check here if your pull request includes additional test coverage.
- [x] I have considered updating the `version` in the `.podspec` file.
- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #632, opened 2023-10-16
Make `authenticateWithUsername` `needsMultifactor` parameter backward compatible

mokagio:
### Description
The parameter being optional and nil by default ensure compatibility with the previous signature which didn't have the parameter at all.

This addresses 0ea80b07c2827ca897718cfe458890bec7328a52 inadvertently breaking compatibility.

See this CI build for a verification of the broken compatibility:
https://buildkite.com/automattic/woocommerce-ios/builds/16542#018b2c09-3960-48f0-93fb-b430349c1dd2/759-2868

Additionally, the method will run the `failure` closure with a custom error if the API response prompted an MFA check but no closure was given. I considered forcing a crash to signify a developer error but decided to "throw" an error instead in the spirit of not breaking compatibility. 

I considered updating the library to reflect the breaking change instead, but the original PR, #629, was explicitly trying to maintain backwards compatibility, something worth respecting.

One final note. I renamed the test file to match the SUT name, but because I also added tests, it looks like a file removal and addition in the diff. For ease of review, the relevant commits are:

- ddfa93f8ee1a29f17d1e8e5921f0ec880cd1bab7
- 4365315e473887e665fe19ef96e25f44fe6cdb60

I copy pasted tests to handle the the no `needsMultifactor` parameter code path. I could have spent more time on this and DRY things, but this is already holding up the code freeze. 

### Testing Details

See updated unit tests.

---

- [x] I have considered updating the `version` in the `.podspec` file. – _I'll do that in a dedicated PR after merging._


WordPressKit.xcodeproj/project.pbxproj
```diff
@@ -1749,27 +1749,10 @@
...
-				93AB06031EE8838400EF8764 /* RemoteTestCase.swift */,
-				74A923B11F2BE2DF00EC8F92 /* RESTTestable.swift */,
-				3F3C9E9B289A3E31009A1357 /* TestCollector+Constants.swift */,
-				FFE247A620C891D1002DF3A2 /* WordPressComOAuthTests.swift */,
-				74B335D91F06F3D60053A184 /* WordPressComRestApiTests.swift */,
-				8B9F0CAD2762414F00DBE144 /* WordPressComRestApiTests+AsyncAwait.swift */,
-				73B3DAD521FBB20D00B2CF18 /* WordPressComRestApiTests+Locale.swift */,
-				FFA4D4A82423B10A00BF5180 /* WordPressOrgRestApiTests.swift */,
-				74B335DB1F06F4180053A184 /* WordPressOrgXMLRPCApiTests.swift */,
-				740B23D51F17F7C100067A2A /* XMLRPCTestable.swift */,
```
mokagio, line 1769: Used this as a chance to sort the group alphabetically

WordPressKitTests/WordPressComOAuthClientTests.swift
```diff
@@ -26,6 +26,31 @@ class WordPressComOAuthClientTests: XCTestCase {
...
+            return fixture(filePath: stubPath!, headers: ["Content-Type" as NSObject: "application/json" as AnyObject])
+        }
+
+        let expect = expectation(description: "One callback should be invoked")
+        let client = WordPressComOAuthClient(clientID: "Fake", secret: "Fake")
+        client.authenticateWithUsername(
+            "fakeUser",
+            password: "fakePass",
+            multifactorCode: nil,
+            success: { (token) in
```
mokagio, line 41: Notice there is no `needsMultifactor` parameter here. This replicates the syntax as it was previously while also exercises the code behavior with the default parameter.

WordPressKit.xcodeproj/project.pbxproj
AliSoftware: 💡  We should probably use https://github.com/venmo/synx at some point on our projects to resynchronize the Finder with the structure of the pbxproj, and let it organize groups and files alphabetically automatically.

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -10,6 +10,11 @@ import Alamofire
     case socialLoginExistingUserUnconnected
     case invalidTwoStepCode
     case unknownUser
+    /// The API response requires handling MFA, but the caller didn't specify a handler.
+    ///
+    /// Use to ensure backwards compatibility in `authenticateWithUsername(...)` calls after support for the MFA closure
+    /// was introduced.
+    case noMultifactorHandlerGiven
```
AliSoftware, line 17: 🚨  Strictly speaking, adding a `case` to an `enum` could be **considered a breaking change** though, because if the client code already has a `switch` on that error value, the switch which was exhaustive before will become non-exhaustive (except if it has an `@unknown default` case)

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -126,7 +131,7 @@ public final class WordPressComOAuthClient: NSObject {
     @objc public func authenticateWithUsername(_ username: String,
                                                password: String,
                                                multifactorCode: String?,
-                                               needsMultifactor: @escaping (_ userID: Int, _ nonceInfo: SocialLogin2FANonceInfo) -> Void,
+                                               needsMultifactor: ((_ userID: Int, _ nonceInfo: SocialLogin2FANonceInfo) -> Void)? = .none,
```
AliSoftware, line 134: ❓  Why not providing a noop closure instead?
```suggestion
                                               needsMultifactor: @escaping (_ userID: Int, _ nonceInfo: SocialLogin2FANonceInfo) -> Void = { _ in },
```

Or was this explicitly because you wanted to handle the case of calling `failure` with an error if no closure was provided explicitly?

AliSoftware (approved):
Approving to unblock the code freeze, but note that adding a `case` to an `enum` is also a breaking change, so I'm not sure this will really solve the core of your problem?

WordPressKit/WordPressComOAuthClient.swift
mokagio: Thanks for catching this @AliSoftware... How embarrassing: Tries to fix a breaking change, introduces a breaking change. I have a feeling there's a meme for that.

It turns out that I was too optimistic in my approach anyway. The fix addresses the API compatibility in Swift, but not Objective-C, and that's where the issue actually is. It was written in plain sight [in the CI error](https://buildkite.com/automattic/woocommerce-ios/builds/16542#018b2c09-3960-48f0-93fb-b430349c1dd2/759-2868), if only I had looked properly:

image

It's already late in my day... I'm going to start the code freeze without updating WordPressKit and continue to work on this tomorrow 😞

WordPressKit/WordPressComOAuthClient.swift
mokagio: > Or was this explicitly because you wanted to handle the case of calling failure with an error if no closure was provided explicitly?

That was the intention. In order to communicate to out-of-date clients that they should do _something_, but without a breaking change or crashing at runtime.

PR #634, opened 2023-10-17
Add tag parameter to /posts

kean:
### Description

Add `tag` parameter to /posts. Test in Postman.

---


momo-ozawa (approved):
Thanks! Can you update the podspec version before merging?

WordPressKit/PostServiceRemoteOptions.h
```diff
@@ -77,4 +77,9 @@ typedef NS_ENUM(NSUInteger, PostServiceResultsOrdering) {
  */
 - (NSString *)meta;
 
+/**
+ The tag to filter by.
+ */
+- (NSString *)tag;
```
mokagio, line 83: @kean @momo-ozawa this turned out to be a breaking change. I reverted it https://github.com/wordpress-mobile/WordPressKit-iOS/pull/639) to unblock the [23.6](https://github.com/wordpress-mobile/WordPress-iOS/milestone/260) code freeze. 

I'm not sure what's the best way forward here. We could simply ship a new major version, but I wonder if there are other ways around this, e.g. migrating the protocol to Swift and providing a default value (can you do that?) or declaring this as `@optional`? I don't know if that would work hey, just an idea to explore.

mokagio: Another idea, it looks like Woo doesn't use this protocol. I wonder how much work it would be to move it into the Jetpack/WordPress domain, so we could iterate faster in the future?

image

Doing so would definitely be a breaking change, by the way 😅

kean: I would ship it because it's a small change, and migrating to it simply requires following the suggested fix by the compiler.

Making it `@optional` should work, but it'll require changing this pod again, which I'm trying to minimize 😬 If we were to rework it, I would consider making it a `struct` or a `class` instead of a protocol.

mokagio: > Making it `@optional` should work, but it'll require changing this pod again, which I'm trying to minimize 😬

Fair. In the scheme of things, it just occurred to me that it would be the best option. 

At the moment, WordPressAuthenticator depends on this library. Bumping its major version would require updating WordPressAuthenticator and its clients, WooCommerce and Jetpack/WordPress.

If we can maintain compatibility, then we only need to update WordPressKit. Jetpack/WordPress already support it in your branch, WordPressAuthenticator and WooCommerce can update automatically without code changes.

Worth noting that, because I reverted adding `tag` in 8.9.1, we'll need to do _some_ additional work here regardless, even if it was just reverting the revert.

kean: Right, that complicates things. Yeah, let's make it easier for folks who end up updating these other repos. I'll try to add `@optional` to see if it works and will push a new version if it does.

PR #644, opened 2023-11-06
Add support for no-site carts

hassaanelgarem:
### Description

This PR adds support for creating a shopping cart that's not tied to a specific site.

Part of https://github.com/wordpress-mobile/WordPress-iOS/issues/21780

---

### DIFF
--- a/WordPressKit/TransactionsServiceRemote.swift
+++ b/WordPressKit/TransactionsServiceRemote.swift
@@ -151,12 +151,12 @@ public enum TransactionsServiceProduct {
 
 public struct CartResponse {
     let blogID: Int
-    let cartKey: Any
+    let cartKey: Any // cart key can be either Int or String
     let products: [Product]
 
     init?(jsonDictionary: [String: AnyObject]) {
         guard
-            let cartKey = jsonDictionary["cart_key"] as? Any,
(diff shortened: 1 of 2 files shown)


WordPressKit/TransactionsServiceRemote.swift
```diff
@@ -151,12 +151,12 @@ public enum TransactionsServiceProduct {
 
 public struct CartResponse {
     let blogID: Int
-    let cartKey: Int
+    let cartKey: Any
```
salimbraksa: ❓ I'm not very familiar with the `me/shopping-card` endpoint, so I'm not sure what to expect from this property of type `Any`. What are the possible values of this property? Maybe a short documentation can be useful here.

hassaanelgarem: Normally the cartKey is an integer, but in the case of a no-site cart, the cart Key is a string with the value "no-site". Hence the change to `Any`.

hassaanelgarem: > Maybe a short documentation can be useful here.

Agreed! Added in 00c748e20adea36dd6208efcde49b4d336c61fae

salimbraksa (approved):
LGTM!

PR #646, opened 2023-11-08
Update WP.com login URL to use login.php

guarani:
### Description

This change allows the app to receive a list of supported 2FA types such as SMS OTPs, authenticator-based OTPs, and passkeys.

### Testing Details

Test via https://github.com/wordpress-mobile/WordPress-iOS/pull/22001

---

### DIFF
--- a/WordPressKit.podspec
+++ b/WordPressKit.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressKit'
-  s.version       = '8.10.0'
+  s.version       = '8.10.1-beta.1'
 
   s.summary       = 'WordPressKit offers a clean and simple WordPress.com and WordPress.org API.'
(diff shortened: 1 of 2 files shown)


guarani:
👋 Hey @Ecarrion, looping you in here since this PR aims to allow users to log in with whichever auth type they have available. For example, if the user has app-based 2FA and passkeys on their account, they should (AFAIK) be able to log in using either. However, before this change, only passkeys worked for me in this scenario. 

Do you agree with using `login.php`? I recall this coming up in discussions to some extent, but I'm unsure what were the pros/cons of this approach vs. the dedicated REST API that was used.

momo-ozawa (approved):
Tested via https://github.com/wordpress-mobile/WordPress-iOS/pull/22001

guarani:
Closing as not being worked on.

PR #648, opened 2023-11-13
Merge 8.11.0 stable release

mokagio:
This version bump PR is part of the code freeze workflow for WordPress iOS 23.7 and is here only to leave a breadcrumb in the release process. Once CI is green, I will push the tag for this version and, once that is successfully deployed, admin-merge this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ _None._
 
 _None._
 
+### New Features
+
+_None._
+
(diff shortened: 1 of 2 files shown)


mokagio:
Note that I shipped this manually because for some reason CI couldn't find the 14.2 image.

I decided not to investigate because we should focus on moving off relying on the older toolchain instead.

PR #650, opened 2023-11-28
Make WordPressComOAuthError conforms to Swift.Error

crazytonyli:
### Context

In order to dropping Alamofire, we need to remove the `AFError` out of the picture too. Currently, most of the existing APIs returns an `NSError` type which can be `AFError` underneath. That means it's possible that the error handling code simply grabs `AFError.errorDescription` to display to user.

To reduce the impact to the existing error handling code, which is spread everywhere (this library, the authenticator library, and the apps), the new error type that's used to replace AFError needs to be localized. That means, if somehow an error case is missed after transition from Alamofire to URLSession, and the new error is displayed to user, at least it's a user friendly error message, instead of a cryptic message generated by Swift like "WordPressXxxError.specialErrorCase (code: 1)"

### Changes

There are two main changes:
1. `enum WordPressComOAuthError` turns into a Swift enum that conforms to Swift.Error. That means it's no longer visible to Objective-C code (*).
2. All APIs in the OAuth client class now returns concrete type `WordPressComOAuthError`, instead of `NSError`.

(*) BTW, all the changes I'm about to make to WordPressKit and WordPressAuthenticator will be Swift first. A Swift friendly API is the first priority. If the new API can be bridged to Objective-C, then great. If it can't, I'll either create a Objective-C friendly wrapper or (at the worst case) translate the Objective-C callers to Swift.

### Breaking change

This PR introduces break changes. But as far as I can tell, the API is only used in WordPressAuthenticator, not the apps.

### Testing Details

Use the test build in https://github.com/wordpress-mobile/WordPress-iOS/pull/22086 to verify WP.com login errors can be displayed correctly.

---


WordPressKitTests/WordPressComOAuthClientTests.swift
```diff
@@ -277,10 +271,11 @@ class WordPressComOAuthClientTests: XCTestCase {
...
             failure: { error in
-                expect.fulfill()
-                XCTAssertEqual(error.domain, WordPressComOAuthClient.WordPressComOAuthErrorDomain)
-                XCTAssertEqual(error.code, WordPressComOAuthError.unknown.rawValue)
-                XCTAssertEqual(error.localizedDescription, "Response requires handling the MFA Webauthn flow but handler given ('needsMultifactor' parameter).")
+                if case .missingMultiFactorHandler = error {
+                    expect.fulfill()
+                } else {
+                    XCTFail("Unexpected error: \(error)")
+                }
```
crazytonyli: All other unit test fixes are adopting the new signature. This one here is the only place where assertion has been changed. Because the `missingMultiFactorHandler` error should never happen in production, it's a programming error and it's not really helpful to show the "require 'needsMultifactor' parameter" message to users.

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -1,15 +1,106 @@
...
-    case invalidClient
-    case unsupportedGrantType
-    case invalidRequest
-    case needsMultifactorCode
-    case invalidOneTimePassword
-    case socialLoginExistingUserUnconnected
-    case invalidTwoStepCode
-    case unknownUser
+/// Errors returned by `WordPressComOAuthClient`.
+public enum WordPressComOAuthError: Error {
```
mokagio: I think this name is fine and also backwards compatible, but the comment made me wonder whether `WordPressComOAuthClientError` would be better as a name? Or even as a nested type `WordPressComOAuthClient.Error`?

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -1,15 +1,106 @@
...
+    /// Can't encode the request arguments into a valid HTTP request. This is a programming error.
+    case requestEncodingFailure
+    /// Error occured in the HTTP connection.
+    case connection(URLError)
+    /// The caller didn't pass a closure to handle MFA authentication request. This is a programming error.
+    case missingMultiFactorHandler
+    /// WordPress.com rejects the authentication request and returns an error. See `AuthenticationFailure` for error details.
```
mokagio:
```suggestion
    /// WordPress.com rejected the authentication request and returned an error. See `AuthenticationFailure` for error details.
```

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -1,15 +1,106 @@
...
+    case authenticationFailure(AuthenticationFailure)
+    /// WordPress.com returns an HTTP response that WordPressKit can't parse.
```
mokagio:
```suggestion
    /// WordPress.com returned an HTTP response that WordPressKit can't parse.
```

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -1,15 +1,106 @@
...
+    case unparsableResponse(response: HTTPURLResponse?, body: Data?)
+    /// Other error occured.
+    case unknown(underlyingError: Error)
+
+    public var authenticationFailureKind: AuthenticationFailure.Kind? {
+        if case let .authenticationFailure(authenticationFailure) = self {
```
mokagio: Oh, interesting. I usually would have written `case .authenticationFailure(let authenticationFailure)` nice to see this works, too.

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -1,15 +1,106 @@
...
+            return authenticationFailure.kind
+        }
+        return nil
```
mokagio: Super nitpick: Have you considered using `guard let` for this early return?

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -1,15 +1,106 @@
...
+            case unsupportedGrantType
+            /// A required field is missing/malformed
+            case invalidRequest
+            /// Multifactor Authentication code is required
+            case needsMultifactorCode
+            /// Supplied MFA code is incorrect
+            case invalidOneTimePassword
+            /// Returned by the social login endpoint if a wpcom user is found, but not connected to a social service.
+            case socialLoginExistingUserUnconnected
+            /// /// Supplied MFA code is incorrect
```
mokagio:
```suggestion
            /// Supplied MFA code is incorrect
```

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -1,15 +1,106 @@
...
+        public var kind: Kind
+        public var localizedErrorMessage: String?
+        public var newNonce: String?
+        public var originalErrorJSON: [String: AnyObject]
+
+        init(apiJSONResponse responseDict: [String: AnyObject]) {
+            originalErrorJSON = responseDict
+
+            // there's either a data object, or an error.
+            if  let errorStr = responseDict["error"] as? String {
```
mokagio: Super nitpick:

```suggestion
            if let errorStr = responseDict["error"] as? String {
```

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -1,15 +1,106 @@
...
+            } else if let data = responseDict["data"] as? [String: AnyObject],
+                let errors = data["errors"] as? NSArray,
+                let err = errors.firstObject as? [String: AnyObject] {
+                let errorCode = err["code"] as? String ?? ""
+                kind = Self.errorsMap[errorCode] ?? .unknown
+                localizedErrorMessage = err["message"] as? String
+                newNonce = data["two_step_nonce"] as? String
+            } else {
+                kind = .unknown
+            }
```
mokagio: Have you considered using `Decodable` support types for this? Or... would it be too convoluted because of hte nested dictionaries?

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -194,20 +281,11 @@ public final class WordPressComOAuthClient: NSObject {
...
-                        failure(
-                            NSError(
-                                domain: WordPressComOAuthClient.WordPressComOAuthErrorDomain,
-                                code: WordPressComOAuthError.unknown.rawValue,
-                                userInfo: [
-                                    NSLocalizedDescriptionKey: "Response requires handling the MFA Webauthn flow but handler given ('needsMultifactor' parameter)."
-                                ]
-                            )
-                        )
+                        failure(WordPressComOAuthError.missingMultiFactorHandler)
```
mokagio: Nice! This also fixes my mistake when writing that description string. It should have been "but **_no_** handler given"

WordPressKitTests/WordPressComOAuthClientTests.swift
mokagio: > Because the missingMultiFactorHandler error should never happen in production, it's a programming error and it's not really helpful to show the "require 'needsMultifactor' parameter" message to users.

This reminds me, I think the multifactor handler is optional because the intent was to not break compatibility, but given this PR already introduces a breaking change, then everything is fair game. I would offer to make the change myself, but I'm doing support rotation this week. That's not to say I won't be able to do that next week, but just in case you are keen and want to jump ahead...

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -125,13 +212,15 @@ public final class WordPressComOAuthClient: NSObject {
...
     @available(*, deprecated, message: "Here for legacy compatiblity. Use the verions with explicit `needsMultifactor` parameter.")
-    @objc public func authenticateWithUsername(_ username: String,
-                                               password: String,
-                                               multifactorCode: String?,
-                                               success: @escaping (_ authToken: String?) -> Void,
-                                               failure: @escaping (_ error: NSError) -> Void ) {
-        authenticateWithUsername(
-            username,
+    public func authenticate(
+        username: String,
```
mokagio: I appreciate this and all the following Swift API syntax improvements.

From the PR description:

>  all the changes I'm about to make to WordPressKit and WordPressAuthenticator will be Swift first. A Swift friendly API is the first priority. If the new API can be bridged to Objective-C, then great. If it can't, I'll either create a Objective-C friendly wrapper or (at the worst case) translate the Objective-C callers to Swift.

100%. It's about time.

mokagio (approved):

WordPressKit/WordPressComOAuthClient.swift
crazytonyli: I prefer its current name without namespace. It helps Swift compiler to distinguish it with `Swift.Error` and potentially other `Error` types under other namespaces if we are giving all error types namespaces.

WordPressKit/WordPressComOAuthClient.swift
crazytonyli: No. I think I wrote these lines on muscle memory, without putting many thoughts into them. 😸

WordPressKit/WordPressComOAuthClient.swift
crazytonyli: I don't think using `Decodable` here would be convoluted. That didn't cross my mind though, because I was blindly making small changes tothe original parsing code. 🙈

WordPressKitTests/WordPressComOAuthClientTests.swift
crazytonyli: It's removed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/650/commits/c1411cd0e7471aa34b15b2e09d942d15d38965d6

PR #653, opened 2023-11-30
Add WordPressAPIError

crazytonyli:
### Description

> [!Note]
> This PR targets the https://github.com/wordpress-mobile/WordPressKit-iOS/pull/650 PR branch.

In https://github.com/wordpress-mobile/WordPressKit-iOS/pull/650, I've refactored [`WordPressComOAuthError` into a Swift error type][Diff-Link]. You'll see there is only one case (out of its five enum cases) that's related to the OAuth domain: `case authenticationFailure(AuthenticationFailure)`. It'd makes sense to further abstract this error type to make it apply to all API endpoints in WordPressKit. This PR did exactly that.

It's not entirely related to this PR, but here is a bit it more context: I have added some helpers around URLSession in [this branch][URLSession-Branch], to make it easier to call in the specific endpoint implementations. This PR's new error is used in those helpers.

### Testing Details

This PR contains only syntactical changes. No extra testing required.

---

[Diff-Link]: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/650/files#diff-10c71f00f75a60cd8268a53f9bd665522ccfd106243f9cf46aedd2b988aaa807R4
[URLSession-Branch]: https://github.com/wordpress-mobile/WordPressKit-iOS/tree/url-session-helpers-bye-bye-alamofire

### DIFF
--- a/WordPressKit/WordPressAPIError.swift
+++ b/WordPressKit/WordPressAPIError.swift
@@ -1,0 +1,43 @@
+import Foundation
+
+public enum WordPressAPIError<EndpointError>: Error where EndpointError: LocalizedError {
+    static var unknownErrorMessage: String {
+        NSLocalizedString(
+            "wordpress-api.error.unknown",
+            value: "Something went wrong, please try again later.",
+            comment: "Error message that describes an unknown error had occured"
+        )
+    }
+
+    /// Can't encode the request arguments into a valid HTTP request. This is a programming error.
+    case requestEncodingFailure
+    /// Error occured in the HTTP connection.
+    case connection(URLError)
+    /// The API call returned an error result. For example, an OAuth endpoint may returns an 'incorrect username or password' error, an upload media endpoint may return an 'unsupported media type' error.
+    case endpointError(EndpointError)
+    /// WordPress.com returned an HTTP response that WordPressKit can't parse.
+    case unparsableResponse(response: HTTPURLResponse?, body: Data?)
+    /// Other error occured.
+    case unknown(underlyingError: Error)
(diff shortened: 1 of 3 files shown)


WordPressKit/WordPressAPIError.swift
```diff
@@ -0,0 +1,43 @@
...
+        )
+    }
+
+    /// Can't encode the request arguments into a valid HTTP request. This is a programming error.
+    case requestEncodingFailure
+    /// Error occured in the HTTP connection.
+    case connection(URLError)
+    /// The API call returned an error result. For example, an OAuth endpoint may returns an 'incorrect username or password' error, an upload media endpoint may return an 'unsupported media type' error.
+    case endpointError(EndpointError)
+    /// WordPress.com returned an HTTP response that WordPressKit can't parse.
```
jkmassel: What about things that aren't WordPress.com? :)

jkmassel (approved):
Looks like a good start – I'm not sure if we need to address my one question yet, because AFAICT this is only scoped to WP.com right now (but the `WordPressAPIError` type might need to expand its scope down the road?)

WordPressKit/WordPressAPIError.swift
crazytonyli: I missed this particular comment, which is incorrect now. Because the new `WordPressAPIError` is for all API errors, not just WP.com ones. Updated in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/653/commits/383490efb89a24820b704b7232cab97796f52d17

PR #654, opened 2023-12-01
Implement building multipart form content

crazytonyli:
### Description

`URLSession` doesn't have an API to build "multipart/form-data" content. This PR implements it ourselves. The new implementation is not used yet, but will be used when I work on replacing Alamofire with URLSession.

I don't think there are many optimizations we can do here, in terms of avoiding copying files. `URLRequest` only accepts `Data` or `InputStream` as its body. I had a though of subclassing `InputStream` to embed many `InputStream` instances internally and build the form-data content one the file when the `read` method is called. But its API is pretty strict: it can only be instantiated using a `Data` or `URL`, which seems to imply the file content behind the stream has to be ready for reading.

### Testing Details

See the unit tests. Let me know if you think there are more tests we can add.

---


WordPressKitTests/Utilities/MultipartFormTests.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    func testPlainText() throws {
+        let af = MultipartFormData()
+        af.append("hello".data(using: .utf8)!, withName: "world")
+        af.append("foo".data(using: .utf8)!, withName: "bar")
+        af.append("the".data(using: .utf8)!, withName: "end")
+        let afEncoded = try af.encode()
+
+        print("# Alamofire")
+        try print(XCTUnwrap(String(data: afEncoded, encoding: .utf8)))
```
mokagio: I suppose we can remove these `print` statements before merging, right?

```suggestion
```

WordPressKit/MultipartForm.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    init(data: Data, name: String, filename: String? = nil, mimeType: String? = nil) {
+        self.inputStream = InputStream(data: data)
+        self.name = name
+        self.filename = filename
+        self.bytes = UInt64(data.count)
+        self.mimeType = mimeType
+    }
+
+    init(fileAtPath path: String, name: String, filename: String? = nil, mimeType: String? = nil) throws {
```
mokagio, line 28: Have you considered removing `filename` and forcing callers to specify the whole path with `fileAtPath`?

mokagio (dismissed):

WordPressKit/MultipartForm.swift
crazytonyli: These two things are related but also should co-exist. The multipart form contains a `filename` field, whose value will be set by the `filename` argument here. We can use the file name of the  `fileAtPath` as the default value. But if some code passes a `filename` argument in, we should use it instead.

WordPressKitTests/Utilities/MultipartFormTests.swift
crazytonyli: Good call. They are gone in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/654/commits/cab9ddf1046d83aa3d2045a89279f9ff6e352cc4.

mokagio (approved):

mokagio:
Approved after @crazytonyli replied to my question about input parameters https://github.com/wordpress-mobile/WordPressKit-iOS/pull/654#discussion_r1446792796

PR #657, opened 2023-12-11
Add HTTPRequestBuilder

crazytonyli:
### Description

This new type will be used to construct `URLRequest`s for all HTTP API requests.

### Testing Details

See the unit tests. Let me know if you think there are more should be added.

---

### DIFF
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -104,7 +104,7 @@ final class HTTPRequestBuilder {
     }
 
     func build() throws -> URLRequest {
-        if path == "" || path.hasPrefix("/") {
+        if path.isEmpty || path.hasPrefix("/") {
             urlComponents.path = path
         } else {
             var newPath = urlComponents.path
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -15,7 +15,6 @@ final class HTTPRequestBuilder {
 
     private var urlComponents: URLComponents
     private var method: Method = .get
-    private var path: String
     private var headers: [String: String] = [:]
     private var bodyBuilder: ((inout URLRequest) throws -> Void)?
 
@@ -24,7 +23,6 @@ final class HTTPRequestBuilder {
         assert(url.host != nil)
 
         urlComponents = URLComponents(url: url, resolvingAgainstBaseURL: true)!
-        path = urlComponents.path
     }
 
     func method(_ method: Method) -> Self {
@@ -32,10 +30,20 @@ final class HTTPRequestBuilder {
         return self
     }
 
-    func path(_ path: String) -> Self {
+    func append(path: String) -> Self {
         assert(!path.contains("?") && !path.contains("#"), "Path should not have query or fragment: \(path)")
 
-        self.path = path
+        var relPath = path
+        if relPath.hasPrefix("/") {
+            _ = relPath.removeFirst()
+        }
+
+        if urlComponents.path.hasSuffix("/") {
+            urlComponents.path = urlComponents.path.appending(relPath)
+        } else {
+            urlComponents.path = urlComponents.path.appending("/").appending(relPath)
+        }
+
         return self
     }
 
@@ -92,22 +100,6 @@ final class HTTPRequestBuilder {
     }
 
     func build() throws -> URLRequest {
-        if path.isEmpty || path.hasPrefix("/") {
-            urlComponents.path = path
-        } else {
-            var newPath = urlComponents.path
-            if !newPath.hasPrefix("/") {
-                newPath = "/" + newPath
-            }
-
-            if newPath.hasSuffix("/") {
-                newPath = newPath + path
-            } else {
-                newPath = newPath + "/" + path
-            }
-            urlComponents.path = newPath
-        }
-
         guard let url = urlComponents.url else {
             throw URLError(.badURL)
         }
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -12,6 +12,7 @@ final class HTTPRequestBuilder {
             self == .post || self == .put || self == .patch
         }
     }
+
     private var urlComponents: URLComponents
     private var method: Method = .get
     private var path: String
@@ -26,20 +27,20 @@ final class HTTPRequestBuilder {
         path = urlComponents.path
     }
 
-    func set(method: Method) -> Self {
+    func method(_ method: Method) -> Self {
         self.method = method
         return self
     }
 
-    func set(path: String) -> Self {
+    func path(_ path: String) -> Self {
         assert(!path.contains("?") && !path.contains("#"), "Path should not have query or fragment: \(path)")
 
         self.path = path
         return self
     }
 
-    func set(value: String?, forHeader header: String) -> Self {
-        headers[header] = value
+    func header(name: String, value: String?) -> Self {
+        headers[name] = value
         return self
     }
 
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -83,18 +83,6 @@ final class HTTPRequestBuilder {
         return self
     }
 
-    // FIXME: Not implemented yet
-    func body(xmlrpc: Any /* XMLRPCRequest */) -> Self {
-        body(xml: {
-            fatalError("To be implemented")
-        })
-    }
-
-    // FIXME: Not implemented yet
-    func appendXMLRPCArgument(value: Any) -> Self {
-        fatalError("To be implemented")
-    }
-
     func body(xml: @escaping () throws -> Data) -> Self {
         headers["Content-Type"] = "application/xml; charset=utf-8"
         bodyBuilder = { req in
@@ -137,5 +125,19 @@ final class HTTPRequestBuilder {
         }
 
         return request
+    }
+}
+
+extension HTTPRequestBuilder {
+    // FIXME: Not implemented yet
+    func body(xmlrpc: Any /* XMLRPCRequest */) -> Self {
+        body(xml: {
+            fatalError("To be implemented")
+        })
+    }
+
+    // FIXME: Not implemented yet
+    func appendXMLRPCArgument(value: Any) -> Self {
+        fatalError("To be implemented")
     }
 }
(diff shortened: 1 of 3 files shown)


WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -0,0 +1,140 @@
...
+        case post = "POST"
+        case put = "PUT"
+        case patch = "PATCH"
+        case delete = "DELETE"
+
+        var allowsHTTPBody: Bool {
+            self == .post || self == .put || self == .patch
+        }
+    }
+    private var urlComponents: URLComponents
```
mokagio:
```suggestion
    }

    private var urlComponents: URLComponents
```

🤓

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -0,0 +1,140 @@
...
+        self.method = method
+        return self
+    }
+
+    func set(path: String) -> Self {
+        assert(!path.contains("?") && !path.contains("#"), "Path should not have query or fragment: \(path)")
+
+        self.path = path
+        return self
+    }
```
mokagio: How do you feel about creating a type to encapsulate this?

```swift
extension URL {

  struct Path {

      let value: String

      /// `Path` cannot not contain URL query fragments (? and #)
      init(from string: String) {
	assert(!string.contains("?") && !string.contains("#"), "Path should not have query or fragment: \(path)")
        value = string
      }
```

The end result is the same: The app will crash if the path is invalid, but I feel explicitly wrapping it in a dedicated type will make what to do more obvious to consumers.

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -0,0 +1,140 @@
...
+
+    func set(value: String?, forHeader header: String) -> Self {
+        headers[header] = value
+        return self
+    }
+
+    func query(name: String, value: String?, override: Bool = false) -> Self {
+        append(query: [URLQueryItem(name: name, value: value)], override: override)
+    }
```
mokagio, line 57: All the methods before this use "set" as their verb. 

I think if it would be good to use it consistently, to have a homogeneous usage at call site. E.g. from

```swift
var request = try HTTPRequestBuilder(url: URL(string: "https://wordpress.org/")!)
            .set(method: .post)
            .body(json: 42)
```

to

```swift
var request = try HTTPRequestBuilder(url: URL(string: "https://wordpress.org/")!)
            .set(method: .post)
            .set(jsonBody: 42)
```

or

```swift
var request = try HTTPRequestBuilder(url: URL(string: "https://wordpress.org/")!)
            .setMethod(.post)
            .setJSON(body: 42)
```

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -0,0 +1,140 @@
...
+
+    func append(query: [URLQueryItem], override: Bool) -> Self {
+        urlComponents.queryItems = (urlComponents.queryItems ?? []) + query
+        return self
+    }
+
+    func body(form: [String: String]) -> Self {
+        headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
+        bodyBuilder = { req in
+            var url = URLComponents(string: "https://wordpress.com")!
```
mokagio, line 67: Should this be 

```suggestion
            var url = self.urlComponents
```

I'm surprised `https://wordpress.com/` is hardcoded here but maybe I'm missing something about the intended use? E.g., only requests to the WordPress.com API can send URL form encoded requests?

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -0,0 +1,140 @@
...
+    func body(xmlrpc: Any /* XMLRPCRequest */) -> Self {
+        body(xml: {
+            fatalError("To be implemented")
+        })
+    }
+
+    // FIXME: Not implemented yet
+    func appendXMLRPCArgument(value: Any) -> Self {
+        fatalError("To be implemented")
+    }
```
mokagio: Much of a muchness, but what do you think of moving these specific methods into an extension? All I've seen so far in `HTTPRequestBuilder` is standard HTTP. These are the only methods that implement [an additional spec](http://xmlrpc.com/).

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -0,0 +1,140 @@
...
+    func body(xml: @escaping () throws -> Data) -> Self {
+        headers["Content-Type"] = "application/xml; charset=utf-8"
+        bodyBuilder = { req in
+            req.httpBody = try xml()
+        }
+        return self
+    }
+
+    func build() throws -> URLRequest {
+        if path == "" || path.hasPrefix("/") {
```
mokagio: Nitpick

```suggestion
        if path.isEmpty || path.hasPrefix("/") {
```

WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
```diff
@@ -0,0 +1,93 @@
...
+
+        request = try HTTPRequestBuilder(url: URL(string: "https://wordpress.org/hello")!)
+            .set(path: "world")
+            .build()
+        XCTAssertEqual(request.url?.absoluteString, "https://wordpress.org/hello/world")
+
+        request = try HTTPRequestBuilder(url: URL(string: "https://wordpress.org/hello")!)
+            .set(path: "/world")
+            .build()
+        XCTAssertEqual(request.url?.absoluteString, "https://wordpress.org/world")
```
mokagio: This behavior, which I read as "when the given `path` is absolute, override the original URL path" is surprising.

Or maybe is just me applying the expectation from `NSString` `appendingPathComponent`:

```
# from the REPL
  2> let s: NSString = "https://foo.test/bar"
s: NSMutableString = "https://foo.test/bar"
  3> s.appendingPathComponent("/baz")
$R0: String = "https:/foo.test/bar/baz"
  4> s.appendingPathComponent("bang")
$R1: String = "https:/foo.test/bar/bang"
  5>
```

I acknowledge that `NSString` explicitly says "appending", though, whereas this API gives no similar clear guardrail on what `path` would do.

WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
```diff
@@ -0,0 +1,93 @@
...
+
+@testable import WordPressKit
+
+class HTTPRequestBuilderTests: XCTestCase {
+
+    func testDefaultMethod() throws {
+        let request = try HTTPRequestBuilder(url: URL(string: "https://wordpress.org")!).build()
+        XCTAssertEqual(request.url?.absoluteString, "https://wordpress.org")
+        XCTAssertEqual(request.httpMethod, "GET")
+    }
```
mokagio: This test made me want to suggest to triangulate setting the URL, that is add at least two different examples (cc @wordpress-mobile/mobile-ui-testing-squad)

```swift
XCTAssertEqual(
  try HTTPRequestBuilder(url: URL(string: "https://wordpress.org/")!).build().url?.absoluteString,
  "https://wordpress.org/"
)
XCTAssertEqual(
  try HTTPRequestBuilder(url: URL(string: "https://wordpress.com/")!).build().url?.absoluteString,
  "https://wordpress.com/"
)
```

The value of this approach is that it gives additional confidence in the robustness of the tests. **But that's mostly redundant when we have more tests below that implicitly exercise this behavior.**

However, something I noticed and that we could use to add a variety of initial `URL` inputs is that there are no tests for how `set(method:)` behaves. Granted, that's pretty straightforward, but given how core this object will be, I think "overtesting" it wouldn't be a waste. The same goes for the `HTTPRequestBuilder.Method` `allowsHTTPBody` logic.

WordPressKit/HTTPRequestBuilder.swift
crazytonyli: This function needs a URL encoded string. I'm just using a temporary `URLComponents` to construct a query which is url encoded.

WordPressKit/HTTPRequestBuilder.swift
crazytonyli: Good catch! I took the other direction: removing the "set" part, because it's redundant considering this is a "builder" type. See https://github.com/wordpress-mobile/WordPressKit-iOS/pull/657/commits/b52e98d572230e251ee70c2d997611b18df1997d.

WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
crazytonyli: After a second look, I do agree with you that this behaviour is not very intuitive.

I have updated the API to only allow appending path. See https://github.com/wordpress-mobile/WordPressKit-iOS/pull/657/commits/4bb836093c79c8ee7c897afe42ded12032766f01

WordPressKit/HTTPRequestBuilder.swift
crazytonyli: I see value in this dedicated type, if we have more places that need to refer to a "URL path". But for this one method in `HTTPRequestBuilder`, I don't feel like it's worth the effort?

mokagio: Yes, fair enough.

WordPressKit/HTTPRequestBuilder.swift
mokagio: Fine by me 😄

WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
```diff
@@ -5,10 +5,26 @@ import XCTest
...
+        XCTAssertFalse(HTTPRequestBuilder.Method.delete.allowsHTTPBody)
+
+        try XCTAssertEqual(HTTPRequestBuilder(url: URL(string: "https://wordpress.org")!).method(.post).build().httpMethod, "POST")
+        XCTAssertTrue(HTTPRequestBuilder.Method.post.allowsHTTPBody)
+
+        try XCTAssertEqual(HTTPRequestBuilder(url: URL(string: "https://wordpress.org")!).method(.patch).build().httpMethod, "PATCH")
+        XCTAssertTrue(HTTPRequestBuilder.Method.patch.allowsHTTPBody)
+
+        try XCTAssertEqual(HTTPRequestBuilder(url: URL(string: "https://wordpress.org")!).method(.put).build().httpMethod, "PUT")
+        XCTAssertTrue(HTTPRequestBuilder.Method.put.allowsHTTPBody)
```
mokagio, line 27: Nice touch grouping the tests this way.

I don't subscribe to the "_one assertion per test_" rule that some tests enthusiast argue for. This kind of functional/semantic grouping strikes a good balance between not repeating the `func test...` cruft and having tests that point to the same behavior.

mokagio (approved):
Thank you for taking my suggestions into consideration.

PR #658, opened 2023-12-11
Add URLSession helpers to send requests and handle responses

crazytonyli:
### Description

> [!NOTE]
> This PR is built on top of #657.

This PR adds two simple helpers that will be used to send API requests and parse the responses.

#### Send requests

`URLSession.apiResult(with:errorType:)` sends a HTTP request and returns the result as `Swift.Result` type.

If the HTTP request has got a response, a "Success" result will be returned, regardless the HTTP response status code.

For all other cases where we can't receive a response, for example, developer passes an incorrect instance as JSON body where it can be encoded as a JSON, or internet connection issue, a "Failure" result will be returned. The failure type is [WordPressAPIError](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/0b17ca5119b63e014a9d60a6860df4bb985e94cb/WordPressKit/WordPressAPIError.swift#L3).

#### Parse responses

`Result.assessStatusCode(acceptable:success:failure:)` can be used to convert a "Success" HTTP connection response into success or failure result based on the response status code. By default, "[200, 299]" is the acceptable status codes.

### Testing Details

See the added unit tests. Let me know if you think there are more should be added.

---

### DIFF
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -2,7 +2,7 @@ import Foundation
 
 public typealias WordPressAPIResult<R, E: LocalizedError> = Result<R, WordPressAPIError<E>>
 
-struct HTTPAPIResponse<Body>{
+struct HTTPAPIResponse<Body> {
     typealias Body = Body
 
     var response: HTTPURLResponse
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -1,18 +1,16 @@
 import Foundation
 
-public typealias WordPressAPIResult<R, E: LocalizedError> = Result<R, WordPressAPIError<E>>
+public typealias WordPressAPIResult<Response, Error: LocalizedError> = Result<Response, WordPressAPIError<Error>>
 
 struct HTTPAPIResponse<Body> {
-    typealias Body = Body
-
     var response: HTTPURLResponse
     var body: Body
 }
 
 extension URLSession {
 
-    func apiResult<E: LocalizedError>(
-        with builder: HTTPRequestBuilder,
+    func perform<E: LocalizedError>(
+        request builder: HTTPRequestBuilder,
         errorType: E.Type = E.self
     ) async -> WordPressAPIResult<HTTPAPIResponse<Data>, E> {
         guard let request = try? builder.build() else {
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -1,0 +1,68 @@
+import Foundation
+
+public typealias WordPressAPIResult<R, E: LocalizedError> = Result<R, WordPressAPIError<E>>
+
+struct HTTPAPIResponse<Body>{
+    typealias Body = Body
+
+    var response: HTTPURLResponse
+    var body: Body
+}
+
+extension URLSession {
+
+    func apiResult<E: LocalizedError>(
+        with builder: HTTPRequestBuilder,
+        errorType: E.Type = E.self
+    ) async -> WordPressAPIResult<HTTPAPIResponse<Data>, E> {
+        guard let request = try? builder.build() else {
+            return .failure(.requestEncodingFailure)
+        }
+
+        let result: (Data, URLResponse)
+        do {
+            result = try await data(for: request)
+        } catch {
+            if let urlError = error as? URLError {
+                return .failure(.connection(urlError))
+            } else {
+                return .failure(.unknown(underlyingError: error))
+            }
+        }
+
+        let (body, response) = result
+
+        guard let response = response as? HTTPURLResponse else {
+            return .failure(.unparsableResponse(response: nil, body: body))
+        }
+
+        return .success(.init(response: response, body: body))
+    }
+
+}
+
+extension Result where Success == HTTPAPIResponse<Data> {
+
+    func assessStatusCode<S, E: LocalizedError>(
+        acceptable: [ClosedRange<Int>] = [200...299],
+        success: (Success) -> S?,
+        failure: (Success) -> E?
+    ) -> WordPressAPIResult<S, E> where Failure == WordPressAPIError<E> {
+        flatMap { response in
+            if acceptable.contains(where: { $0 ~= response.response.statusCode }) {
+                if let result = success(response) {
+                    return .success(result)
+                } else {
+                    return .failure(.unparsableResponse(response: response.response, body: response.body))
+                }
+            } else {
+                if let endpointError = failure(response) {
+                    return .failure(.endpointError(endpointError))
+                } else {
+                    return .failure(.unparsableResponse(response: response.response, body: response.body))
+                }
+            }
+        }
+    }
+
+}
--- a/WordPressKitTests/Utilities/URLSessionHelperTests.swift
+++ b/WordPressKitTests/Utilities/URLSessionHelperTests.swift
@@ -1,0 +1,129 @@
+import Foundation
+import XCTest
+import OHHTTPStubs
+
+@testable import WordPressKit
+
+class URLSessionHelperTests: XCTestCase {
+
+    override func tearDown() {
+        HTTPStubs.removeAllStubs()
+    }
+
+    func testConnectionError() async throws {
+        stub(condition: isPath("/hello")) { _ in
+            HTTPStubsResponse(error: URLError(.serverCertificateUntrusted))
+        }
+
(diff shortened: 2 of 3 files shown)


WordPressKit/HTTPClient.swift
```diff
@@ -0,0 +1,68 @@
+import Foundation
+
+public typealias WordPressAPIResult<R, E: LocalizedError> = Result<R, WordPressAPIError<E>>
```
mokagio: Nitpick. I've seen Swift generics use more verbose labels, e.g. [`Result<Success, Failure>`](https://developer.apple.com/documentation/swift/result)

```suggestion
public typealias WordPressAPIResult<Response, Error: LocalizedError> = Result<Response, WordPressAPIError<Error>>
```

Not sure if `Error` will work or get confused with `Swift.Error`. If it doesn't work, then maybe `Failure`?

WordPressKit/HTTPClient.swift
```diff
@@ -0,0 +1,68 @@
...
+
+struct HTTPAPIResponse<Body> {
+    typealias Body = Body
```
mokagio: Is this `typealias` necessary?

WordPressKit/HTTPClient.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    var response: HTTPURLResponse
+    var body: Body
+}
+
+extension URLSession {
+
+    func apiResult<E: LocalizedError>(
+        with builder: HTTPRequestBuilder,
+        errorType: E.Type = E.self
```
mokagio, line 14: `errorType` looks unused?

image

mokagio: I'll keep rolling with the unused assumption and make an additional suggestion, which would be valid anyway but results in my opinion cleaner this way. 

What do you think of renaming the method to a verb like "perform"? The fact that the method returns an API result is already encoded in the type signature.

This way, including the removed error type, we could go from a usage like:

```swift
let result = await URLSession.shared.apiResult(with: .init(url: URL(string: "https://wordpress.org/hello")!), errorType: TestError.self)
```

to:

```swift
let result = await URLSession.shared.perform(.init(url: URL(string: "https://wordpress.org/hello")!)
```

or, if I misunderstood the error type requirement, 

```swift
let result = await URLSession.shared.perform(.init(url: URL(string: "https://wordpress.org/hello")!), errorType: TestError.self)
```

An alternative to `perform` could also be `request`

mokagio (commented):
Just a few comments. I haven't looked at the test well yet.

WordPressKit/HTTPClient.swift
crazytonyli: > errorType looks unused?

Yes. But that's intended. We don't really need it in the implementation. It's similar to `JSONEncoder.encode` where callers can use it to tell Swift compiler what the `E` type is instead of letting the compiler infers it based on the context. You can see examples of this usage in the unit tests.

> What do you think of renaming the method to a verb like "perform"?

Sounds good to me! 👍

WordPressKitTests/Utilities/URLSessionHelperTests.swift
```diff
@@ -0,0 +1,130 @@
...
+        _ = result
+            .assessStatusCode { result in
+                XCTAssertEqual(String(data: result.body, encoding: .utf8), "success")
+                expectation.fulfill()
+                return result
+            } failure: { _ in
+                // Do nothing
+                return nil
+            }
+        await fulfillment(of: [expectation])
```
mokagio, line 50: Have you considered making the code synchronous?

```suggestion
        _ = result
            .assessStatusCode { result in
                XCTAssertEqual(String(data: result.body, encoding: .utf8), "success")
                return result
            } failure: { _ in
                XCTFail("Expected result to be successful and this branch not to run")
                return nil
            }
```

WordPressKitTests/Utilities/URLSessionHelperTests.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    func testUnacceptable500() async throws {
+        stub(condition: isPath("/hello")) { _ in
+            HTTPStubsResponse(data: "Internal server error".data(using: .utf8)!, statusCode: 500, headers: nil)
+        }
+
+        let result = await URLSession.shared.perform(request: .init(url: URL(string: "https://wordpress.org/hello")!), errorType: TestError.self)
+
+        // The result is a successful result. This line should not throw
+        _ = try result.get()
```
mokagio, line 61: I was surprised for a moment. "The result is a successful result," but the stub was configured to "fail" the request. The difference, however, is that the request _succeeded to get a response_ and that _the response contained a failure_.

I wonder if there's a way to make this clearer from consumers. Or maybe I'm overthinking it...

WordPressKitTests/Utilities/URLSessionHelperTests.swift
```diff
@@ -0,0 +1,130 @@
...
+            .assessStatusCode { result in
+                return result
+            } failure: { result in
+                expectation.fulfill()
+                if result.response.statusCode == 404 {
+                    return .postNotFound
+                }
+                return nil
+            }
+        await fulfillment(of: [expectation])
```
mokagio, line 118: If I read this correctly based also on the info in the PR description, these tests are previews of how we'd use the code in real life to make API requests.

With that in mind, what do you think of using a name like map?

```swift
let session = ...

let result = await session.perform(request: .init(url: URL(string: "https://wordpress.org/hello")!), errorType: TestError.self)
   .map(
     acceptableStatusCodes: [200...299],
     success: { result in ... }
     failure: { result in ... }
  )
```

With default parameter:

```swift
let session = ...

let result = await session.perform(request: .init(url: URL(string: "https://wordpress.org/hello")!), errorType: TestError.self)
   .map(
     success: { result in ... }
     failure: { result in ... }
  )
```

I'm suggesting this because "assess" doesn't make me think of an operation that would apply the given transformations to `self` and return the result. 🤷‍♂️

WordPressKit/HTTPClient.swift
mokagio: > [`errorType` works] similar to JSONEncoder.encode where callers can use it to tell Swift compiler what the E type is instead of letting the compiler infers it based on the context.

Gotcha. If we don't specify it, then we're left to explicitly set the type of the result we expect from the API call, as can be seen in this edit I made locally.

image

To be honest, I'd rather to that than pass unused arguments, but I can see how the current version is easier on the eyes. Also, as you point out, `JSONEncoder` and `JSONDecoder` set a precedent for this style, which might then result more familiar to Swift devs.

mokagio (approved):
Just a few naming notes, nothing blocking.

Going forward, do you plan on adding a `Result` `HTTPAPIResponse<Data> ` add on similar to `assess...` that can automatically try to transform the `Data` into a `Decodable` object?

I once [played around with an API abstraction](https://github.com/mokagio/MinimalNetworking/blob/cf66d7e47f195b88e50f99a827f3f90b75c877d7/Tests/MinimalNetworkingTests/MinimalNetworkingTests.swift) that allowed describing API endpoints as dedicated objects as long as the expected resource in the app domain conformed to `Decodable`.

WordPressKitTests/Utilities/URLSessionHelperTests.swift
crazytonyli: If neither closures of `assessStatusCode` is called, the updated test would pass. But the original test case would fail and catch the bug.

WordPressKitTests/Utilities/URLSessionHelperTests.swift
crazytonyli: I had the same thought too. On one hand, what's done here follows `URLSession`'s API, where it doesn't take HTTP semantic into account. On the other hand, I do want to make these helpers somewhat opinionated: locking them down to suit for our cases rather than making it too general like `URLSession`.

We can add an "acceptable status codes" argument to the `perform(request:)` function. That means we'll need to add another error case `WordPressAPIError.unaccpetableStatusCode(HTTPAPIResponse<Data>)`. The helper that deals with success and error parsing (parsing generic JSON to specific model, parsing http error response to a concrete error type) will need to change too.

I think these changes may potentially be implemented later on. These helpers are internal API after they, they can be easily changed without much affects to others.

At the moment, I only looked into the code in `WordPressOAuthClient.swift`, and the helpers are all created based on the existing implementation. That's why I haven't put too much though into supporting `Decodable`, which I think is very likely going to happen once I start looking into the existing WP.com REST API implementation.

PR #660, opened 2023-12-13
Use URLSession in WordPressOAuthClient

crazytonyli:
### Description

This PR replaces some usages of Alamofire in `WordPressOAuthClient`, with `URLSession`. There will be more similar PRs later for other usages.

I noticed there are many `Alamofire.SessionManager` instances are used. I don't understand why, but at the same time I'm too afraid to change it. For now, I'll follow the existing pattern and continue using many `URLSession` instances too.

### Testing Details

The refactored features are: login with WP.com username password and request code via SMS. You can use https://github.com/wordpress-mobile/WordPress-iOS/pull/22215 to test this change. Here are some test scenarios I can think of:
1. Login with username and password. The app presents 2FA code screen if the account enables 2FA.
2. The app shows appropriate error messages if password is incorrect, or 2fa code is incorrect.
3. Request one time code via SMS works.

---

### DIFF
--- a/WordPressKit/WordPressComOAuthClient.swift
+++ b/WordPressKit/WordPressComOAuthClient.swift
@@ -689,3 +689,16 @@ extension WordPressComOAuthClient {
         }
     }
 }
+
+private extension WordPressComOAuthClient {
+    static func processError(_ response: HTTPAPIResponse<Data>) -> AuthenticationFailure? {
+        guard [400, 409, 403].contains(response.response.statusCode),
+              let responseObject = try? JSONSerialization.jsonObject(with: response.body, options: .allowFragments),
+              let responseDictionary = responseObject as? [String: AnyObject]
+        else {
+            return nil
+        }
+
+        return .init(apiJSONResponse: responseDictionary)
+    }
+}
--- a/WordPressKit/WordPressComOAuthClient.swift
+++ b/WordPressKit/WordPressComOAuthClient.swift
@@ -102,10 +102,6 @@ public final class WordPressComOAuthClient: NSObject {
         WordPressComOAuthClient.urlSession()
     }()
 
-    private let oauth2SessionManager: SessionManager = {
-        return WordPressComOAuthClient.sessionManager()
-    }()
-
     private let webauthnSessionManager: SessionManager = {
         return WordPressComOAuthClient.sessionManager()
     }()
--- a/WordPressKit/WordPressComOAuthClient.swift
+++ b/WordPressKit/WordPressComOAuthClient.swift
@@ -85,8 +85,8 @@ public final class WordPressComOAuthClient: NSObject {
         case socialLogin2FA = "/wp-login.php?action=two-step-authentication-endpoint&version=1.0"
         case socialLoginNewSMS2FA = "/wp-login.php?action=send-sms-code-endpoint"
 
-        func url(base: String) -> URL {
-            return URL(string: self.rawValue, relativeTo: URL(string: base))!
+        func url(base: URL) -> URL {
+            return URL(string: self.rawValue, relativeTo: base)!
         }
     }
 
@@ -95,8 +95,8 @@ public final class WordPressComOAuthClient: NSObject {
     private let clientID: String
     private let secret: String
 
-    private let wordPressComBaseUrl: String
-    private let wordPressComApiBaseUrl: String
+    private let wordPressComBaseUrl: URL
+    private let wordPressComApiBaseUrl: URL
 
     private let oauth2SessionManager: SessionManager = {
         return WordPressComOAuthClient.sessionManager()
@@ -160,8 +160,8 @@ public final class WordPressComOAuthClient: NSObject {
                       wordPressComApiBaseUrl: String = WordPressComOAuthClient.WordPressComOAuthDefaultApiBaseUrl) {
         self.clientID = clientID
         self.secret = secret
-        self.wordPressComBaseUrl = wordPressComBaseUrl
-        self.wordPressComApiBaseUrl = wordPressComApiBaseUrl
+        self.wordPressComBaseUrl = URL(string: wordPressComBaseUrl)!
+        self.wordPressComApiBaseUrl = URL(string: wordPressComApiBaseUrl)!
     }
 
     /// Authenticates on WordPress.com using the OAuth endpoints.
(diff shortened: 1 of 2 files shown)


WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -78,15 +78,14 @@ public final class WordPressComOAuthClient: NSObject {
...
-        case oAuthBase = "/oauth2/token"
         case webauthnChallenge = "wp-login.php?action=webauthn-challenge-endpoint"
         case webauthnAuthentication = "wp-login.php?action=webauthn-authentication-endpoint"
         case socialLogin = "/wp-login.php?action=social-login-endpoint&version=1.0"
         case socialLogin2FA = "/wp-login.php?action=two-step-authentication-endpoint&version=1.0"
         case socialLoginNewSMS2FA = "/wp-login.php?action=send-sms-code-endpoint"
 
-        func url(base: String) -> URL {
-            return URL(string: self.rawValue, relativeTo: URL(string: base))!
+        func url(base: URL) -> URL {
```
mokagio, line 98: 👍 for requiring `base` to be a `URL` and pushing the responsibility to create good URLs to the callers. 

IMHO, in the ideal world, we'd only have one point in the whole app / library where we go from `String` to `URL`. The rest of the code deals with "valid" instances, without having to worry about validation.

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -95,11 +94,12 @@ public final class WordPressComOAuthClient: NSObject {
...
     private let secret: String
 
-    private let wordPressComBaseUrl: String
-    private let wordPressComApiBaseUrl: String
+    private let wordPressComBaseUrl: URL
+    private let wordPressComApiBaseUrl: URL
 
-    private let oauth2SessionManager: SessionManager = {
-        return WordPressComOAuthClient.sessionManager()
+    // Question: Is it necessary to use these many URLSession instances?
```
mokagio, line 111: What do you think of making this a `TODO` or `FIXME` just so tooling can pick it up?

mokagio (commented):
Just a few comments... I didn't finish the review but wanted to avoid losing these.

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -160,8 +166,13 @@ public final class WordPressComOAuthClient: NSObject {
                       wordPressComApiBaseUrl: String = WordPressComOAuthClient.WordPressComOAuthDefaultApiBaseUrl) {
         self.clientID = clientID
         self.secret = secret
-        self.wordPressComBaseUrl = wordPressComBaseUrl
-        self.wordPressComApiBaseUrl = wordPressComApiBaseUrl
+        self.wordPressComBaseUrl = URL(string: wordPressComBaseUrl)!
+        self.wordPressComApiBaseUrl = URL(string: wordPressComApiBaseUrl)!
```
mokagio, line 181: If we plan on more breaking changes, we could also slot one in to require those `String` parameters to be `URL`s.

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -160,8 +166,13 @@ public final class WordPressComOAuthClient: NSObject {
...
+    }
+
+    public enum AuthenticationResult {
+        case authenticated(token: String)
+        case needsMultiFactor(userID: Int, nonceInfo: SocialLogin2FANonceInfo)
     }
```
mokagio, line 187: Nice

crazytonyli:
Need to rebase this PR after https://github.com/wordpress-mobile/WordPressKit-iOS/pull/668 is merged.

WordPressKit/WordPressComOAuthClient.swift
mokagio: Or... have you discovered an answer since writing that comment with the extra work you've done on this library?

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -170,64 +181,88 @@ public final class WordPressComOAuthClient: NSObject {
...
 
-                case .failure(let error):
-                    let nserror = self.processError(response: response, originalError: error)
-                    WPKitLogError("Error receiving OAuth2 token: \(nserror)")
-                    failure(nserror)
-                }
-            })
+                    return .needsMultiFactor(userID: userID, nonceInfo: nonceInfo)
+                },
+                failure: Self.processError(_:)
```
mokagio: Nitpick: I find calling `Self.` clunky, but that might just be me. 

Just for fun, we could

- Make `processError` a `private` free function; or
- It could be a method on `HTTPAPIResponse<Data>`; or
- An `init?` on `AuthenticationFailure`

But... 🤷‍♂️ This is internal code that's only used here.

WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -689,3 +733,22 @@ extension WordPressComOAuthClient {
         }
     }
 }
+
+private extension WordPressComOAuthClient {
+    func tokenRequestBuilder() -> HTTPRequestBuilder {
+        HTTPRequestBuilder(url: wordPressComApiBaseUrl)
+            .method(.post)
+            .append(path: "/oauth2/token")
+    }
```
mokagio, line 752: Love to see these builders in action and to DRY things.

mokagio (approved):

WordPressKit/WordPressComOAuthClient.swift
crazytonyli: I didn't use `TODO` or `FIXME` because I'm not sure it needs a change.

But the pattern sure raises a question, at least to me. Because it's pretty rare to see code that uses different URLSession instances to send requests to endpoints that are in the same area (in this case, authentication). But considering the "endpoints" here are all `php` pages, rather than the common REST style API endpoints, maybe it wants to avoid sharing cookies among those API calls?

WordPressKit/WordPressComOAuthClient.swift
crazytonyli: Good call! I've changed it to `AuthenticationFailure` initialiser in ~https://github.com/wordpress-mobile/WordPressKit-iOS/pull/660/commits/a243bb859c4072c6af310b69632910df030577c0~ https://github.com/wordpress-mobile/WordPressKit-iOS/pull/660/commits/a83f361719e570ae14632aa8c986d355181c7d8c

PR #662, opened 2023-12-13
Fix incorrect error message implementation in WordPressAPIError

crazytonyli:
### Description

I made two mistakes in the existing implementation:
1. A copy-paste error where I should have made `WordPressAPIError` conforms to `LocalizedError`.
2. The implementation should always return a non-nil error message, but I missed that in the `endpointError` case, because `errorDescription` is `String?`.

### Testing Details

I have added unit tests in this PR to verify the fixes.

---

### DIFF
--- a/WordPressKit/WordPressAPIError.swift
+++ b/WordPressKit/WordPressAPIError.swift
@@ -24,20 +24,27 @@ public enum WordPressAPIError<EndpointError>: Error where EndpointError: Localiz
 extension WordPressAPIError: LocalizedError {
 
     public var errorDescription: String? {
+        // Considering `WordPressAPIError` is the error that's surfaced from this library to the apps, its instanes
+        // may be displayed on UI directly. To prevent Swift's default error message (i.e. "This operation can't be
+        // completed. <SwiftTypeName> (code=...)") from being displayed, we need to make sure this implementation
+        // always returns a non-nil value.
+        let localizedErrorMessage: String
         switch self {
         case .requestEncodingFailure, .unparsableResponse:
             // These are usually programming errors.
-            return Self.unknownErrorMessage
+            localizedErrorMessage = Self.unknownErrorMessage
         case let .endpointError(error):
-            return error.errorDescription ?? Self.unknownErrorMessage
+            localizedErrorMessage = error.errorDescription ?? Self.unknownErrorMessage
         case let .connection(error):
-            return error.localizedDescription
+            localizedErrorMessage = error.localizedDescription
         case let .unknown(underlyingError):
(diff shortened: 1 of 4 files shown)


WordPressKit/WordPressAPIError.swift
```diff
@@ -21,15 +21,15 @@ public enum WordPressAPIError<EndpointError>: Error where EndpointError: Localiz
...
+extension WordPressAPIError: LocalizedError {
 
     public var errorDescription: String? {
         switch self {
         case .requestEncodingFailure, .unparsableResponse:
             // These are usually programming errors.
             return Self.unknownErrorMessage
         case let .endpointError(error):
-            return error.errorDescription
+            return error.errorDescription ?? Self.unknownErrorMessage
```
mokagio: > The implementation should always return a non-nil error message

Gotcha. So even though `errorDescription` at the `LocalizedError` level allows for a `nil` value, we agree not to do that.

Would it be worth documenting it?

mokagio: _Maybe..._ there could be an internal `_errorDescription` or `strongErrorDescription` or some other better name, that is `String` and `public var errorDescription: String?` could delegate to that. The idea being to have the compiler help us remember not to return any `Optional.none`.

WordPressKitTests/Utilities/WordPressAPIErrorTests.swift
```diff
@@ -0,0 +1,26 @@
...
+
+class WordPressAPIErrorTests: XCTestCase {
+
+    func testLocalizedMessage() {
+        struct TestError: LocalizedError {
+            var errorDescription: String? = "this is a test error"
+        }
+
+        let error = WordPressAPIError.endpointError(TestError())
+        XCTAssertEqual((error as NSError).localizedDescription, "this is a test error")
```
mokagio, line 14: For consistency with the test below, what do you think of adding

```suggestion
        XCTAssertEqual(error.errorDescription, "this is a test error")
        XCTAssertEqual((error as NSError).localizedDescription, "this is a test error")
```

mokagio (approved):
Would it be worth adding tests for the other `errorDescription` mappings, or since they are logic free we can do without?

I don't have an answer for that, but the criteria I usually use is "how likely is this to change and how important is it for us that it changes _intentionally_"

WordPressKit/WordPressAPIError.swift
crazytonyli: Good point. I didn't make more changes to the implementation because I thought the returning nil case is covered by the added unit tests. But I'll add a bit documentation too.

PR #664, opened 2023-12-13
Allow special characters in url-encoded form

crazytonyli:
### Description

The existing function that encodes key-value pair as url-encoded form is too loss. I should have included more special characters in the first implementation.

The special characters that are added in this PR are the same ones used by [Alamofire](https://github.com/Alamofire/Alamofire/blob/827a879ae2dcc473dfcae8ae188b992f66cd07d6/Source/URLEncodedFormEncoder.swift#L1144-L1150). But please note, unlike Alamofire, here in this library, these special characters are only taken into consideration for url-encoded form, not URL queries. We'll continue to use iOS's API (`URLComponent`) to build URL queries.

### Testing Details

I have added a test case. Let me know if there are more tests you think I should add.

---

### DIFF
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -64,9 +64,11 @@ final class HTTPRequestBuilder {
     func body(form: [String: String]) -> Self {
         headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
         bodyBuilder = { req in
-            var url = URLComponents(string: "https://wordpress.com")!
-            url.queryItems = form.map { URLQueryItem(name: $0, value: $1) }
-            req.httpBody = url.percentEncodedQuery?.data(using: .utf8)
+            let content = form.map {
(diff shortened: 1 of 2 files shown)


WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -64,9 +64,11 @@ final class HTTPRequestBuilder {
...
         headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
         bodyBuilder = { req in
-            var url = URLComponents(string: "https://wordpress.com")!
-            url.queryItems = form.map { URLQueryItem(name: $0, value: $1) }
-            req.httpBody = url.percentEncodedQuery?.data(using: .utf8)
+            let content = form.map {
+                    "\(HTTPRequestBuilder.urlEncode($0))=\(HTTPRequestBuilder.urlEncode($1))"
+                }
+                .joined(separator: "&")
+            req.httpBody = content.data(using: .utf8)
```
mokagio, line 71: How odd that Apple doesn't have something built-in for this 🤔

mokagio (approved):
> We'll continue to use iOS's API (URLComponent) to build URL queries.

👍

PR #666, opened 2023-12-18
Add a couple of helpers to parse HTTP content

crazytonyli:
### Description

The helpers are for
1. Parsing HTTP response body as plain text.
2. Getting value of HTTP header parameters. For example, `charset` of `Content-Type: text/html; charset=utf-8`.

### Testing Details

---

### DIFF
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -7,6 +7,18 @@ struct HTTPAPIResponse<Body> {
     var body: Body
 }
 
+extension HTTPAPIResponse where Body == Data {
+    var bodyText: String? {
+        var encoding: String.Encoding?
+        if let charset = response.textEncodingName {
+            encoding = String.Encoding(ianaCharsetName: charset)
+        }
+
+        let defaultEncoding = String.Encoding.isoLatin1
+        return String(data: body, encoding: encoding ?? defaultEncoding)
+    }
+}
+
 extension URLSession {
 
     func perform<E: LocalizedError>(
--- a/WordPressKit/HTTPProtocolHelpers.swift
+++ b/WordPressKit/HTTPProtocolHelpers.swift
@@ -1,6 +1,7 @@
 import Foundation
 
 extension String.Encoding {
+    /// See: https://www.iana.org/assignments/character-sets/character-sets.xhtml
     init?(ianaCharsetName: String) {
         let encoding: CFStringEncoding = CFStringConvertIANACharSetNameToEncoding(ianaCharsetName as CFString)
         guard encoding != kCFStringEncodingInvalidId,
--- a/WordPressKit/HTTPProtocolHelpers.swift
+++ b/WordPressKit/HTTPProtocolHelpers.swift
@@ -47,7 +47,7 @@ extension HTTPURLResponse {
     /// Return parameter value in a header field.
     ///
     /// For example, you can use this method to get "charset" value from a 'Content-Type' header like
-    /// `Content-Type: applicaions/json; charset=utf-8`.
+    /// `Content-Type: applications/json; charset=utf-8`.
     func value(ofParameter parameterName: String, inHeaderField headerName: String, stripQuotes: Bool = true) -> String? {
         guard let headerValue = value(forHTTPHeaderField: headerName) else {
             return nil
--- a/WordPressKit/HTTPProtocolHelpers.swift
+++ b/WordPressKit/HTTPProtocolHelpers.swift
@@ -1,0 +1,87 @@
+import Foundation
+
+extension String.Encoding {
+    init?(ianaCharsetName: String) {
+        let encoding: CFStringEncoding = CFStringConvertIANACharSetNameToEncoding(ianaCharsetName as CFString)
+        guard encoding != kCFStringEncodingInvalidId,
+              let builtInEncoding = CFStringBuiltInEncodings(rawValue: encoding)
+        else {
+            return nil
+        }
+
+        switch builtInEncoding {
+        case .macRoman:
+            self = .macOSRoman
+        case .windowsLatin1:
+            self = .windowsCP1252
+        case .isoLatin1:
+            self = .isoLatin1
+        case .nextStepLatin:
+            self = .nextstep
+        case .ASCII:
+            self = .ascii
+        case .unicode:
+            self = .unicode
+        case .UTF8:
+            self = .utf8
+        case .nonLossyASCII:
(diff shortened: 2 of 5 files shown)


WordPressKit/HTTPClient.swift
```diff
@@ -7,6 +7,18 @@ struct HTTPAPIResponse<Body> {
     var body: Body
 }
 
+extension HTTPAPIResponse where Body == Data {
+    var bodyText: String? {
+        var encoding: String.Encoding?
```
mokagio, line 12: Nitpick. I think this should work, too:

```suggestion
        let encoding: String.Encoding?
```

No difference at runtime, I don't think, but simply the "best practice" to prefer immutable values whenever possible.

WordPressKit/HTTPClient.swift
```diff
@@ -7,6 +7,18 @@ struct HTTPAPIResponse<Body> {
...
+        if let charset = response.textEncodingName {
+            encoding = String.Encoding(ianaCharsetName: charset)
+        }
+
+        let defaultEncoding = String.Encoding.isoLatin1
```
mokagio, line 17: What prompted the choice for this locale? 

I did a super quick search and [found this on Wikipedia](https://en.wikipedia.org/wiki/ISO/IEC_8859-1):

> ISO-8859-1 was (according to the standard, at least) the default encoding of documents delivered via [HTTP](https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol) with a [MIME type](https://en.wikipedia.org/wiki/Media_type) beginning with "text/" ([HTML5](https://en.wikipedia.org/wiki/HTML5) changed this to [Windows-1252](https://en.wikipedia.org/wiki/Windows-1252)).[[1]](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#cite_note-1)[[2]](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#cite_note-2) As of December 2023, 1.3% of all (and only 14 of the top 1000[[3]](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#cite_note-3)) [web sites](https://en.wikipedia.org/wiki/Website) use ISO/IEC 8859-1.[[4]](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#cite_note-encoding-4)[[5]](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#cite_note-5) It is the most declared single-byte character encoding in the world on the Web, but as Web browsers interpret it as the superset [Windows-1252](https://en.wikipedia.org/wiki/Windows-1252), the documents may include characters from that set.

This make me think that [`.windowsCP1252`](https://developer.apple.com/documentation/swift/string/encoding/windowscp1252) might be another valid choice?

🤷‍♂️

WordPressKit/HTTPProtocolHelpers.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+
+extension String.Encoding {
+    init?(ianaCharsetName: String) {
```
mokagio, line 5: TIL about IANA, the Internet Assigned Numbers Authority. What do you think of adding a link?


```suggestion
    /// See: https://www.iana.org/assignments/character-sets/character-sets.xhtml
    init?(ianaCharsetName: String) {
```

WordPressKit/HTTPProtocolHelpers.swift
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+}
+
+extension HTTPURLResponse {
+
+    /// Return parameter value in a header field.
+    ///
+    /// For example, you can use this method to get "charset" value from a 'Content-Type' header like
+    /// `Content-Type: applicaions/json; charset=utf-8`.
```
mokagio:
```suggestion
    /// `Content-Type: applications/json; charset=utf-8`.
```

WordPressKitTests/Utilities/HTTPHeaderValueParserTests.swift
```diff
@@ -0,0 +1,69 @@
...
+    func testCaseInsensitiveParameter() {
+        XCTAssertEqual(
+            HTTPURLResponse.value(ofParameter: "CharSet", inHeaderValue: "application/json; charset=utf-8"),
+            "utf-8"
+        )
+    }
+
+    func testFirstParameter() {
+        XCTAssertEqual(
+            HTTPURLResponse.value(ofParameter: "CharSet", inHeaderValue: "application/json; charset=utf-8;"),
```
mokagio, line 24: Given the test is "first parameter" and we also have a "last parameter" test below with the same input, I'd say the input here should have been swapped?

```suggestion
            HTTPURLResponse.value(ofParameter: "CharSet", inHeaderValue: "charset=utf-8; application/json;"),
```

mokagio (approved):

WordPressKitTests/Utilities/HTTPHeaderValueParserTests.swift
crazytonyli: Good catch! The charset here is the first and last parameter. I have updated the other test case to make charset as last parameter only. https://github.com/wordpress-mobile/WordPressKit-iOS/pull/666/commits/44a4fb1d746fcdbc2e35c4e88f3d4d9c85943fdc

I don't think `charset=utf-8; application/json;` is a valid header value. The 'application/json' part is the true header value ('Content-Type'). Other texts followed are parameters, which are additional information.

WordPressKit/HTTPClient.swift
crazytonyli: I took Alamofire as a reference. 😄

WordPressKit/HTTPClient.swift
crazytonyli: Making it a `let` would require us adding a `else` case below and assigning it a nil value. The additional code doesn't seem to be worth it 😄

WordPressKit/HTTPClient.swift
mokagio: 👌

WordPressKitTests/Utilities/HTTPHeaderValueParserTests.swift
mokagio: > I don't think charset=utf-8; application/json; is a valid header value.

Good point.

I wonder what's a reasonable behavior for the code in that case 🤔

crazytonyli: "undefined behaviour" I guess 🥲 .

PR #667, opened 2023-12-18
Refactor retrieving org REST API nonce

crazytonyli:
### Description

> [!Note]
> This PR is built on top of https://github.com/wordpress-mobile/WordPressKit-iOS/pull/666. Also, it might be easier to review commit-by-commit.

Currently the retrieving nonce code is coupled with `CookieNonceAuthenticator` and Alamofire. This PR moves them out as of the authenticator implementation and makes it depending on `URLSession`. The new code will be used in upcoming .org REST API client refactor.

#### Authentication Details

.org REST API uses [cookie+nonce authentication](https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/#cookie-authentication). After user logins into the app, the app _should_ have the authentication related cookies available, which will be added automatically by `URLSession`. The `nonce` part is what we need to active find.

The library finds a nonce from either `wp-admin/post-new.php` webpage html or `wp-admin/admin-ajax.php?action=rest-nonce`. Both webpages require user signed in (authenticated via cookies, like all webpages). If user has not signed in, both webpages return non-200 status code (which can be tested using `curl -I https://<self-host-site>/<link-to-webpage>`). In that case, the library performs a sign in HTTP request with a redirection to one of the above two webpages. 

The nonce will be passed to .org REST API via `X-WP-Nonce` header. With cookies and nonce available to REST API, user can now make authenticated API calls.

#### A small change in this PR

At the moment, the library always makes the aforementioned sign in requests, which I think in most cases are redundant. Because at the time a org REST API is called, user should already signed in to the app. So the small change I made in this PR was, always getting the webpage content as the first step, and signing in with a redirection link as the fallback.

### Testing Details

I have added a few unit tests in this library. `NonceRetrievalTests.swift` is pretty much a duplication of `CookieNonceAuthenticatorTests` (which will be removed along with Alamofire).

To test this in a real app:
1. Update the Podfile in your local WordPress-iOS checkout to depend on this PR branch.
2. Set a breakpoint at [this line](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/76e18a702b010b2c53563fb5d3906e107948e874/WordPressKit/WordPressOrgRestApi.swift#L64).
3. Launch the app from Xcode.
4. Sign in using a self hosted site
5. Go to Posts and open an existing post (Gutenberg editor makes .org REST API requests).
6. In a couple seconds, Xcode should stop at the breakpoint set above. Observe the API result: it should succeed.

---


WordPressKitTests/WordPressAPI/CookieNonceAuthenticatorTests.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    func testUsingNewPostPage() {
+        stubLoginRedirect(dest: Self.newPostURL)
+        stubNewPostPage(statusCode: 200)
+
+        let authenticator = CookieNonceAuthenticator(username: "test", password: "password", loginURL: Self.siteLoginURL, adminURL: Self.siteAdminURL, version: "4.0")
+        let session = SessionManager(configuration: .ephemeral)
+        session.adapter = authenticator
+        session.retrier = authenticator
+        wait(for: [apiCallShouldSucceed(using: session)], timeout: 0.1)
```
mokagio, line 42: Not a big deal, but this syntax was a bit deceiving for me. Usually, when we `wait(for` the input is an `XCTExpectation` instance. The method called here returns the expectation but also produces the side-effect of making and API request and validating its response.

Having said that, I think it's quite a neat patter 💡 😄 . I wonder if there is a way to write the code differently to make it clearer. At the same time, it might be gold plating and we have more impactful things to think about 😄 🙃

WordPressKit/NonceRetrieval.swift
```diff
@@ -35,11 +60,37 @@ enum NonceRetrievalMethod {
         return nonce
     }
 
-    func readNonceFromAjaxAction(html: String) -> String? {
-        guard !html.isEmpty else {
+    private func readNonceFromAjaxAction(html: String) -> String? {
+        guard !html.isEmpty,
+              html.allSatisfy({ $0.isNumber || $0.isLetter })
```
mokagio, line 65: Surprised there's not a built-in for checking alphanumeric in `String`. 🤷‍♂️

mokagio (approved):
Love to see and review neat commit-by-commit refactors 👍 

For the record: I haven't tested this on device. Approving based on unit tests.

WordPressKit/NonceRetrieval.swift
crazytonyli: There is a [`CharacterSet.alphanumerics`](https://developer.apple.com/documentation/foundation/nscharacterset/1407466-alphanumerics). But it's more than number & letters.

WordPressKitTests/WordPressAPI/CookieNonceAuthenticatorTests.swift
crazytonyli: 🙈 I forgot why I did that.

PR #668, opened 2023-12-19
Add a new 'unacceptableStatusCode' error case to WordPressAPIError

crazytonyli:
### Description

This PR is a follow-up to https://github.com/wordpress-mobile/WordPressKit-iOS/pull/658#discussion_r1423315579.

At the moment, when an API call returns an failure result (non-200 status code), the existing URLSession helper returns a `Swift.Result.success` case. After dealing with it a while, I find it too weird... So, this PR changed that by reporting a new `WordPressAPIError.unacceptableStatusCode` error upon unacceptable status code(non-200 by default, but it is customizable).

This PR introduces breaking changes, in terms of compile time backwards compatibility: swift-case statements on `WordPressAPIError` may fail. However, I don't think our apps/libraries actually have such statements.

### Testing Details

See the updated unit tests.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-- Add a new `unacceptableStatusCode` error case to `WordPressAPIError`.
+- Add a new `unacceptableStatusCode` error case to `WordPressAPIError`. [#668]
 
 ### New Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- Add a new `unacceptableStatusCode` error case to `WordPressAPIError`.
 
 ### New Features
 
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -11,6 +11,7 @@ extension URLSession {
 
     func perform<E: LocalizedError>(
         request builder: HTTPRequestBuilder,
+        acceptableStatusCodes: [ClosedRange<Int>] = [200...299],
         errorType: E.Type = E.self
     ) async -> WordPressAPIResult<HTTPAPIResponse<Data>, E> {
         guard let request = try? builder.build() else {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- Add a new `unacceptableStatusCode` error case to `WordPressAPIError`. [#668]
```
mokagio, line 37: How timely. I left a comment on a PR recently that introduced breaking changes, https://github.com/wordpress-mobile/WordPress-iOS/pull/22320#issuecomment-1882125134, asking to make them non-breaking with a reminder to update them when we have enough of a corpus of breaking changes to warrant the "effort" of updating WordPressAuthenticator as well. 

This seems to be the case now 😄

WordPressKit/HTTPClient.swift
```diff
@@ -34,32 +35,57 @@ extension URLSession {
...
+            guard let newSuccess = transform(success) else {
+                return .failure(.unparsableResponse(response: success.response, body: success.body))
+            }
+
+            return .success(newSuccess)
+        }
+    }
+
+    func mapSuccess<NewSuccess: Decodable, E: LocalizedError>(
+        _ decoder: JSONDecoder = JSONDecoder()
```
mokagio: How do you feel about calling this `decodeSuccess`?

The motivation for this suggestion comes from looking at the usage in the unit tests:

```swift
        let result: WordPressAPIResult<Post, TestError> = await URLSession.shared
            .perform(request: .init(url: URL(string: "https://wordpress.org/hello")!))
            .mapSuccess()
```

Seeing that `mapSuccess` call with no argument was surprising as I'm used to seeing map-like calls taking passing a function or a closure.

mokagio (approved):
Only a non blocking naming suggestion

WordPressKit/HTTPClient.swift
crazytonyli: Renamed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/668/commits/cab530e224bdbcac065e48314d1b05688f01bbf3.

CHANGELOG.md
crazytonyli: Followed up here: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/678

PR #669, opened 2023-12-20
Support tracking HTTP request overall progress and cancellation

crazytonyli:
### Description

> [!Note]
> This PR is built on top of https://github.com/wordpress-mobile/WordPressKit-iOS/pull/668

There are many existing APIs that returns a `Progress` instance, which can be used to track progress and some can be used to perform cancellation. This PR implements these features in the new URLSession helper.

The existing APIs return a `Progress` as function return value. We can't do the same, because the new helper is an `async` function: the `Progress` will be no use if it's used as a return value. I used a parent progress instance instead. The caller now needs to pass in a valid `Progress` instance that's created exclusively for the HTTP request, and it'll be completed upon the completion of the HTTP request.

Another approach I tried is using a `inout Progress` argument, so that we don't need to create a duplicated progress instance. I envisioned it can be used as below, but ended up getting a compiler error.

```swift
var progress = Progress(totalUnitCount: 100)
Task {
    // Compiler error. The `progress` reference can't be captured in this closure.
    let _ = await perform(request, &progress)
}
// Use the progress to track the HTTP request.
track(progress)
```

### Testing Details

See the added unit tests.

---

### DIFF
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -11,13 +11,14 @@ extension URLSession {
 
     /// Send a HTTP request and return its response as a `WordPressAPIResult` instance.
     ///
-    /// ## Progress Tracking
+    /// ## Progress Tracking and Cancellation
     ///
     /// You can track the HTTP request's overall progress by passing a `Progress` instance to the `fulfillingProgress`
     /// parameter, which must satisify following requirements:
     /// - `totalUnitCount` must not be zero.
     /// - `completedUnitCount` must be zero.
     /// - It's used exclusivity for tracking the HTTP request overal progress: No children in its progress tree.
+    /// - `cancellationHandler` must be nil. You can call `fulfillingProgress.cancel()` to cancel the ongoing HTTP request.
     ///
     ///  Upon completion, the HTTP request's progress fulfills the `fulfillingProgress`.
     ///
@@ -58,6 +59,10 @@ extension URLSession {
             if let parentProgress, parentProgress.totalUnitCount > parentProgress.completedUnitCount {
                 let pending = parentProgress.totalUnitCount - parentProgress.completedUnitCount
                 parentProgress.addChild(task.progress, withPendingUnitCount: pending)
+
+                parentProgress.cancellationHandler = { [weak task] in
+                    task?.cancel()
+                }
             }
         }
     }
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -9,37 +9,86 @@ struct HTTPAPIResponse<Body> {
 
 extension URLSession {
 
+    /// Send a HTTP request and return its response as a `WordPressAPIResult` instance.
+    ///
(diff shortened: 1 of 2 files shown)


WordPressKit/HTTPClient.swift
```diff
@@ -9,37 +9,92 @@ struct HTTPAPIResponse<Body> {
...
 
+    /// Send a HTTP request and return its response as a `WordPressAPIResult` instance.
+    ///
+    /// ## Progress Tracking and Cancellation
+    ///
+    /// You can track the HTTP request's overall progress by passing a `Progress` instance to the `fulfillingProgress`
+    /// parameter, which must satisify following requirements:
+    /// - `totalUnitCount` must not be zero.
+    /// - `completedUnitCount` must be zero.
+    /// - It's used exclusivity for tracking the HTTP request overal progress: No children in its progress tree.
```
mokagio, line 20:
```suggestion
    /// - It's used exclusively for tracking the HTTP request overall progress: No children in its progress tree.
```

WordPressKit/HTTPClient.swift
```diff
@@ -9,37 +9,92 @@ struct HTTPAPIResponse<Body> {
...
+    /// - `cancellationHandler` must be nil. You can call `fulfillingProgress.cancel()` to cancel the ongoing HTTP request.
```
mokagio, line 21: What do you think about encoding these constraints in either a dedicated type that wraps `Progress` or a subclass/pre-configured instance?

#### Type wrapping `Progress`

More coding and less flexibility for advanced users, but no chance of misuse since we'd be controlling everything from here

#### Subclass / Preconfigured instance

Could be misuses, but by letting the method accept a `Progress` type, we'd give power users more flexibility

mokagio: Update: I saw the `assert` below. They do a good job at ensuring runtime safety. Still, from a "how many things does a user need to know?" perspective, I think leveraging the type system somehow would be beneficial.

WordPressKit/HTTPClient.swift
```diff
@@ -9,37 +9,92 @@ struct HTTPAPIResponse<Body> {
...
+                )
+
+                continuation.resume(returning: result)
             }
-        }
+            task.resume()
 
-        let (body, response) = result
+            if let parentProgress, parentProgress.totalUnitCount > parentProgress.completedUnitCount {
+                let pending = parentProgress.totalUnitCount - parentProgress.completedUnitCount
```
mokagio, line 61: Do we need this math given the assert above for `completedUnitCount == 0`?

mokagio (approved):

WordPressKit/HTTPClient.swift
crazytonyli: No, I don't think it's needed. But, in my head, the code expresses the idea more explicitly: letting the `task.progress` fill `parentProgress` without overflowing.

mokagio: Fair 👍

PR #670, opened 2023-12-20
Return upon failure result

crazytonyli:
### Description

A simple one line change: I don't see why we need to continue sending the request if the URL is not an HTTP URL.

---


mokagio (approved):
Good point 👍

PR #671, opened 2023-12-21
Add unit tests for RSD parser

crazytonyli:
### Description

I found a bug in `WordPressRSDParser`, where it fails to parse xmlrpc link if WordPress is the only link in the RSD content. See the new unit test `testWordPressEntryOnly`, which would fail on trunk but pass in this PR.

### Testing Details

See added unit tests.

---

### DIFF
--- a/WordPressKit/WordPressRSDParser.swift
+++ b/WordPressKit/WordPressRSDParser.swift
@@ -19,6 +19,10 @@ open class WordPressRSDParser: NSObject, XMLParserDelegate {
         if parser.parse() {
             return endpoint
         }
+        // Return the 'WordPress' API link, if found.
+        if let endpoint {
+            return endpoint
+        }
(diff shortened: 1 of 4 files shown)


WordPressKit/WordPressRSDParser.swift
```diff
@@ -19,6 +19,10 @@ open class WordPressRSDParser: NSObject, XMLParserDelegate {
         if parser.parse() {
             return endpoint
         }
+        // Return the 'WordPress' API link, if found.
+        if let endpoint {
+            return endpoint
+        }
```
mokagio, line 25: It took me a bit to understand the code. I had to expand the source and realize that `parser.parse()` calls a delegate.

I wish there was a clearer API. I suppose we could build one by wrapping `XMLParser` but it might be overkill.

mokagio (approved):

PR #672, opened 2023-12-22
Add unit tests for XMLRPC endpoint discovery

crazytonyli:
### Description

What says in the title. No library code was changed in this PR.

The added unit tests stub many different kinds of HTTP requests sent by the XMLRPC endpoint discovery code (`WordPressOrgXMLRPCValidator`). I'm not happy about their current state though: they are too verbose and the test functions are not very easy to read. I'm hoping to make some helper functions to so that the test functions can be concise enough to read, but I'll leave that to next year. 😄 

### Testing Details

None needed. This PR only adds unit tests.

---


mokagio (approved):
Appreciate the thorough testing. Thank you.

The tests are indeed a bit verbose. I couple of syntax sugar suggestions from me:

- Push `https://www.apple.com` into a `let` property to DRY
- Add some helpers for repeated stubs like `stub(condition: isHost("www.apple.com") && isPath("..."))...`
- Some of the async expectations could simplified with Nimble's `waitUntil`, but those that have `enforceOrder` might need to remain vanilla XCTest

None of these suggestions is particularly life-changing, IMHO. So, for what is worth, I think we can live with these verbose but comprehensive tests.

mokagio:
> None of these suggestions is particularly life-changing, IMHO. So, for what is worth, I think we can live with these verbose but comprehensive tests.

On top of that. Is is my understanding that we'd like to drop XMLRPC sooner rather than later, so I wouldn't advocate spend too much time on it.

PR #673, opened 2024-01-03
Support both String and Int value format when parsing `StatsSummaryTimeIntervalData` 

### LINKED ISSUE #22313: EXC_BREAKPOINT: Exception 6, Code 1, Subcode 4306439400
Sentry Issue: [JETPACK-IOS-17JR](https://a8c.sentry.io/issues/4789906957/?referrer=github_integration)

```
EXC_BREAKPOINT: Exception 6, Code 1, Subcode 4306439400
  File "<compiler-generated>", in range
  File "<compiler-generated>", in Array._checkSubscript
  File "<compiler-generated>", in Array.subscript.getter
  File "ThisWeekWidgetStats.swift", line 41, in ThisWeekWidgetStats.daysFrom
  File "<compiler-generated>", in ThisWeekWidgetStats.daysFrom
...
(21 additional frame(s) were not displayed)
```

staskus:
### Description

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/22313

> When parsing `StatsSummaryTimeIntervalData` sometimes cast fails when the number (e.g. the views count) in the JSON we get from the server is represented by a `NSTaggedPointerString` instead of an Int when it's deserialized by `JSONSerialization`
> -- @guarani 

More context: p1704230918784719/1704187478.513589-slack-C012H19SZQ8 and https://github.com/wordpress-mobile/WordPress-iOS/pull/22314#issuecomment-1874628765

## Solution

1. Wrote a failing test that uses `JSON` with mixed number format
2. Updated `StatsSummaryTimeIntervalData` parsing to support both String and Int as an acceptable type

### Codable

> A better fix would be to use Codable instead of NSJSONSerialization 
> -- @guarani 

The JSON structure is pretty unconventional and has a "fields" array defining the order of the values in the "data" array. In this case, this manual parsing logic makes sense to me and I didn't want to invest time in refactoring it. However, since this logic is covered by tests it makes refactoring easier and safer.

### Testing Details

Unit Tests should succeed.

---


staskus:
@guarani closing this issue. I think this change is unnecessary given it comes from the backend side (p1704282088796849/1704187478.513589-slack-C012H19SZQ8). We could likely find plenty of places in the code where casting would fail if JSON types change.

PR #674, opened 2024-01-03
Add Remote Feature Flag Params to Dashboard Cards endpoint

salimbraksa:
### Description

Part of https://github.com/wordpress-mobile/WordPress-iOS/issues/22315

This PR passes the following params to the dashboard cards endpoint: 
* `device_id`
* `platform`
* `build_number`
* `marketing_version`
* `identifier`

### Testing Details
See https://github.com/wordpress-mobile/WordPress-iOS/pull/22320

## Note 
Please remove the `[Status] DO NOT MERGE` label and merge after:
* the backend change is deployed D133371-code
* the WPiOS PR is approved so that the two PRs are merged together and no breakages is introduced in `trunk`

---

### DIFF
--- a/WordPressKit/DashboardServiceRemote.swift
+++ b/WordPressKit/DashboardServiceRemote.swift
@@ -36,12 +36,7 @@ open class DashboardServiceRemote: ServiceRemoteWordPressComREST {
         let cardsParams: [String: AnyObject] = [
             "cards": cards.joined(separator: ",") as NSString
         ]
-        let featureFlagParams = try {
-            let params = FeatureFlagRemote.FetchAllEndpointParams(deviceId: deviceId)
-            let encoder = JSONEncoder()
-            let data = try encoder.encode(params)
-            return try JSONSerialization.jsonObject(with: data) as? [String: AnyObject]
-        }()
+        let featureFlagParams = try FeatureFlagRemote.FetchAllEndpointParams(deviceId: deviceId).dictionaryRepresentation()
         return cardsParams.merging(featureFlagParams ?? [:]) { first, second in
             return first
         }
--- a/WordPressKit/DashboardServiceRemote.swift
+++ b/WordPressKit/DashboardServiceRemote.swift
@@ -1,12 +1,21 @@
 import Foundation
 
 open class DashboardServiceRemote: ServiceRemoteWordPressComREST {
-    open func fetch(cards: [String], forBlogID blogID: Int, success: @escaping (NSDictionary) -> Void, failure: @escaping (Error) -> Void) {
+    open func fetch(
+        cards: [String],
+        forBlogID blogID: Int,
+        deviceId: String,
+        success: @escaping (NSDictionary) -> Void,
+        failure: @escaping (Error) -> Void
+    ) {
         let requestUrl =  self.path(forEndpoint: "sites/\(blogID)/dashboard/cards-data/", withVersion: ._2_0)
+        var params: [String: AnyObject]?
 
-        let params: [String: AnyObject] = [
-            "cards": cards.joined(separator: ",") as NSString
-        ]
+        do {
+            params = try self.makeQueryParams(cards: cards, deviceId: deviceId)
+        } catch {
+            failure(error)
+        }
 
         wordPressComRestApi.GET(requestUrl,
                                 parameters: params,
@@ -21,6 +30,21 @@ open class DashboardServiceRemote: ServiceRemoteWordPressComREST {
             failure(error)
             WPKitLogError("Error fetching dashboard cards: \(error)")
         })
+    }
+
+    private func makeQueryParams(cards: [String], deviceId: String) throws -> [String: AnyObject] {
+        let cardsParams: [String: AnyObject] = [
+            "cards": cards.joined(separator: ",") as NSString
+        ]
+        let featureFlagParams = try {
+            let params = FeatureFlagRemote.RemoteFeatureFlagsEndpointParams(deviceId: deviceId)
+            let encoder = JSONEncoder()
+            let data = try encoder.encode(params)
+            return try JSONSerialization.jsonObject(with: data) as? [String: AnyObject]
+        }()
+        return cardsParams.merging(featureFlagParams ?? [:]) { first, second in
+            return first
+        }
     }
 
     enum ResponseError: Error {
--- a/WordPressKit/DashboardServiceRemote.swift
+++ b/WordPressKit/DashboardServiceRemote.swift
@@ -36,7 +36,7 @@ open class DashboardServiceRemote: ServiceRemoteWordPressComREST {
         let cardsParams: [String: AnyObject] = [
             "cards": cards.joined(separator: ",") as NSString
         ]
-        let featureFlagParams = try FeatureFlagRemote.FetchAllEndpointParams(deviceId: deviceId).dictionaryRepresentation()
+        let featureFlagParams = try SessionDetails(deviceId: deviceId).dictionaryRepresentation()
         return cardsParams.merging(featureFlagParams ?? [:]) { first, second in
             return first
         }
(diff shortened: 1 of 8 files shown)


WordPressKit/DashboardServiceRemote.swift
```diff
@@ -21,17 +32,19 @@ open class DashboardServiceRemote: ServiceRemoteWordPressComREST {
...
-        path?.queryItems = [URLQueryItem(name: "cards", value: cardsEncoded)]
-
-        guard let endpoint = path?.string else {
-            return nil
+    private func makeQueryParams(cards: [String], deviceId: String) throws -> [String: AnyObject] {
+        let cardsParams: [String: AnyObject] = [
+            "cards": cards.joined(separator: ",") as NSString
+        ]
+        let featureFlagParams = try {
+            let params = FeatureFlagRemote.RemoteFeatureFlagsEndpointParams(deviceId: deviceId)
```
hassaanelgarem: I wonder if this struct should still live inside `FeatureFlagRemote` now that it's shared with an unrelated class.

WordPressKit/FeatureFlagRemote.swift
```diff
@@ -9,17 +9,22 @@ open class FeatureFlagRemote: ServiceRemoteWordPressComREST {
     }
 
     open func getRemoteFeatureFlags(forDeviceId deviceId: String, callback: @escaping FeatureFlagResponseCallback) {
+        self.getRemoteFeatureFlags(params: .init(deviceId: deviceId), callback: callback)
+    }
 
+    open func getRemoteFeatureFlags(params: RemoteFeatureFlagsEndpointParams, callback: @escaping FeatureFlagResponseCallback) {
```
hassaanelgarem: The new `getRemoteFeatureFlags(params:callback:)`  is only used from within the `getRemoteFeatureFlags(forDeviceId deviceId:callback:)` function. This can simplify things a bit:
```suggestion
    open func getRemoteFeatureFlags(forDeviceId deviceId: String, callback: @escaping FeatureFlagResponseCallback) {
        let params = RemoteFeatureFlagsEndpointParams(deviceId: deviceId)
```

WordPressKit/FeatureFlagRemote.swift
```diff
@@ -9,17 +9,22 @@ open class FeatureFlagRemote: ServiceRemoteWordPressComREST {
...
-            "platform": "ios" as NSString,
-            "build_number": NSString(string: Bundle.main.infoDictionary?["CFBundleVersion"] as? String ?? "Unknown"),
-            "marketing_version": NSString(string: Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? "Unknown"),
-            "identifier": NSString(string: Bundle.main.bundleIdentifier ?? "Unknown")
-        ]
+        do {
+            let encoder = JSONEncoder()
+            let data = try encoder.encode(params)
+            parameters = try JSONSerialization.jsonObject(with: data) as? [String: AnyObject]
+        } catch let error {
```
hassaanelgarem: [Suggestion] How about moving this to a helper function on `RemoteFeatureFlagsEndpointParams`? Something like this:
```
    func dictionaryRepresentation() throws -> [String: AnyObject]?  {
        let encoder = JSONEncoder()
        let data = try encoder.encode(self)
        return try JSONSerialization.jsonObject(with: data) as? [String: AnyObject]
    }
```

This way it can be used here and also in `DashboardServiceRemote.swift:L41`

hassaanelgarem (approved):
Looks good to me! 🚀 
Kudos on adding tests!
Only left a few code improvement suggestions.

WordPressKit/FeatureFlagRemote.swift
salimbraksa: Pushed commits resolving this ✅

WordPressKit/FeatureFlagRemote.swift
```diff
@@ -60,7 +55,7 @@ open class FeatureFlagRemote: ServiceRemoteWordPressComREST {
                                 })
     }
 
-    public struct RemoteFeatureFlagsEndpointParams {
+    public struct FetchAllEndpointParams {
```
hassaanelgarem: @salimbraksa `FetchAllEndpointParams` feels a bit confusing. How about `SessionDetails`?

Also, can we move it outside of the `FeatureFlagRemote` scope? Referencing it by `FeatureFlagRemote.FetchAllEndpointParams` hurts the readability as it implies it is tighly coupled with feature flags while it's not. This is what I meant here https://github.com/wordpress-mobile/WordPressKit-iOS/pull/674#discussion_r1442456627

WordPressKit/DashboardServiceRemote.swift
salimbraksa: That was the plan initially, but couldn't figure out a name for that type. 😅

WordPressKit/FeatureFlagRemote.swift
salimbraksa: My concern with global scope is the following:
1. Let's say another developer added a valid "session" attribute in `SessionDetails`. 
2. The new attribute is encoded and returned in the dictionary representation.
3. Now that attribute will be sent in all other endpoints, and I'm not sure how endpoints handle extra attributes that don't need.

This is why I kept that type tightly related to `Feature Flags`, and if you think about it, the new params added in Dashboard Cards are for the Feature Flags mechanism.

hassaanelgarem: > Now that attribute will be sent in all other endpoints, and I'm not sure how endpoints handle extra attribute that don't need.

We can assume that developers will seek to understand what this type is used for before changing it.
However, extra params won't cause an issue either way.

salimbraksa: > We can assume that developers will seek to understand what this type is used for before changing it.

I agree, but my concern is that if we expand the SessionDetails with new 'session' attributes in the future, we need to ensure they are not inadvertently included in the Feature Flags and Dashboard Cards endpoint query params.

If we rename that type to `SessionDetails`, I think the Feature Flag Remote and Dashboard Cards service should select only the params they need.

Also, while I'm sure developers double check API usage before making changes, I think it's also important to write resilient code that mitigates the risks associated with human error.

salimbraksa: I'll rename the type to `SessionsDetails` and update the Unit Tests to ensure **only** the the params required by the endpoint are sent.

hassaanelgarem: > However, extra params won't cause an issue either way.

Given this, I think we would be over-engineering to avoid something that's not critical.

salimbraksa: Resolved in 28cc006f19a92ddb7e0b307926cedd1c9449bca4

> However, extra params won't cause an issue either way.

Passing extra params to endpoints is a code smell in my opinion. Moreover, some backend APIs may reject the request if it includes additional parameters as a security measure.

I agree that we shouldn't spend time on imaginary problems. For now, it's enough to have Unit Tests to check for this. I looked at the tests we already have, and they'll fail if extra params are sent to `Feature Flags` and `Dashboard Cards` endpoints.

mokagio:
Thank you for working on this @salimbraksa and for adding tests. I also appreciate the helper functions, which enable writing othre tests in the future 🎉 

I landed here while preparing the library for Jetpack and WordPress iOS [24.0](https://github.com/wordpress-mobile/WordPress-iOS/milestone/265). I noticed this PR introduce a breaking change, which is something we should be careful with because of the downstream implication in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS which depends on WordPressKit.

Given the app PR that uses this logic, https://github.com/wordpress-mobile/WordPress-iOS/pull/22320, has not been merged at this time and won't be part of 24.0, I will not ship a new release with this breaking change just yet.

Ideally, we'd want to provide a non-breaking API. If that's not possible, given we'll need to go through the effort of shipping the breaking change, we might as well look for other breaking low hanging fruits to slot in. 

Cheers (and Happy New Year!)

salimbraksa:
Hey @mokagio, the other PR was supposed to be merged on Friday, but I just realized it didn't because of a flaky unit test unrelated to my PR.

The PR was already been reviewed and approved, so I’ll work on fixing the failing test and aim to merge it into release/24.0.

PR #680, opened 2024-01-11
Update block editor settings tests

crazytonyli:
### Description

Currently `BlockEditorSettingsServiceRemoteTests` mocks API clients (i.e. `MockWordPressComRestApi`) to make assertions. This PR replaces those mocks with HTTP stubs, so that the tests send actually HTTP requests and don't rely on specific types (`WordPressComRestApi` and `WordPressOrgRestApi`).

The main motivation of this change is, I can reuse the same unit tests (with minimal code changes) in the upcoming `BlockEditorSettingsServiceRemote` refactor.

### Testing Details

None needed.

---


WordPressKitTests/BlockEditorSettingsServiceRemoteTests.swift
```diff
@@ -42,23 +49,25 @@ extension BlockEditorSettingsServiceRemoteTests {
...
-        var mockedResponse = mockedData(withFilename: twentytwentyoneResponseFilename) as! [[String: Any]]
+        stub(condition: isHost("public-api.wordpress.com") && isPath("/wp/v2/sites/1/themes") && containsQueryParams(["status": "active"])) { _ in
+            var mockedResponse = self.mockedData(withFilename: self.twentytwentyoneResponseFilename) as! [[String: Any]]
+
+            // Clear out Gradients
+            var theme = mockedResponse[0]
+            var themeSupport = theme[RemoteEditorTheme.CodingKeys.themeSupport.stringValue] as! [String: Any]
+            themeSupport[RemoteEditorThemeSupport.CodingKeys.gradients.stringValue] = "false"
+            theme[RemoteEditorTheme.CodingKeys.themeSupport.stringValue] = themeSupport
+            mockedResponse[0] = theme
```
mokagio, line 65: From the PR description:

> The main motivation of this change is, I can reuse the same unit tests (with minimal code changes) in the upcoming `BlockEditorSettingsServiceRemote` refactor.

Would you also consider extracting this and the other response data manipulations in dedicated methods? It would for sure improve readability. That's useful per se, but at this level I wouldn't make a strong case for it unless there was a case for reuse. If the tests are going to be reused somehow, then the refactor investment to extract some of this code would be useful.

WordPressKitTests/BlockEditorSettingsServiceRemoteTests.swift
```diff
@@ -1,23 +1,27 @@
...
+        stub(condition: { _ in true }) {
+            XCTFail("Unexpected request: \($0)")
+            return HTTPStubsResponse(error: URLError(URLError.Code.networkConnectionLost))
+        }
+        service = BlockEditorSettingsServiceRemote(remoteAPI: WordPressComRestApi())
     }
 
     func mockedData(withFilename filename: String) -> AnyObject {
-        let json = Bundle(for: BlockEditorSettingsServiceRemoteTests.self).url(forResource: filename, withExtension: "json")!
+        let json = Bundle(for: BlockEditorSettingsServiceRemoteTests.self).url(forResource: filename, withExtension: nil)!
```
mokagio, line 24: Just curious why you decided to add the JSON extension to the file names and use `nil` here rather than keeping the code as it was?

To be clear, I like it. It makes it clearer at the filename `let` level what extension it is. We have little reason to expect to be using files with other extensions in the future, so we might as well keep the information (name + extension) all in the same place.

mokagio (approved):

WordPressKitTests/BlockEditorSettingsServiceRemoteTests.swift
crazytonyli: Using full filename means we can reuse the constants in here and the `OHPathForFile` calls.

WordPressKitTests/BlockEditorSettingsServiceRemoteTests.swift
crazytonyli: I wasn't clear. By reusing, I don't mean reusing the test code somewhere else. Rather, reusing the same unit tests to test refactored `BlockEditorSettingsServiceRemote`, with only a few simple code changes.

I will soon open a PR to refactor `BlockEditorSettingsServiceRemote`, I don't think there will be much changes in the unit tests, because the unit tests (the ones that changed in this PR) now stub HTTP requests, instead of using fake implementations which can change if the `BlockEditorSettingsServiceRemote` internal changes (which is exactly what is going to happen in the refactor). By reusing, I mean that the unit test no longer rely on the `BlockEditorSettingsServiceRemote` API (still a little bit, because it needs to create a instance of it, but not so much deep into its internal). And the same tests can verify the refactored result.

PR #661, opened 2023-12-13
Use URLSession in WebAuthn related API calls

crazytonyli:
### Description

This PR continues to remove Alamofire from `WordPressOAuthClient`.

### Testing Details

Testing this PR on the app side requires using passkey as the second authentication factor. Since passkey authentication is disabled at the moment, you'll need to make some code changes to use it. Here are a summary of what needs to be done:

1. Checkout the trunk branch of the WordPress-iOS repo.
2. Make the following code changes:

<details>

```diff
diff --git a/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift b/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
index 57dece8291..2029f1eca1 100644
--- a/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
+++ b/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
@@ -79,7 +79,7 @@ extension WordPressAuthenticationManager {
                                                    enableSignupWithGoogle: AppConfiguration.allowSignUp,
                                                    enableUnifiedAuth: true,
                                                    enableUnifiedCarousel: true,
-                                                   enablePasskeys: false,
+                                                   enablePasskeys: true,
                                                    enableSocialLogin: true)
     }
 
diff --git a/WordPress/Jetpack/JetpackDebug.entitlements b/WordPress/Jetpack/JetpackDebug.entitlements
index 301d465cda..cf55625401 100644
--- a/WordPress/Jetpack/JetpackDebug.entitlements
+++ b/WordPress/Jetpack/JetpackDebug.entitlements
@@ -14,7 +14,6 @@
 		<string>webcredentials:wordpress.com</string>
 		<string>applinks:apps.wordpress.com</string>
 		<string>applinks:wordpress.com</string>
-		<string>applinks:*.wordpress.com</string>
 		<string>applinks:public-api.wordpress.com</string>
 		<string>applinks:links.wp.a8cmail.com</string>
 	</array>

```

</details>

3. Update the `WordPressKit` dependency in Podfile to point to this branch and run `bundle exec pod install`.
4. Use Xcode to build and run the app. See p91TBi-aIl-p2 if you use simulator to test.
5. Sign in to the app using your A8C account (which should have a registered passkey).

---

- [x] Please check here if your pull request includes additional test coverage.
- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #681, opened 2024-01-12
Use URLSession in social login endpoints

crazytonyli:
### Description

This PR continues to remove Alamofire from the OAuth client class.

### Testing Details

Please refer to https://github.com/wordpress-mobile/WordPressKit-iOS/pull/661

---

### DIFF
--- a/WordPressKit/WordPressComOAuthClient.swift
+++ b/WordPressKit/WordPressComOAuthClient.swift
@@ -89,7 +89,6 @@ public final class WordPressComOAuthClient: NSObject {
     @objc public static let WordPressComOAuthDefaultApiBaseUrl = "https://public-api.wordpress.com"
 
     enum WordPressComURL: String {
-        case socialLogin = "/wp-login.php?action=social-login-endpoint&version=1.0"
         case socialLogin2FA = "/wp-login.php?action=two-step-authentication-endpoint&version=1.0"
         case socialLoginNewSMS2FA = "/wp-login.php?action=send-sms-code-endpoint"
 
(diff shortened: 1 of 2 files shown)


WordPressKit/WordPressComOAuthClient.swift
```diff
@@ -802,4 +844,25 @@ private extension WordPressComOAuthClient {
...
+        case authenticateWith2FA = "two-step-authentication-endpoint"
+
+        var queryItems: [URLQueryItem] {
+            var items = [URLQueryItem(name: "action", value: rawValue)]
+            if self == .authenticate || self == .authenticateWith2FA {
+                items.append(URLQueryItem(name: "version", value: "1.0"))
+            }
+            return items
+        }
+    }
```
mokagio, line 860: What do you think of adding unit tests for this conditional logic?

```swift
XCTAssertEqual(sendOTPViaSMS.queryItems.count, 1)
XCTAssertTrue(sendOTPViaSMS.queryItems.contains[URLQueryItem(name: "action", value: "send-sms-code-endpoint")])
...
```

Admittedly, it is trivial logic, but one that could be hard to find if broken.

mokagio (approved):

WordPressKit/WordPressComOAuthClient.swift
crazytonyli: [The existing tests](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/d182249d336ce5e3d34e830470e93edf82cf0ab7/WordPressKitTests/WordPressComOAuthClientTests.swift#L12-L13), which stub URL requests by their full URL, have covered these query parameters. Removing either one of these two query will cause the existing tests to fail. So, I think we are all good in terms of coverage? But let me know if you think we should add dedicated tests for these query parameters specifically.

PR #682, opened 2024-01-12
Add syntax sugar to execute callbacks on `Result`

mokagio:
Seeing those

```swift
            switch result {
            case let .success(data):
                success(data)
            case let .failure(error):
                failure(error)
            }
```

"bothered" me. I feels like a lot of coding to type and read for something so straightforward. 

This is my attempt to be proactive about it instead of just complaining 😄


crazytonyli (approved):
Nice! 💯

PR #683, opened 2024-01-14
Use URLSession in request sms code during social login API

crazytonyli:
### Description

This PR continues to remove Alamofire from the OAuth client class.

#### An implicit API change

Please note, there is an implicit API change in this PR. This PR refactor the function to always call `success` with a "new nonce" code, whereas the original code may call either `success` or `failure` closure with a "new nonce".

The changed function is only called in [one place in WordPressAuthenticator](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/6aeeb301fe2331efdd38fb6d730d04c56d722124/WordPressAuthenticator/Services/LoginFacade.swift#L33-L51). As you can see in the call site, "success" and "failure" scenarios are treated almost exactly the same: The only difference is tracking. So, I don't think this API change will have any impact at all in terms of app features.

### Testing Details

Please refer to https://github.com/wordpress-mobile/WordPressKit-iOS/pull/661

---


mokagio (approved):
Sorry for the late review, resulting in the need for a merge conflict resolution on this branch.

About the implicit change... 🤔 The good news is that, as far as I can see, `nonceSMS` which the consumer function in WordPressAuthenticator updates is used only in the library. 

For ease of conversation, here's the function:

```swift
    func requestSocial2FACode(with loginFields: LoginFields) {
        guard let nonce = loginFields.nonceInfo?.nonceSMS else {
            return
        }

        wordpressComOAuthClientFacade.requestSocial2FACode(
            userID: loginFields.nonceUserID,
            nonce: nonce,
            success: { [weak self] newNonce in
                guard let self = self else {
                    return
                }

                loginFields.nonceInfo?.nonceSMS = newNonce

                if self.tracker.shouldUseLegacyTracker() {
                    WordPressAuthenticator.track(.twoFactorSentSMS)
                }
        }) { (_, newNonce) in
            if let newNonce = newNonce {
                loginFields.nonceInfo?.nonceSMS = newNonce
            }
            WPAuthenticatorLogError("Failed to request one time code")
        }
```

With these changes, in case of failure, `loginFields.nonceInfo?.nonceSMS` would not be updated.

What would this result to at runtime? From my reading of the code, it would mean that if the user calls `requestSocial2FACode(with loginFields: LoginFields)` the `nonce` passed to the API would be the same as the previous request. Given the previous call failed, one would expect the new call to fail as well. This would leave the user stuck in a failure loop, which is far from ideal.

At the same time, I do wonder how much of a concern this is _in the real world_. Mainly, if the call failed in the first place, then something was wrong with the user data either on the app or on the backend. Would make a new call really fix it? And if the call failed because of issues such as network time out, then we wouldn't have had a new nonce back anyway.

All in all, I think we can afford to ship the change as is and monitor it's behavior.

PR #686, opened 2024-01-15
Use hard-coded error domain string in unit tests

crazytonyli:
### Description

This PR simply replaces all `String(reflecting: WordPressComRestApiError.self)` with its actual value `"WordPressKit.WordPressComRestApiError"`.

The `WordPressComRestApiError` type will be renamed in a upcoming PR. The only purpose of this PR is reducing some noise in the upcoming PR.

---

### DIFF
--- a/WordPressKitTests/AccountServiceRemoteRESTTests.swift
+++ b/WordPressKitTests/AccountServiceRemoteRESTTests.swift
@@ -90,7 +90,7 @@ class AccountServiceRemoteRESTTests: RemoteTestCase, RESTTestable {
                 expect.fulfill()
                 return
             }
-            XCTAssertEqual(error.domain, String(reflecting: WordPressComRestApiError.self), "The error domain should be WordPressComRestApiError")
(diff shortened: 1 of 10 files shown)


mokagio (approved):
> The `WordPressComRestApiError` type will be renamed in a upcoming PR. The only purpose of this PR is reducing some noise in the upcoming PR.

Does this mean that the `domain` value in the tests will remain the same, i.e. `WordPressKit.WordPressComRestApiError`. Or will it change also? 

If it will change as well, then I'd expect the new string entries in this PR to need to be update, too, right? 

Anyways, using the literal value over the computed one seems like a good idea to me as it decouples from implementation details.

crazytonyli:
> Does this mean that the domain value in the tests will remain the same, i.e. WordPressKit.WordPressComRestApiError. Or will it change also?

@mokagio They will remain the same. The unit tests should continue to pass, and the checks (where compare domain value `WordPressKit.WordPressComRestApiError`) in the apps continue to pass too.

PR #687, opened 2024-01-16
Fix an issue where the override parameter was ignored

crazytonyli:
### Description

I noticed the `override` argument in `func append(query: [URLQueryItem], override: Bool)` wasn't used. This PR fixes that.

This bug doesn't have any impact in the existing (which is only a couple of places) usages of `HTTPRequestBuilder`.

### Testing Details

See the added unit tests. Let me know if you think there are more should be added.

---

### DIFF
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -57,7 +57,17 @@ final class HTTPRequestBuilder {
     }
 
     func append(query: [URLQueryItem], override: Bool) -> Self {
-        urlComponents.queryItems = (urlComponents.queryItems ?? []) + query
+        var allQuery = urlComponents.queryItems ?? []
+
+        if override {
+            let newKeys = Set(query.map { $0.name })
+            allQuery.removeAll(where: { newKeys.contains($0.name) })
+        }
+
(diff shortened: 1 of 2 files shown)


mokagio (approved):
Good catch and sorry for missing this.

What do you think about setting a default value?

It's a bit tricky because "append" with "override true" doesn't behave like a true append, more like "add or update", but I feel like that would be the best behavior in the real world, where most of the time we want only one up-to-date value for a query item key?

crazytonyli:
@mokagio Default value added in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/687/commits/437444dbc064d9637305005bf655b3dd0992bc63.

crazytonyli:
> I feel like that would be the best behavior in the real world, where most of the time we want only one up-to-date value for a query item key?

In the existing APIs that use Alamofire, call sites can pass multiple value to the same query key and it'll be encoded as `key[]=value1&key[]=value2`. It's hard to say which is preferred or should be default. I named the function as `append` because I'm kindof following Foundation's API here: where `URLComponents`'s query is defined as `[URLQueryItem]` and it doesn't imply query keys should be uniq in any way.

PR #688, opened 2024-01-16
Add a test case for the existing URL query encoding behaviour

crazytonyli:
### Description

Adding this test to ensure an upcoming refactor to `WordPressComRestApi` doesn't break how URL query is encoded into API URLs.

### Testing Details

None needed.

---

- [x] Please check here if your pull request includes additional test coverage.
- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #689, opened 2024-01-16
Support encoding 'Any' query in HTTPRequestBuilder

crazytonyli:
> [!Note]
> This PR is built on top of #688.

### Description

The implementation is the same as the existing query encoding behaviour.

### Testing Details

See the new unit test.

---

### DIFF
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -56,6 +56,29 @@ final class HTTPRequestBuilder {
         append(query: [URLQueryItem(name: name, value: value)], override: override)
     }
 
+    func query(_ dict: [String: Any]) -> Self {
+        dict.reduce(self) {
+            $0.query(name: $1.key, value: $1.value)
+        }
+    }
+
+    func query(name: String, value: Any) -> Self {
+        switch value {
+        case let array as [Any]:
+            return array.reduce(self) {
+                $0.query(name: "\(name)[]", value: $1)
+            }
+        case let object as [String: Any]:
+            return object.reduce(self) {
+                $0.query(name: "\(name)[\($1.key)]", value: $1.value)
+            }
+        case let value as Bool:
+            return query(name: name, value: value ? "1" : "0", override: false)
+        default:
+            return query(name: name, value: "\(value)", override: false)
+        }
+    }
+
     func append(query: [URLQueryItem], override: Bool = false) -> Self {
         var allQuery = urlComponents.queryItems ?? []
 
--- a/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
+++ b/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
@@ -140,6 +140,60 @@ class HTTPRequestBuilderTests: XCTestCase {
         )
     }
 
+    @available(iOS 16.0, *)
+    func testSetQueryWithDictionary() throws {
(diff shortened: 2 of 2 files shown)


crazytonyli:
Converting this PR to draft because we should use the same logic to implement building URL query and url encoded ~"multipart/form-data"~ "application/x-www-form-urlencoded".

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -56,6 +56,29 @@ final class HTTPRequestBuilder {
         append(query: [URLQueryItem(name: name, value: value)], override: override)
     }
 
+    func query(_ dict: [String: Any]) -> Self {
+        dict.reduce(self) {
+            $0.query(name: $1.key, value: $1.value)
+        }
+    }
+
+    func query(name: String, value: Any) -> Self {
```
mokagio, line 65: How do you feel about

```suggestion
    func queryParameter(name: String, value: Any) -> Self {
```

?

I don't know why (which is in itself a reason to discard my suggestion) but simply "query" for a method that sets a parameter only seems unclear.

WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
```diff
@@ -140,6 +140,60 @@ class HTTPRequestBuilderTests: XCTestCase {
         )
     }
 
+    @available(iOS 16.0, *)
+    func testSetQueryWithDictionary() throws {
+        // The test data and assertions hre should be kept the same as the ones in `WordPressComRestApiTests.testQuery()`.
```
mokagio, line 145:
```suggestion
        // The test data and assertions here should be kept the same as the ones in `WordPressComRestApiTests.testQuery()`.
```

mokagio, line 145: > The test data and assertions here should be kept the same as the ones in `WordPressComRestApiTests.testQuery()`.

Have you considered extracting them? I noticed there is at least one additional value compared to `testQuery()` but I guess that could be addressed either with a dedicated test or by appending it here?

However, I realize that it might be overkill to got to that length for only two tests.

WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
```diff
@@ -140,6 +140,60 @@ class HTTPRequestBuilderTests: XCTestCase {
...
+            "nested-dict[outter2][inner1]=value1",
+            "nested-dict[outter2][inner2]=value2",
+            "array[]=true",
+            "array[]=1",
+            "array[]=0",
+        ]
+
+        for item in expected {
+            XCTAssertTrue(query?.contains(item) == true, "Missing query item: \(item)")
+        }
```
mokagio, line 194: Same suggestion as in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/688/files#r1454598703

WordPressKit/HTTPRequestBuilder.swift
crazytonyli: I feel like these APIs look less strange from call sites. Here is an example:

```
HTTPRequestBuilder(url: "https://public-api.wordpress.com")
    .method(.post)
    .append(path: "/rest/v1.2/sites/\(id)/posts")
    .query(name: "context", value: "mobile")
    .query(name: "mode", value: "html")
```

crazytonyli:
Re-implemented in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/694.

PR #691, opened 2024-01-16
Change baseUrlString to baseURL

crazytonyli:
### Description

This PR changes the base url variable in `WordPressComRestApi` from `String` to `URL`.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- `WordPressComRestApi` initialisers now accept a `baseURL: URL` parameter instead of `baseUrlString: String`. [#691]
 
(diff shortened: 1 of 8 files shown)


mokagio (approved):
Did you consider deprecating the `String`-based method and having it side-by-side the better `URL`-based one?

I'm all in for using a stronger type. Love it. I only wonder about another major version.

Then again, maybe there are more breaking changes coming up that cannot be as easily made compatible? Plus, it's really no big deal to ship another major version, only an additional PR on WordPressAuthenticator...

Approving to unblock, but with auto-merge disabled. 👍

crazytonyli:
@mokagio #692 is another breaking change. I have been trying to avoid breaking changes, but when it's unavoidable, I'll try to ship breaking changes in one major version.

PR #694, opened 2024-01-17
Unify query and url-encoded-form implementation

crazytonyli:
### Description

There are two changes in this PR.

#### URL encoding in query and form

We used to have different implementation around encoding name-value pair into URL query and web form in POST request body: `URLComponent`'s query encoding implementation is used in building URL query, and our own customized implementation (which aligns with Alamofire's implementation for backwards compatibility) is used to build web form.

Now, the same implementation, which is our customized one, is used to build both query and web form. Backwards compatibility is the only reason for this change.

#### HTTPRequestBuilder now does not modify the original URL

Say we have `HTTPRequestBuilder` instance created using an URL "https://wp.org/api?locale=en". When adding a query `.query(name: "locale", value: "zh")`, the new query replaces the one in the original url. This behaviour is changed in this PR. The added query will be appended onto the original URL, which means the original url content will always be in the final result. To be honest, I don't like this new behavior, but that's what Alamofire does and we'll need to do it for backwards compatibility.

### Testing Details

See added unit tests.

---


WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
```diff
@@ -5,6 +5,43 @@ import XCTest
 
 class HTTPRequestBuilderTests: XCTestCase {
 
+    static let nestedParameters: [String: Any] =
+        [
+            "number": 1,
+            "nsnumbe-true": NSNumber(value: true),
```
jkmassel: nit: typo?

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -1,5 +1,9 @@
 import Foundation
 
+/// A builder type that adding HTTP request data onto an URL.
```
jkmassel: nit: typo?

jkmassel (approved):
Little things, the rest we discussed on Zoom

WordPressKit/HTTPRequestBuilder.swift
crazytonyli: 🙈 Fixed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/694/commits/59bd717519e4ce71a4091193bd2015ea59454607.

PR #695, opened 2024-01-17
Add a underlyingError property to a couple of error cases

crazytonyli:
### Description

When encounter URL encoding errors or response parsing errors, we should surface those errors through API, instead of silently swallowing them.

This is another breaking change that is not used by the apps.

### Testing Details

I have added a couple of unit tests. Let me know if you think there are more should be added.

---

### DIFF
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -114,14 +114,14 @@ extension URLSession {
 extension WordPressAPIResult {
 
     func mapSuccess<NewSuccess, E: LocalizedError>(
-        _ transform: (Success) -> NewSuccess?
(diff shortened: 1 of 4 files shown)


mokagio (approved):
Glad to see the `return nil` replaced with `throw <error>`.

In defense of the code, it might be that the `throw` mechanic wasn't around at the time it was written. Still, about time we made full use of modern syntax.

crazytonyli:
> In defense of the code, it might be that the throw mechanic wasn't around at the time it was written.

It's actually [an oversight on my part][PR]. 🙈 

[PR]: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/653

PR #696, opened 2024-01-18
Refactor WordPressComRestApiError

crazytonyli:
### Description

In upcoming PRs, I'll add functions to `WordPressComRestApi` which use `URLSession` to make API calls. The newly added `URLSession` returns `enum WordPressAPIError` type. That means, [`NSError` instances that are returned by `WordPressComRestApi`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/10.0.0/WordPressKit/WordPressComRestApi.swift#L54) will change internally: In cases where returns a `WordPressComRestApiError`, they will return `WordPressAPIError` after the refactor(*). That means the domain and code value in those `NSError` instances will be changed too. This change of course impacts the error handling code on the app side.

> [!Note]
> There are some info in `WordPressKit/WordPressAPIError+NSErrorBrdige.swift` that I don't repeat here. Please have a read there too.

There are mainly two kinds of code that handles `NSError` returned by `WordPressComRestApi`:
1. Checking its `domain` and `code`: `domain == "WordPressComRestApiError" && code == WordPressComRestApiError...`
1. Casting into `WordPressComRestApiError`: `error as? WordPressComRestApiError`.

Once the aforementioned refactor is implemented, it's pretty easy to migrate the second error handling code: by simply resolving Swift compiler errors.

However, the first error handling code is more tricky to resolve. Because it's largely used in Objective-C code which can't access `enum WordPressAPIError<EndpointError>` Swift type.

One way to solve the issue is making sure returned `NSError` instances (their domain, code, user info) do not change, even though the type of errors from where they are casted has been changed: the three `NSError` instances below are equal.
1. `WordPressComRestApiError.<code> as NSError` (old code)
2. `WordPressComRestApiEndpointError(code: <code>) as NSError` (new code).
3. `WordPressComRestApiEndpointError(code: <code>) as NSError` (new code).

(*): Since we haven't implemented the new API call functions, this PR changes the existing functions to return `WordPressComRestApiEndpointError` instead of `WordPressAPIError` for now. But eventually, the errors will be `WordPressAPIError<WordPressComRestApiEndpointError>`

> [!Note]
> It might be easier to review this PR commit by commit.

### Testing Details

See the updated unit tests.

---


WordPressKit/WordPressAPIError+NSErrorBrdige.swift
```diff
@@ -0,0 +1,114 @@
...
+    public static var errorDomain: String {
+        (EndpointError.self as? CustomNSError.Type)?.errorDomain
+            ?? String(describing: Self.self)
+    }
+
+    public var errorCode: Int {
+        switch self {
+        case let .endpointError(endpointError):
+            return (endpointError as NSError).code
+        // Use negative values for other cases to reduce chances of collision with `EndpointError`.
```
mokagio, line 22: Smart

WordPressKit/WordPressAPIError+NSErrorBrdige.swift
```diff
@@ -0,0 +1,114 @@
...
+
+/// Custom `NSError` bridge implementation.
+///
+/// The implementation ensures `NSError` instances that are casted from `WordPressAPIError<EndpointError>.endpointError`
+/// are the same as those instances that are casted directly from the underlying `EndpointError` instances.
+///
+/// In theory, we should not need to implement this bridging, because we should never cast errors to `NSError`
+/// instances in any error handling code. But since there are still Objective-C callers, providing this custom bridging
+/// implementation comes in handy for those cases. See the `WordPressComRestApiEndpointError` extension below.
+extension WordPressAPIError: CustomNSError {
```
mokagio, line 11: TIL about [`CustomNSError`](https://developer.apple.com/documentation/foundation/customnserror)

mokagio (dismissed):

WordPressKit/WordPressComRestApi.swift
```diff
@@ -546,20 +546,20 @@ extension WordPressComRestApi {
...
         }
-        var userInfo = [String: Any]()
-        userInfo[WordPressComRestApi.ErrorKeyErrorCode] = "too_many_requests"
-        userInfo[WordPressComRestApi.ErrorKeyErrorMessage] = NSLocalizedString("Limit reached. You can try again in 1 minute. Trying again before that will only increase the time you have to wait before the ban is lifted. If you think this is in error, contact support.", comment: "Message to show when a request for a WP.com API endpoint is throttled")
-        userInfo[NSLocalizedDescriptionKey] = userInfo[WordPressComRestApi.ErrorKeyErrorMessage]
-        let nsError = WordPressComRestApiError.tooManyRequests as NSError
-        let errorWithLocalizedMessage = NSError(domain: nsError.domain, code: nsError.code, userInfo: userInfo)
-        return errorWithLocalizedMessage
+
+        let message = NSLocalizedString("Limit reached. You can try again in 1 minute. Trying again before that will only increase the time you have to wait before the ban is lifted. If you think this is in error, contact support.", comment: "Message to show when a request for a WP.com API endpoint is throttled")
```
mokagio: Good occasion to add key as well? I mean, I understand this is a **refactor** so behavior should not change, but I think we can squeeze this in:

```suggestion
        let message = NSLocalizedString(
            "wordpresskit.api.message.endpoint_throttled",
            value: "Limit reached. You can try again in 1 minute. Trying again before that will only increase the time you have to wait before the ban is lifted. If you think this is in error, contact support.",
            comment: "Message to show when a request for a WP.com API endpoint is throttled"
)
```

WordPressKit/WordPressComRestApi.swift
```diff
@@ -357,8 +357,8 @@ open class WordPressComRestApi: NSObject {
                         progress.completedUnitCount = progress.totalUnitCount
                         success(responseObject as AnyObject, response.response)
                     case .failure(let error):
-                        let nserror = self.processError(response: response, originalError: error)
-                        failure(nserror, response.response)
+                        let nserror = self.processError(response: response, originalError: error).flatMap { $0 as NSError }
+                        failure(nserror ?? (error as NSError), response.response)
```
mokagio: The original naming of the value made me feel confused when reading `(error as NSError)`.

What do you think of `processedError`

```suggestion
                        let processedError = self.processError(response: response, originalError: error).flatMap { $0 as NSError }
                        failure(processedError ?? (error as NSError), response.response)
```

mokagio (approved):
Left a couple of code suggestions, nothing blocking.

WordPressKit/WordPressComRestApi.swift
crazytonyli: Code squeezed. https://github.com/wordpress-mobile/WordPressKit-iOS/pull/696/commits/3cc7a80506a7cf9cacf21a726d9ce535c031a409

kean:
It's a breaking change for WordPressAuthenticator. I suggest adding a deprecation for backward compatibility:

```
@available(*, deprecated, renamed: "WordPressComRestApiErrorCode", message: "`WordPressComRestApiError` is renamed to `WordPressRestApiErrorCode`, and no longer conforms to `Swift.Error`")
public typealias WordPressComRestApiError = WordPressComRestApiErrorCode
```

I can do that in the scope of https://github.com/wordpress-mobile/WordPressKit-iOS/pull/704

crazytonyli:
> It's a breaking change for WordPressAuthenticator.

Yes indeed. As noted in the changelog file, this PR introduces a breaking change.

> I suggest adding a deprecation for backward compatibility:

The new typealias still introduces a breaking change, because `WordPressComRestApiErrorCode` no longer conforms to `Swift.Error`.

BTW, I have made some changes (https://github.com/wordpress-mobile/WordPress-iOS/pull/22434, https://github.com/woocommerce/woocommerce-ios/pull/11755, and https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/829) to adopt these breaking changes. They are still in draft, because we haven't released a new WordPressKit version yet. But, if these breaking changes are blocking you, I can look into releasing new versions of the libraries?

PR #702, opened 2024-01-24
Get the original url path using percentEncodedPath instead of path

crazytonyli:
### Description

This PR fixes an issue in `HTTPRequestBuilder` where the url-encoded part in a path is url-decoded in the final request. For example, `HTTPRequestBuilder(url: "https://wp.org/foo%2Fbar").build()` (%2F is url-encoded '/') returns a request with URL `https://wp.org/foo/bar`.

### Testing Details

See the added unit test.

---


jkmassel (approved):
This is a really great clarification!

PR #708, opened 2024-01-29
Minor changes to prepare for using URLSession directly in WordPressComRestApi

crazytonyli:
### Description

What says in the title. This PR doesn't change the library much. There will be bigger changes upcoming. I'm splitting them into separate PRs for easier code review.

### Testing Details

None. All should be covered by unit tests.

---

### DIFF
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -256,9 +256,10 @@ private extension Dictionary where Key == String, Value == Any {
     }
 
     func flatten() -> [URLQueryItem] {
-        reduce(into: []) { result, entry in
-            Self.urlEncode(into: &result, name: entry.key, value: entry.value)
-        }
+        sorted { $0.key < $1.key }
+            .reduce(into: []) { result, entry in
+                Self.urlEncode(into: &result, name: entry.key, value: entry.value)
+            }
     }
 
 }
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -18,7 +18,7 @@ final class HTTPRequestBuilder {
     }
 
     private let original: URLComponents
-    private var method: Method = .get
+    private(set) var method: Method = .get
     private var appendedPath: String = ""
     private var headers: [String: String] = [:]
     private var defaultQuery: [URLQueryItem] = []
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -171,14 +171,17 @@ open class WordPressComRestApi: NSObject {
     }
 
     deinit {
-        sessionManager.session.finishTasksAndInvalidate()
-        uploadSessionManager.session.finishTasksAndInvalidate()
+        for session in [urlSession, sessionManager.session, uploadSessionManager.session] {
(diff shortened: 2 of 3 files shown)


WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -256,9 +256,10 @@ private extension Dictionary where Key == String, Value == Any {
     }
 
     func flatten() -> [URLQueryItem] {
-        reduce(into: []) { result, entry in
-            Self.urlEncode(into: &result, name: entry.key, value: entry.value)
-        }
+        sorted { $0.key < $1.key }
+            .reduce(into: []) { result, entry in
+                Self.urlEncode(into: &result, name: entry.key, value: entry.value)
+            }
```
crazytonyli, line 262: There are unit tests that stubs API requests with certain query string like "foo=1&bar=2". This change here makes sure given the same dictionary `[String: Any]`, the encoded query will always be the same, so that the query lookup in unit tests can pass reliably.

WordPressKit/WordPressComRestApi.swift
```diff
@@ -171,23 +171,27 @@ open class WordPressComRestApi: NSObject {
...
     /**
      Cancels all ongoing taks and makes the session invalid so the object will not fullfil any more request
      */
     @objc open func invalidateAndCancelTasks() {
-        sessionManager.session.invalidateAndCancel()
-        uploadSessionManager.session.invalidateAndCancel()
+        for session in [urlSession, sessionManager.session, uploadSessionManager.session] {
+            session.invalidateAndCancel()
+        }
     }
```
crazytonyli, line 195: I don't see any reason to not include all URLSession instances in these generic methods above.

WordPressKit/HTTPRequestBuilder.swift
jkmassel: This is a tricky one, because it's potentially a breaking change (what happens if I add two `URLQueryItem`s with the same key and different values – I'd expect newer values to overwrite older ones right now – do we still have that guarantee now?)

I'd say the tests should be reworked for this particular case, but WDYT?

jkmassel (changes requested):
One thing I wanted to check on

WordPressKit/HTTPRequestBuilder.swift
crazytonyli: This `flatten` method is on `[String: Any]` type. So, it's not possible to have two query items that has the same key.

But the overriding problem you mentioned does exist in other cases, for example adding query one by one. Those cases are handled by [the adding query items APIs](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/cb5f349e92110e40a6c9cdc11d3da8cd0f6c71ad/WordPressKit/HTTPRequestBuilder.swift#L62), which has a `override` parameter—the callers decides the overriding behaviour. And they are covered by existing unit tests. Are there more tests you think I should add here?

https://github.com/wordpress-mobile/WordPressKit-iOS/blob/cb5f349e92110e40a6c9cdc11d3da8cd0f6c71ad/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift#L108-L126

jkmassel: Sounds like we're covered then, thanks!!

jkmassel (approved):

PR #690, opened 2024-01-16
Add more unit tests in preparation for an upcoming refactor

crazytonyli:
### Description

What says in the title. 😄 

---

- [x] Please check here if your pull request includes additional test coverage.
- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #709, opened 2024-01-29
Decouple WP.com REST API error handling from Alamofire

crazytonyli:
### Description

> [!Note]
> This PR is built on top of #708

The `processError` function specializes some `AFError` into `WordPressComRestApiEndpointError`. I'm going to use the same function to handle errors from `URLSession`. This PR extract a new `processError` based on current function, which will be used by both Alamofire and URLSession.

### Testing Details

The changes should be covered by unit tests. Let me know if you think there are more unit tests should be added.

---

### DIFF
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -484,10 +484,21 @@ extension WordPressComRestApi {
 
     /// A custom error processor to handle error responses when status codes are betwen 400 and 500
     func processError(response: DataResponse<Any>, originalError: Error) -> WordPressComRestApiEndpointError? {
-        guard let afError = originalError as?  AFError, case AFError.responseValidationFailed(_) = afError, let httpResponse = response.response, (400...500).contains(httpResponse.statusCode), let data = response.data else {
-            if let afError = originalError as? AFError, case AFError.responseSerializationFailed(_) = afError {
-                return .init(code: .responseSerializationFailed)
-            }
+        if let afError = originalError as? AFError, case AFError.responseSerializationFailed(_) = afError {
+            return .init(code: .responseSerializationFailed)
+        }
+
+        guard let httpResponse = response.response, let data = response.data else {
+            return nil
+        }
+
+        return processError(response: httpResponse, body: data, additionalUserInfo: (originalError as NSError).userInfo)
+    }
+
+    func processError(response httpResponse: HTTPURLResponse, body data: Data, additionalUserInfo: [String: Any]?) -> WordPressComRestApiEndpointError? {
+        // Not sure if it's intentional to include 500 status code, but the code seems to be there from the very beginning.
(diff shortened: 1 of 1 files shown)


WordPressKit/WordPressComRestApi.swift
```diff
@@ -528,10 +539,13 @@ extension WordPressComRestApi {
 
         let mappedError = errorsMap[errorCode] ?? .unknown
         if mappedError == .invalidToken {
-            invalidTokenHandler?()
+            // Call `invalidTokenHandler in the main thread since it's typically used by the apps to present an authentication UI.
```
jkmassel, line 542: Oof – good to preserve existing behaviour, but sad to add this dispatch here.

Do the apps care about the internal state of this object? Just wondering if we should use `DispatchQueue.main.sync` to wait for whatever handler to run before we carry on?

WordPressKit/WordPressComRestApi.swift
```diff
@@ -484,10 +484,21 @@ extension WordPressComRestApi {
 
     /// A custom error processor to handle error responses when status codes are betwen 400 and 500
     func processError(response: DataResponse<Any>, originalError: Error) -> WordPressComRestApiEndpointError? {
```
jkmassel, line 486: Do we have test coverage on this? It looks the same, but this seems like a good opportunity to add some, if not!

jkmassel (changes requested):

WordPressKit/WordPressComRestApi.swift
crazytonyli: Yes, they are covered. I have pre-added a few unit tests to expand the test coverage before these real library code changes (here is an example: #690).

Below is a screenshot of test coverage from Xcode. The only uncovered code is `?? [:]` in the red line.
Screenshot 2024-01-30 at 11 01 43 AM

WordPressKit/WordPressComRestApi.swift
crazytonyli: When using Alamofire, this `processError` function is called from the main thread. That means, upon receiving an error, the `invalidTokenHandler` from the app will be called from the main thread, and later the specific app feature that sends the API request will receive a failure result from the main thread. And these two calls are probably called in the same runloop.

When using URLSession, this `processError` function is called from a background thread. That means, the `invalidTokenHandler` will be called asynchronously from the main thread, whilst the `processError` continues to run in the background which eventually calls the specific app feature code's failure block from the main thread. In this scenario, these two calls are definitely not called in the same runloop. But the `invalidTokenHandler` _should_ (considering it's dispatched to the main queue first) be called from the app feature code.

So, I think we don't need to wait for `invalidTokenHandler` to be called here.

jkmassel: If it matches the existing behaviour, then fair enough – it still feels a little fragile IMHO(?) but that's for another day

jkmassel (approved):

PR #714, opened 2024-01-30
Update XMLRPC API progress from the main thread

crazytonyli:
### Description

WordPress iOS app (the uploading media part specifically) observes API requests progress and update UI upon progress updates. Updating progress from a background thread causes issues on the app side.

A more proper fix is probably updating the app's code to ensure UI updates happen on the main thread. But it's a bit difficult to find _all_ places that observe the XMLRPC progress instances. So, I decided to change the progress updating logic here, which kind of makes sense: now we have an implicit API that all progress updates from WordPressKit should happen on the main thread.

### Testing Details

You can produce this updating UI from a background thread issue on the WordPress iOS app.
1. Launch the app from Xcode, so that you can catch the issue from Xcode debugger.
2. Sign in to a self hosted WordPress site.
3. Go to Media screen and upload a media.
4. Upon finishing uploading, Xcode debugger should stop at [this observer code](https://github.com/wordpress-mobile/WordPress-iOS/blob/1b0c0a46ae2e3ed681df62e68dbc72030e357199/WordPress/Classes/ViewRelated/Media/SiteMedia/Views/SiteMediaCollectionCellViewModel.swift#L167-L169).

---

### DIFF
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -155,14 +155,15 @@ open class WordPressOrgXMLRPCApi: NSObject {
                 progress.totalUnitCount = requestProgress.totalUnitCount + 1
                 progress.completedUnitCount = requestProgress.completedUnitCount
             }.response(queue: DispatchQueue.global()) { (response) in
-                progress.completedUnitCount = progress.totalUnitCount
                 do {
                     let responseObject = try self.handleResponseWithData(response.data, urlResponse: response.response, error: response.error as NSError?)
                     DispatchQueue.main.async {
+                        progress.completedUnitCount = progress.totalUnitCount
                         success(responseObject, response.response)
                     }
                 } catch let error as NSError {
                     DispatchQueue.main.async {
+                        progress.completedUnitCount = progress.totalUnitCount
                         failure(error, response.response)
                     }
                     return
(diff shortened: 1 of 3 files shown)


WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -208,12 +209,13 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
                     do {
                         let responseObject = try self.handleResponseWithData(response.data, urlResponse: response.response, error: response.error as NSError?)
-                        progress.completedUnitCount = progress.totalUnitCount
                         DispatchQueue.main.async {
+                            progress.completedUnitCount = progress.totalUnitCount
                             success(responseObject, response.response)
                         }
                     } catch let error as NSError {
                         DispatchQueue.main.async {
+                            progress.completedUnitCount = progress.totalUnitCount
```
crazytonyli, line 218: Other changes are all moving an existing code from a background queue to the main queue. This line is new, but I think it's necessary—marking the progress as completed—to be consistent with the success scenario and the `callMethod` function above.

mokagio: A bonus bug fix 😄

mokagio (approved):
The need for this kind of code makes one appreciate the value of structured concurrency. _One day..._

crazytonyli:
> The need for this kind of code makes one appreciate the value of structured concurrency. One day...

That day will come, pretty soon... 😆

PR #715, opened 2024-01-30
Add asserts for debugging flaky LoadMediaLibraryTests

crazytonyli:
### Description

`LoadMediaLibraryRPCTests` occasionally fails on CI, but never locally fro me. I can't find any useful information in the xcresult file from CI jobs. This PR adds more assertions, which hopefully will help us debugging the failure.

### Testing Details

None.

---

### DIFF
--- a/WordPressKitTests/MediaLibraryTestSupport.swift
+++ b/WordPressKitTests/MediaLibraryTestSupport.swift
@@ -95,13 +95,15 @@ extension MediaLibraryTestSupport {
         } else if let body = request.httpBody {
             parser = XMLParser(data: body)
         } else {
+            XCTFail("The request doesn't have a request body: \(request)")
             return .init(error: URLError(.cannotDecodeContentData))
         }
 
         let delegate = RequestParser()
         parser.delegate = delegate
         guard parser.parse() else {
-            return .init(data: Data(), statusCode: 200, headers: nil)
+            XCTFail("Unexpected error: \(String(describing: parser.parserError))")
+            return .init(error: URLError(.cannotDecodeContentData))
         }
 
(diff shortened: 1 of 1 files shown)


WordPressKitTests/MediaLibraryTestSupport.swift
```diff
@@ -95,13 +95,15 @@ extension MediaLibraryTestSupport {
...
+            XCTFail("The request doesn't have a request body: \(request)")
             return .init(error: URLError(.cannotDecodeContentData))
         }
 
         let delegate = RequestParser()
         parser.delegate = delegate
         guard parser.parse() else {
-            return .init(data: Data(), statusCode: 200, headers: nil)
+            XCTFail("Unexpected error: \(String(describing: parser.parserError))")
+            return .init(error: URLError(.cannotDecodeContentData))
```
mokagio, line 106: Just wanted to double check going from returning a 200 with empty `Data()` to a `.cannotDecodeContentData` error was intentional here?

It makes sense to me, given `parser.parse()` failed, but it is a deviation from the previous behavior. 

The tests passed, so either this does not matter on the outside, or this branch was never used in the tests.

mokagio (approved):
Approving to unblock under the assumption my guess on the intentionality of the changes is correct.

WordPressKitTests/MediaLibraryTestSupport.swift
crazytonyli: > ..., or this branch was never used in the tests.

That's my hope. When everything is correct, this branch is never used. However, I wonder if the flakiness is caused by this branch: the parsing fails and an empty data is returned. That's why I added an `XCTFail` here to surface the error if this branch ever gets used.

PR #716, opened 2024-01-30
Add XMLRPC request support to HTTPRequestBuilder

crazytonyli:
### Description

What says in the title. Plus, this PR extra kind of unifies building multipart form and XMLPRC request, because their requests are very similar, just have different kinds of content.

### Testing Details

See the added unit tests.

---

### DIFF
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -1,4 +1,5 @@
 import Foundation
+import wpxmlrpc
 
 /// A builder type that appends HTTP request data to a URL.
 ///
@@ -25,6 +26,7 @@ final class HTTPRequestBuilder {
     private var appendedQuery: [URLQueryItem] = []
     private var bodyBuilder: ((inout URLRequest) throws -> Void)?
     private(set) var multipartForm: [MultipartFormField]?
+    private(set) var xmlrpcRequest: XMLRPCRequest?
 
     init(url: URL) {
         assert(url.scheme == "http" || url.scheme == "https")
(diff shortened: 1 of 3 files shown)


WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -174,30 +178,48 @@ final class HTTPRequestBuilder {
...
+            if let fileSize = fileSize as? NSNumber {
+                request.setValue(fileSize.stringValue, forHTTPHeaderField: "Content-Length")
+            }
+
+            return .right(fileURL)
+        } else {
+            let data = try encoder.dataEncoded()
+            request.setValue("\(data.count)", forHTTPHeaderField: "Content-Length")
+            return try .left(encoder.dataEncoded())
+        }
```
crazytonyli, line 215: This implementation is the same as the existing one.

https://github.com/wordpress-mobile/WordPressKit-iOS/blob/ce6b99a0d32c2d651a85cf64227aafaef5532fff/WordPressKit/WordPressOrgXMLRPCApi.swift#L232-L255

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -156,13 +158,15 @@ final class HTTPRequestBuilder {
...
-                request.httpBodyStream = InputStream(url: url)
+        if encodeBody {
+            let body = try encodeMultipartForm(request: &request, forceWriteToFile: false) ?? encodeXMLRPC(request: &request, forceWriteToFile: false)
+            if let body {
+                switch body {
+                case let .left(data):
+                    request.httpBody = data
+                case let .right(url):
+                    request.httpBodyStream = InputStream(url: url)
+                }
```
mokagio, line 169: Maybe a bit too compact, but what do you think of 

```suggestion
        if encodeBody, let body = try encodeMultipartForm(request: &request, forceWriteToFile: false) ?? encodeXMLRPC(request: &request, forceWriteToFile: false) {
            switch body {
            case let .left(data):
                request.httpBody = data
            case let .right(url):
                request.httpBodyStream = InputStream(url: url)
            }
```

mokagio (approved):

WordPressKit/HTTPRequestBuilder.swift
crazytonyli: I tried that, but it looks like many things are squeezed together, so I went with current version that has a few more lines.

PR #717, opened 2024-01-31
Add URLSession instances to WordPressXMLRPCApi

crazytonyli:
### Description

What says in the title. The new URLSessions will be used to send request in upcoming PRs.

### Testing Details

None.

---

### DIFF
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -90,7 +90,10 @@ open class WordPressOrgXMLRPCApi: NSObject {
         return URLSession(configuration: sessionConfiguration, delegate: sessionDelegate, delegateQueue: nil)
     }
 
+    // swiftlint:disable weak_delegate
+    /// `URLSessionDelegate` for the URLSession instances in this class.
     private let sessionDelegate = SessionDelegate()
+    // swiftlint:enable weak_delegate
 
     /// Creates a new API object to connect to the WordPress XMLRPC API for the specified endpoint.
     ///
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -7,6 +7,8 @@ open class WordPressOrgXMLRPCApi: NSObject {
     public typealias SuccessResponseBlock = (AnyObject, HTTPURLResponse?) -> Void
     public typealias FailureReponseBlock = (_ error: NSError, _ httpResponse: HTTPURLResponse?) -> Void
 
+    public static var useURLSession = false
+
     private let endpoint: URL
     private let userAgent: String?
     private var backgroundUploads: Bool
@@ -25,6 +27,14 @@ open class WordPressOrgXMLRPCApi: NSObject {
     /// Minimum WordPress.org Supported Version.
     ///
     @objc public static let minimumSupportedVersion = "4.0"
+
+    private lazy var urlSession: URLSession = makeSession(configuration: .default)
+    private lazy var uploadURLSession: URLSession = {
+        if self.backgroundUploads {
+            return makeSession(configuration: .background(withIdentifier: self.backgroundSessionIdentifier))
+        }
+        return urlSession
+    }()
 
     private var _sessionManager: Alamofire.SessionManager?
     private var sessionManager: Alamofire.SessionManager {
@@ -71,6 +81,15 @@ open class WordPressOrgXMLRPCApi: NSObject {
         return sessionManager
     }
 
+    private func makeSession(configuration sessionConfiguration: URLSessionConfiguration) -> URLSession {
+        var additionalHeaders: [String: AnyObject] = ["Accept-Encoding": "gzip, deflate" as AnyObject]
+        if let userAgent = self.userAgent {
+            additionalHeaders["User-Agent"] = userAgent as AnyObject?
+        }
+        sessionConfiguration.httpAdditionalHeaders = additionalHeaders
+        return URLSession(configuration: sessionConfiguration, delegate: sessionDelegate, delegateQueue: nil)
+    }
+
     private let sessionDelegate = SessionDelegate()
 
     /// Creates a new API object to connect to the WordPress XMLRPC API for the specified endpoint.
@@ -98,16 +117,18 @@ open class WordPressOrgXMLRPCApi: NSObject {
     }
 
     deinit {
-        _sessionManager?.session.finishTasksAndInvalidate()
-        _uploadSessionManager?.session.finishTasksAndInvalidate()
+        for session in [urlSession, uploadURLSession, sessionManager.session, uploadSessionManager.session] {
+            session.finishTasksAndInvalidate()
+        }
     }
 
     /**
      Cancels all ongoing and makes the session so the object will not fullfil any more request
      */
     @objc open func invalidateAndCancelTasks() {
-        sessionManager.session.invalidateAndCancel()
-        uploadSessionManager.session.invalidateAndCancel()
+        for session in [urlSession, uploadURLSession, sessionManager.session, uploadSessionManager.session] {
+            session.invalidateAndCancel()
+        }
     }
 
     // MARK: - Network requests
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -30,10 +30,9 @@ open class WordPressOrgXMLRPCApi: NSObject {
 
     private lazy var urlSession: URLSession = makeSession(configuration: .default)
     private lazy var uploadURLSession: URLSession = {
-        if self.backgroundUploads {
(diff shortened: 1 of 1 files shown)


WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -59,18 +69,32 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
+        let sessionDidReceiveChallengeWithCompletion: ((URLSession, URLAuthenticationChallenge, @escaping(URLSession.AuthChallengeDisposition, URLCredential?) -> Void) -> Void) = { [sessionDelegate] session, authenticationChallenge, completionHandler in
+            sessionDelegate.urlSession(session, didReceive: authenticationChallenge, completionHandler: completionHandler)
         }
         sessionManager.delegate.sessionDidReceiveChallengeWithCompletion = sessionDidReceiveChallengeWithCompletion
 
-        let  taskDidReceiveChallengeWithCompletion: ((URLSession, URLSessionTask, URLAuthenticationChallenge, @escaping(URLSession.AuthChallengeDisposition, URLCredential?) -> Void) -> Void) = { [weak self] session, task, authenticationChallenge, completionHandler in
-            self?.urlSession(session, task: task, didReceive: authenticationChallenge, completionHandler: completionHandler)
+        let  taskDidReceiveChallengeWithCompletion: ((URLSession, URLSessionTask, URLAuthenticationChallenge, @escaping(URLSession.AuthChallengeDisposition, URLCredential?) -> Void) -> Void) = { [sessionDelegate] session, task, authenticationChallenge, completionHandler in
```
mokagio: 🔍 

```suggestion
        let taskDidReceiveChallengeWithCompletion: ((URLSession, URLSessionTask, URLAuthenticationChallenge, @escaping(URLSession.AuthChallengeDisposition, URLCredential?) -> Void) -> Void) = { [sessionDelegate] session, task, authenticationChallenge, completionHandler in
```

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -329,7 +355,7 @@ open class WordPressOrgXMLRPCApi: NSObject {
     }
 }
 
-extension WordPressOrgXMLRPCApi {
+private class SessionDelegate: NSObject, URLSessionDelegate {
 
     @objc public func urlSession(_ session: URLSession,
```
mokagio: Interesting that we can declare the methods `public` in a `private` class 🤔

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -59,18 +69,32 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
+            additionalHeaders["User-Agent"] = userAgent as AnyObject?
+        }
+        sessionConfiguration.httpAdditionalHeaders = additionalHeaders
+        return URLSession(configuration: sessionConfiguration, delegate: sessionDelegate, delegateQueue: nil)
+    }
+
+    // swiftlint:disable weak_delegate
+    /// `URLSessionDelegate` for the URLSession instances in this class.
+    private let sessionDelegate = SessionDelegate()
+    // swiftlint:enable weak_delegate
```
mokagio, line 95: Seeing the annotation to disable `weak_delegate` made me wonder whether we are misusing `URLSessionDelegate`...

Nothing obvious jumps out to me. I see this retained instance used for the purpose of containing the `URLSession` management inside this API client. So, 👍 

But I thought I'd share the question here in case you have thoughts about it. I also understand this is an intermediate stage and that things will change in the future.

WordPressKit/WordPressOrgXMLRPCApi.swift
mokagio: For what is worth, I removed the `public` access level from the first method and Xcode didn't warn me.

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -26,6 +28,14 @@ open class WordPressOrgXMLRPCApi: NSObject {
     ///
     @objc public static let minimumSupportedVersion = "4.0"
 
+    private lazy var urlSession: URLSession = makeSession(configuration: .default)
+    private lazy var uploadURLSession: URLSession = {
+        if self.backgroundUploads {
+            return makeSession(configuration: .background(withIdentifier: self.backgroundSessionIdentifier))
```
mokagio: Nitpick: I tried locally, and it compiled without the explicit `self.`

```suggestion
        if backgroundUploads {
            return makeSession(configuration: .background(withIdentifier: backgroundSessionIdentifier))
```

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -26,6 +28,14 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
+        }
+        return urlSession
```
mokagio: Also a nitpick:

```suggestion
        guard backgroundUploads else {
            return urlSession
        }

        return makeSession(configuration: .background(withIdentifier: backgroundSessionIdentifier))
```

I can be pedantic about `guard` for early returns and I can see how the code is basically the same as before. But I find it to be a much clearer pattern.

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -96,16 +120,18 @@ open class WordPressOrgXMLRPCApi: NSObject {
     }
 
     deinit {
-        _sessionManager?.session.finishTasksAndInvalidate()
-        _uploadSessionManager?.session.finishTasksAndInvalidate()
+        for session in [urlSession, uploadURLSession, sessionManager.session, uploadSessionManager.session] {
+            session.finishTasksAndInvalidate()
```
mokagio, line 123: The decision to call [`finishTasksAndInvalidate`](https://developer.apple.com/documentation/foundation/urlsession/1407428-finishtasksandinvalidate) was here already, and I admit I wouldn't have thought about it if I hadn't seen the `invalidateAndCancel` call below, but I wonder how appropriate it is to let existing requests finish while/after the API client that started them is destroyed? 🤔 

Maybe this is a non-problem at runtime. Still... I can see the value for finishing requests, but at the same time I would assume that if an API client has been released, then the requests it was responsible for starting should be invalidated. ⚖️

mokagio (approved):
Looking forward to see this in action and actively replacing Alamofire

WordPressKit/WordPressOrgXMLRPCApi.swift
crazytonyli: Good catch. This `public` doesn't really make much sense. Removed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/717/commits/470594cf471e2ba90ab95f6aa2ab0c1c3ad9284a

WordPressKit/WordPressOrgXMLRPCApi.swift
crazytonyli: swiftlint throws a warning/error on this simply because it sees a variable with a type that has "Delegate" in it is not declared as a weak reference. It doesn't really know how the delegate is supposed to be used, or if there is a real retain cycle. In this case here, swiftlint reports a false positive. So, it's okay to disable the rule on this particular variable.

`URLSessionDelegate` is a weird one: [`URLSession` holds a strong reference to its delegate](https://developer.apple.com/documentation/foundation/urlsession#2934758). So, we can't make `WordPressOrgXMLRPCApi`, which holds a strong reference to `URLSession` instances, implements `URLSessionDelegate`.

Technically, we can declare this variable as a weak reference, which will put the responsiblity of not getting it deallocated to the `URLSession` instances. I used a strong reference here, simply because I don't see any harm in it, which IMO seems more secure because we do need it during the whole lifetime of `WordPressOrgXMLRPCApi`.

WordPressKit/WordPressOrgXMLRPCApi.swift
crazytonyli: I have changed this implementation to use ternary conditional operation which seems cleaner. https://github.com/wordpress-mobile/WordPressKit-iOS/pull/717/commits/b6506ceea28e1895d49f92777e8e32b0406f8890

mokagio: On nice. It also uses the implicit `return`. 👨‍🍳 💋

WordPressKit/WordPressOrgXMLRPCApi.swift
crazytonyli: I didn't look into these calls too much—the new code just continues to do what they used to do. 😄 

However, I just took a closer look into how HTTP requests are sent both in this class and the WP.com rest API one. I think it's impossible to be in a situation where this API client instance is deallocated whilst there are ongoing session tasks. Because [`self` is retained in API requests' completion handler](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/ea2d47e6a9723a2325cfd9d41c36357893ba0f63/WordPressKit/WordPressOrgXMLRPCApi.swift#L184).

mokagio: 👌

PR #719, opened 2024-02-05
Use URLSession to send XMLRPC requests

crazytonyli:
### Description

This PR updates the XMLRPC API client to send HTTP requests using URLSession, when `useURLSession` is set to true ([it's false by default](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/WordPressOrgXMLRPCApi.swift#L10)).

#### Error handling

This PR also changes the error handling in XMLPRC API and the API it surfaces.

When using Alamofire, in most cases, the XMLRPC API client returns `NSError` instances (plain and simple NSError instance, not the ones that are cast from Swift errors). It may return `URLError` too. It's quite different from the WP.com REST API client, which returns `NSError` instances that are cast from Swift errors, which is why I implemented [a custom NSError bridge logic](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/WordPressAPIError%2BNSErrorBrdige.swift).

Since there is no "bridging" happening in `WordPressOrgXMLRPCApi`, I didn't reuse the custom error bridging code. Instead, I re-used the existing function that creates new `NSError` instances, _for the exiting closure-based APIs_. The new async function `call(method:parameters:...)` returns `WordPressAPIError<WordPressOrgXMLRPCApiFault>` error instances. That means, depending on which method you call, you'll get different kinds of errors.

The new implementation is in the `decodeXMLRPCResult` and `asNSError` functions, which together should work exactly like [the existing `handleResponseWithData` function](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/WordPressOrgXMLRPCApi.swift#L292). And this "non-change" should be covered by [the existing error case unit tests](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/b0fd9e309b74594c8c5109cb007232b40aa040cd/WordPressKitTests/WordPressOrgXMLRPCApiTests.swift).

### Testing Details

Similar to https://github.com/wordpress-mobile/WordPressKit-iOS/pull/720, the unit tests pass when `useURLSession` is set to true.

https://github.com/wordpress-mobile/WordPress-iOS/pull/22540 can be used to get an app build which enables `useURLSession` by default.

---

### DIFF
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -282,6 +282,17 @@ open class WordPressOrgXMLRPCApi: NSObject {
         return progress
     }
 
+    /// Call an XMLRPC method.
+    ///
+    /// ## Error handling
+    ///
+    /// Unlike the closed-based APIs, this method returns a concrete error type. You should consider handle the errors
+    /// as they are, instead of casing them to `NSError` instance. But in case you do need to cast them to `NSError`,
+    /// considering using the `asNSError` function if you need backward compatiblity with existing code,
+    ///
+    /// - Parameters:
+    ///   - streaming: set to true if there are large data (i.e. uploading files) in given `parameters`. `false` by default.
+    /// - Returns: A `Result` type that contains the XMLRPC success or failure result.
     func call(method: String, parameters: [AnyObject]?, fulfilling progress: Progress, streaming: Bool = false) async -> WordPressAPIResult<HTTPAPIResponse<AnyObject>, WordPressOrgXMLRPCApiFault> {
         let session = streaming ? uploadURLSession : urlSession
         let builder = HTTPRequestBuilder(url: endpoint)
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -508,16 +508,21 @@ extension WordPressOrgXMLRPCApiError: LocalizedError {
 }
 
 public struct WordPressOrgXMLRPCApiFault: LocalizedError, HTTPURLResponseProviding {
-    var response: HTTPAPIResponse<Data>
+    public var response: HTTPAPIResponse<Data>
+    public let code: Int?
+    public let message: String?
 
-    var code: Int?
-    var message: String?
+    public init(response: HTTPAPIResponse<Data>, code: Int?, message: String?) {
+        self.response = response
+        self.code = code
+        self.message = message
+    }
 
     public var errorDescription: String? {
         message
     }
 
-    var httpResponse: HTTPURLResponse? {
+    public var httpResponse: HTTPURLResponse? {
         response.response
     }
 }
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -547,7 +547,7 @@ private extension WordPressAPIResult<HTTPAPIResponse<Data>, WordPressOrgXMLRPCAp
                 }
             }
 
-            if ["application/xml", "text/xml"].filter({ (type) -> Bool in return contentType.hasPrefix(type)}).count == 0 {
+            guard contentType.hasPrefix("application/xml") || contentType.hasPrefix("text/xml") else {
                 return .failure(.unparsableResponse(response: response.response, body: response.body, underlyingError: WordPressOrgXMLRPCApiError.unknown))
             }
 
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -286,12 +286,12 @@ open class WordPressOrgXMLRPCApi: NSObject {
     ///
     /// ## Error handling
     ///
-    /// Unlike the closed-based APIs, this method returns a concrete error type. You should consider handle the errors
-    /// as they are, instead of casing them to `NSError` instance. But in case you do need to cast them to `NSError`,
-    /// considering using the `asNSError` function if you need backward compatiblity with existing code,
+    /// Unlike the closure-based APIs, this method returns a concrete error type. You should consider handling the errors
+    /// as they are, instead of casting them to `NSError` instance. But in case you do need to cast them to `NSError`,
+    /// considering using the `asNSError` function if you need backward compatibility with existing code.
     ///
     /// - Parameters:
-    ///   - streaming: set to true if there are large data (i.e. uploading files) in given `parameters`. `false` by default.
+    ///   - streaming: set to `true` if there are large data (i.e. uploading files) in given `parameters`. `false` by default.
     /// - Returns: A `Result` type that contains the XMLRPC success or failure result.
     func call(method: String, parameters: [AnyObject]?, fulfilling progress: Progress, streaming: Bool = false) async -> WordPressAPIResult<HTTPAPIResponse<AnyObject>, WordPressOrgXMLRPCApiFault> {
         let session = streaming ? uploadURLSession : urlSession
--- a/WordPressKit/WordPressOrgXMLRPCApi.swift
+++ b/WordPressKit/WordPressOrgXMLRPCApi.swift
@@ -165,6 +165,20 @@ open class WordPressOrgXMLRPCApi: NSObject {
                            parameters: [AnyObject]?,
                            success: @escaping SuccessResponseBlock,
                            failure: @escaping FailureReponseBlock) -> Progress? {
+        guard !WordPressOrgXMLRPCApi.useURLSession else {
+            let progress = Progress.discreteProgress(totalUnitCount: 100)
+            Task { @MainActor in
+                let result = await self.call(method: method, parameters: parameters, fulfilling: progress, streaming: false)
+                switch result {
+                case let .success(response):
+                    success(response.body, response.response)
+                case let .failure(error):
+                    failure(error.asNSError(), error.response)
+                }
+            }
+            return progress
+        }
+
         // Encode request
         let request: URLRequest
         do {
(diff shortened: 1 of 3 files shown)


WordPressKitTests/WordPressOrgXMLRPCApiTests.swift
```diff
@@ -200,8 +200,16 @@ class WordPressOrgXMLRPCApiTests: XCTestCase {
                 XCTAssertEqual(error.domain, WPXMLRPCFaultErrorDomain)
                 // 403 is the 'faultCode' in the HTTP response xml.
                 XCTAssertEqual(error.code, 403)
-                XCTAssertNil(error.userInfo[WordPressOrgXMLRPCApi.WordPressOrgXMLRPCApiErrorKeyData as String])
                 XCTAssertNil(error.userInfo[WordPressOrgXMLRPCApi.WordPressOrgXMLRPCApiErrorKeyStatusCode as String])
+
+                // The change highlights one difference between the existing Alamofire-backed API and the new
+                // URLSession-backed API: the error returned by the new one may has HTTP response body which may not
+                // be the case exist in the old API. I think this is an acceptable change.
```
mokagio:
```suggestion
                // URLSession-backed API: the error returned by the new one has an HTTP response body which is not
                // the case in the old API. I think this is an acceptable change.
```

I think it's an acceptable change, too. In particular because the new way has more information than the old. If it was the other way around, then one would have to consider if the loss of information is dangerous for the client. But luckily that's not our case 😄

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -254,6 +282,33 @@ open class WordPressOrgXMLRPCApi: NSObject {
         return progress
     }
 
+    /// Call an XMLRPC method.
+    ///
+    /// ## Error handling
+    ///
+    /// Unlike the closed-based APIs, this method returns a concrete error type. You should consider handle the errors
+    /// as they are, instead of casing them to `NSError` instance. But in case you do need to cast them to `NSError`,
+    /// considering using the `asNSError` function if you need backward compatiblity with existing code,
```
mokagio:
```suggestion
    /// Unlike the closure-based APIs, this method returns a concrete error type. You should consider handling the errors
    /// as they are, instead of casting them to `NSError` instance. But in case you do need to cast them to `NSError`,
    /// considering using the `asNSError` function if you need backward compatibility with existing code.
```

_I only noticed the typo in compatibility when I loaded the text in this `textarea` in GitHub and the spell checker picked it up._

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -254,6 +282,33 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
+    ///
+    /// - Parameters:
+    ///   - streaming: set to true if there are large data (i.e. uploading files) in given `parameters`. `false` by default.
```
mokagio:
```suggestion
    ///   - streaming: set to `true` if there are large data (i.e. uploading files) in given `parameters`. `false` by default.
```

"false" at the end is code fenced, so "true" should be, too.

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -332,13 +387,13 @@ open class WordPressOrgXMLRPCApi: NSObject {
...
         let responseCode = statusCode == 403 ? 403 : error.code
         if let data = data {
-            var userInfo: [AnyHashable: Any] = error.userInfo
-            userInfo[type(of: self).WordPressOrgXMLRPCApiErrorKeyData] = data
-            userInfo[type(of: self).WordPressOrgXMLRPCApiErrorKeyDataString] = NSString(data: data, encoding: String.Encoding.utf8.rawValue)
-            userInfo[type(of: self).WordPressOrgXMLRPCApiErrorKeyStatusCode] = statusCode
+            var userInfo: [String: Any] = error.userInfo
+            userInfo[Self.WordPressOrgXMLRPCApiErrorKeyData as String] = data
+            userInfo[Self.WordPressOrgXMLRPCApiErrorKeyDataString as String] = NSString(data: data, encoding: String.Encoding.utf8.rawValue)
+            userInfo[Self.WordPressOrgXMLRPCApiErrorKeyStatusCode as String] = statusCode
```
mokagio, line 396: I was surprised by the `as String` so I drilled into the type definition. All those keys are `NSError.UserInfoKey` which is an `NSString`.

I wonder why Apple decided to define `userInfo` keys as `NSString` when other similar entities are `String`s:

image

mokagio: Side note: From the quick look I had at the code, I got the impression all usages ended up cast as `String`.

Unless I'm wrong, we could followup with a simplification PR that defines those keys as `String`.

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -451,3 +506,109 @@ extension WordPressOrgXMLRPCApiError: LocalizedError {
         }
     }
 }
+
+public struct WordPressOrgXMLRPCApiFault: LocalizedError, HTTPURLResponseProviding {
+    var response: HTTPAPIResponse<Data>
+
+    var code: Int?
+    var message: String?
```
mokagio: Do these need to be mutable?

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -451,3 +506,109 @@ extension WordPressOrgXMLRPCApiError: LocalizedError {
...
+                if let decoder = WPXMLRPCDecoder(data: response.body), decoder.isFault() {
+                    // when XML-RPC is disabled for authenticated calls (e.g. xmlrpc_enabled is false on WP.org),
+                    // it will return a valid fault payload with a non-200
+                    return .failure(.endpointError(.init(response: response, code: decoder.faultCode(), message: decoder.faultString())))
+                } else {
+                    return .failure(.unacceptableStatusCode(response: response.response, body: response.body))
+                }
+            }
+
+            if ["application/xml", "text/xml"].filter({ (type) -> Bool in return contentType.hasPrefix(type)}).count == 0 {
```
mokagio:
```suggestion
            if ["application/xml", "text/xml"].filter({ (type) -> Bool in return contentType.hasPrefix(type)}).isEmpty {
```

From SwiftLint, I learned that `isEmpty` is more efficient than `count == 0`. I'm pretty sure the difference is not something we would notice at runtime, but I still think it's a good practice to use `isEmpty`.

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -451,3 +506,109 @@ extension WordPressOrgXMLRPCApiError: LocalizedError {
...
+            guard !decoder.isFault() else {
+                return .failure(.endpointError(.init(response: response, code: decoder.faultCode(), message: decoder.faultString())))
+            }
+
+            if let decoderError = decoder.error() {
+                return .failure(.unparsableResponse(response: response.response, body: response.body, underlyingError: decoderError))
+            }
+
+            guard let responseXML = decoder.object() else {
+                return .failure(.unparsableResponse(response: response.response, body: response.body))
```
mokagio, line 567: Nitpick: The case were the decoding fails because the decoder has no object and the case where the decoder could not be created call `failure` with the same error.

I don't think this is a major problem because I doubt at runtime we'll ever run into a case where `decoder.error` is `nil` and so is `decoder.object`.

WordPressKit/WordPressOrgXMLRPCApi.swift
```diff
@@ -451,3 +506,109 @@ extension WordPressOrgXMLRPCApiError: LocalizedError {
...
+            data = nil
+            statusCode = nil
+        case let .connection(urlError):
+            error = urlError as NSError
+            data = nil
+            statusCode = nil
+        case let .endpointError(fault):
+            error = NSError(domain: WPXMLRPCFaultErrorDomain, code: fault.code ?? 0, userInfo: [NSLocalizedDescriptionKey: fault.message].compactMapValues { $0 })
+            data = fault.response.body
+            statusCode = nil
```
mokagio, line 601: Would this work?

```suggestion
            statusCode = fault.response.statusCode
```

mokagio (approved):
I thought about suggesting to duplicated the tests and have one version run with `useURLSession` `true` and the other with `false`.

This might be achieved without lots of copy-paste by running each test via something like `[true, false].each { useURLSession in ... }`.

But, I expect using `URLSession` will soon become the only way to use the API. In which case, we might as well save the time to modify the tests and move on with the Alamofire removal work.

WordPressKit/WordPressOrgXMLRPCApi.swift
crazytonyli: Depending on what you mean by "work" 😄 . It sure makes sense to me to provide HTTP status code when it's available. However, in [the existing implementation](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/WordPressOrgXMLRPCApi.swift#L309), only the status code is only set when it's 400 - 600. I'm just being super cautious to try keeping the existing behaviour changed when possible.

WordPressKit/WordPressOrgXMLRPCApi.swift
crazytonyli: 🙈 Thanks! Updated in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/719/commits/bcf8ea9e81da5b5471411cf25c879b6b252efc82

WordPressKit/WordPressOrgXMLRPCApi.swift
crazytonyli: Addressed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/719/commits/c4f376df1ede3867179c59c7b0a9fc4da7a79358

WordPressKit/WordPressOrgXMLRPCApi.swift
crazytonyli: I have simplified this code in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/719/commits/aeb9becfefd00339e339232c8bfd5d3605f854c2

WordPressKit/WordPressOrgXMLRPCApi.swift
mokagio: > Depending on what you mean by "work" 😄 .

Ahah. Good point @crazytonyli . Makes sense to be cautious 👍 

Besides, ideally the HTTP level details should remain at the networking library level, and the app should only need to check the typed errors it gets, without needing to look into HTTP error codes.

crazytonyli: > the app should only need to check the typed errors it gets, without needing to look into HTTP error codes.

Unfortunately that is not the case at the moment ([here is an example](https://github.com/search?q=%28repo%3Awordpress-mobile%2FWordPress-iOS+OR+repo%3Awordpress-mobile%2FWordPressAuthenticator-iOS%29+%22%3D%3D+403%22&type=code)), which is one of the reasons that I'd like to be cautious.

PR #720, opened 2024-02-05
Use URLSession to send WP.com REST API requests

crazytonyli:
### Description

This PR updates the WP.com REST API client to send HTTP requests using URLSession, when `useURLSession` is set to true (it's false by default).

> [!Note]
> I'd recommend reviewing this PR commit-by-commit.

### Testing Details

The unit test suite passes with `useURLSession` set to ture. I had to updated one test case to make it pass, because the new URLSession API does not return AFError.

In theory, all the new code in this PR are dead code at the moment. I will update the apps to enable `useURLSession` as in-development feature flag, and we can see how the new implementation work out in the app.

https://github.com/wordpress-mobile/WordPress-iOS/pull/22540 can be used to get an app build which enables `useURLSession` by default.

---

### DIFF
--- a/WordPressKit/WordPressComRestApi.swift
+++ b/WordPressKit/WordPressComRestApi.swift
@@ -301,7 +301,7 @@ open class WordPressComRestApi: NSObject {
             case let .success(response):
                 success(response.body, response.response)
             case let .failure(error):
-                failure(error as NSError, error.response)
+                failure(error.asNSError(), error.response)
             }
         }
 
@@ -360,7 +360,7 @@ open class WordPressComRestApi: NSObject {
             case let .success(response):
                success(response.body, response.response)
             case let .failure(error):
-               failure(error as NSError, error.response)
+               failure(error.asNSError(), error.response)
             }
         }
 
@@ -397,7 +397,7 @@ open class WordPressComRestApi: NSObject {
                 case let .success(response):
                     success(response.body, response.response)
                 case let .failure(error):
-                    failure(error as NSError, error.response)
+                    failure(error.asNSError(), error.response)
                 }
             }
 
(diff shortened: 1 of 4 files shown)


WordPressKit/WordPressComRestApi.swift
```diff
@@ -612,6 +639,38 @@ open class WordPressComRestApi: NSObject {
             }
     }
 
+    public func upload(
+        _ URLString: String,
```
mokagio: Nitpick. Value names should start with lowercase letters.

```suggestion
        _ urlString: String,
```

WordPressKit/WordPressComRestApi.swift
```diff
@@ -285,18 +288,47 @@ open class WordPressComRestApi: NSObject {
...
 
-        dataRequest(method: .get,
-                    urlString: URLString,
-                    parameters: parameters,
-                    encoding: URLEncoding.default,
-                    completion: completion)
+            completion(
+                result
+                    .map { ($0.body, $0.response) }
+                    .mapError { error -> Error in error }
```
mokagio, line 331: It wasn't obvious to me why this `mapError` was necessary.

What to you think of adding a line such as

```suggestion
                    // The completion expects a result with any Error.
                    // We need to "downcast" the WordPressComRestApiEndpointError error in the APIResult.
                    .mapError { error -> Error in error }
```

Also, using `$0` would work, too. But, having the explicit function signature with `Error` might be clearer for the reader in this unusual scenario.

mokagio (approved):
Only a couple of nitpicks.

WordPressKit/WordPressComRestApi.swift
crazytonyli: I have updated this part in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/720/commits/a3c19a9e07d601e9341cabe150bdfa2b2bc24e1b. which probably made it worse, in terms of the name... I used `URLString` because that's the term used throughout this class...

PR #721, opened 2024-02-07
Refactor BlazeService to use the new URLSession-backed API

crazytonyli:
### Description

I want to use the new async API in an existing code to showcase its difference compared to the existing closure-based API.

Please note, the updated function used to return `DecodingError` when fail to decode the response. It now returns `WordPressComRestApi.responseSerializationFailed`. I have checked the error handling code on the app side, which only checks if error is nil. So, it should be okay to make this change.

### Testing Details

There are [existing unit tests](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/b5c2bd39ec8f478dd53ecb7533ad025aad48f42e/WordPressKitTests/BlazeServiceRemoteTests.swift).

---

### DIFF
--- a/WordPressKit/BlazeServiceRemote.swift
+++ b/WordPressKit/BlazeServiceRemote.swift
@@ -12,20 +12,18 @@ open class BlazeServiceRemote: ServiceRemoteWordPressComREST {
     open func searchCampaigns(forSiteId siteId: Int, page: Int = 1, callback: @escaping (Result<BlazeCampaignsSearchResponse, Error>) -> Void) {
         let endpoint = "sites/\(siteId)/wordads/dsp/api/v1/search/campaigns/site/\(siteId)"
         let path = self.path(forEndpoint: endpoint, withVersion: ._2_0)
-        wordPressComRestApi.GET(path, parameters: [
-            "page": "\(page)" as AnyObject
-        ], success: { response, _ in
-            do {
-                let data = try JSONSerialization.data(withJSONObject: response)
-                let response = try JSONDecoder.apiDecoder.decode(BlazeCampaignsSearchResponse.self, from: data)
-                callback(.success((response)))
-            } catch {
-                WPKitLogError("Error parsing campaigns response: \(error), \(response)")
-                callback(.failure(error))
-            }
-        }, failure: { error, _ in
-            WPKitLogError("Error retrieving blaze campaigns: ", error)
-            callback(.failure(error))
-        })
+        Task { @MainActor in
+            let result = await self.wordPressComRestApi
+                .perform(
+                    .get,
+                    URLString: path,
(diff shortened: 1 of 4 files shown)


WordPressKit/WordPressComRestApi.swift
```diff
@@ -326,9 +326,7 @@ open class WordPressComRestApi: NSObject {
             completion(
                 result
                     .map { ($0.body, $0.response) }
-                    // The completion expects a result with any Error.
-                    // We need to "downcast" the WordPressComRestApiEndpointError error in the APIResult.
-                    .mapError { error -> Error in error }
+                    .eraseToError()
```
mokagio, line 329: I often leave this comment in code reviews arguing for replacing comments with code:

> From ["Clean Code" by Uncle Bob](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882):
>
> Screen Shot 2019-10-31 at 9 23 28 am

I love that you applied the observation I usually make _to the comment I suggested adding to the code_. 😄 🤓

WordPressKit/BlazeServiceRemote.swift
```diff
@@ -12,20 +12,18 @@ open class BlazeServiceRemote: ServiceRemoteWordPressComREST {
...
+                .perform(
+                    .get,
+                    URLString: path,
+                    parameters: ["page": page] as [String: AnyObject],
+                    jsonDecoder: JSONDecoder.apiDecoder,
+                    type: BlazeCampaignsSearchResponse.self
+                )
+                .map { $0.body }
+                .eraseToError()
+            callback(result)
```
mokagio, line 26: Thinking out loud. Given we'll be using this pattern a bunch, we could add a syntax sugar to write it like

```swift
            await wordPressComRestApi
                .perform(
                    .get,
                    URLString: path,
                    parameters: ["page": page] as [String: AnyObject],
                    jsonDecoder: JSONDecoder.apiDecoder,
                    type: BlazeCampaignsSearchResponse.self
                )
                .map { $0.body }
                .eraseToError()
                .forward(to: callback)
```

mokagio (approved):
Definitely an improvement in how the code looks, as well as the various benefits from dropping Alamofire and using more modern Swift.

> There are [existing unit tests](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/b5c2bd39ec8f478dd53ecb7533ad025aad48f42e/WordPressKitTests/BlazeServiceRemoteTests.swift).

Excellent choice for a refactor 🧠 👈

WordPressKit/BlazeServiceRemote.swift
crazytonyli: 👍 I'll make the change in a upcoming PR.

PR #722, opened 2024-02-12
Remove redundant JSON encoding and decoding

crazytonyli:
### Description

There are many places in the library that performs a redundant encoding and decoding JSON data. This PR refactor some of those cases so that we only parse JSON once.

> [!Note]
> I recommend reviewing this PR commit by commit. The changes in different files are very similar.

### Testing Details

All changes have existing test coverage.

---

### DIFF
--- a/WordPressKit/Result+Callback.swift
+++ b/WordPressKit/Result+Callback.swift
@@ -8,6 +8,10 @@ extension Swift.Result {
         }
     }
 
+    func execute(completion: (Self) -> Void) {
+        completion(self)
+    }
+
     func eraseToError() -> Result<Success, Error> {
         mapError { $0 }
     }
--- a/WordPressKit/Result+Callback.swift
+++ b/WordPressKit/Result+Callback.swift
@@ -8,7 +8,7 @@ extension Swift.Result {
         }
     }
 
-    func execute(completion: (Self) -> Void) {
+    func execute(_ completion: (Self) -> Void) {
         completion(self)
     }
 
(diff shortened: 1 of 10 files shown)


WordPressKit/Result+Callback.swift
```diff
@@ -8,6 +8,10 @@ extension Swift.Result {
         }
     }
 
+    func execute(completion: (Self) -> Void) {
+        completion(self)
+    }
```
crazytonyli, line 13: I used "execute", instead of "forward(to:)" suggested in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/721#discussion_r1482356171, to align with the existing `execute(onSuccess:onFailure:)` function.

mokagio: Fair 👍 

What do you think of removing the named parameter?

I feel the version without label reads better.

```swift
result.execute(completion: { print($0) })
result.execute({ print($0) })
```

WordPressKit/ShareAppContentServiceRemote.swift
```diff
@@ -11,20 +11,18 @@ open class ShareAppContentServiceRemote: ServiceRemoteWordPressComREST {
...
+                .perform(
+                    .get,
+                    URLString: requestURLString,
+                    parameters: params,
+                    jsonDecoder: .apiDecoder,
+                    type: RemoteShareAppContent.self
+                )
+                .map { $0.body }
+                .mapError { error -> Error in error.asNSError() }
+                .execute(completion: completion)
```
mokagio: Back to my earlier suggestion to remove `completion:` (I'm reviewing commit by commit), `.execute(completion)` seems to me tidier than `.execute(completion: completion)`.

mokagio (approved):

WordPressKit/ShareAppContentServiceRemote.swift
crazytonyli: Addressed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/722/commits/66e53921654ec77dca7de570d6dc9572aeb09d72.

mokagio: Thanks 😄

PR #723, opened 2024-02-12
Use a different GET function

crazytonyli:
### Description

This PR simply uses the `GET` function that takes success and failure callbacks, to replace the one that takes a completion callback.

The reason of this change is the completion block one comes from [WordPressRestApi](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/trunk/WordPressKit/WordPressRestApi.swift#L19), which will be removed as part of my `WordPressOrgRestApi` rewrite.

### Testing Details

None needed. These changes look pretty safe to me. 😄 

---

### DIFF
--- a/WordPressKit/ProductServiceRemote.swift
+++ b/WordPressKit/ProductServiceRemote.swift
@@ -43,10 +43,8 @@ open class ProductServiceRemote {
 
         serviceRemote.wordPressComRestApi.GET(
             path,
-            parameters: [:]) { result, _ in
-
-            switch result {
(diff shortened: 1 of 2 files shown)


mokagio (approved):
Agreed that the changes look safe enough 😄 

The code from `SharingServiceRemote` has unit tests 🎉 The code from `ProductServiceRemote` does not. I briefly considered adding them, then realized I'd have had to find a JSON stub and decided to continue with the work I already have going on.

PR #724, opened 2024-02-14
Rewrite `WordPressOrgRestApi`

crazytonyli:
> [!Note]
> This PR introduces breaking changes, and unfortunately can't be split into smaller ones (they won't be ship-able individually).
>
> I recommend you reviewing this PR commit by commit. However, when it comes to the commit which rewrites `WordPressOrgRestApi`, I'd suggest you ignore the commit diff and review [the new implementation](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/wordpress-org-rest-api-v2/WordPressKit/WordPressOrgRestApi.swift) directly.

## Description

Unlike the refactor to WP.com and XMLRPC API client, this PR rewrites `WordPressOrgRestApi` which deals with [.org REST API](https://developer.wordpress.org/rest-api/).

### About .org REST API

Every WordPress site ships with .org REST API, [which is accessible via `<site-url>/wp-json/<route>`](https://developer.wordpress.org/rest-api/using-the-rest-api/discovery/). The same .org REST API is also available on WordPress.com, but [under a different URL][Blog]: `https://public-api.wordpress.com/wp/v2/site/<site-id>/<route>`.

[Blog]: https://developer.wordpress.com/2016/11/11/wordpress-rest-api-on-wordpress-com/

### Current implementation

Currently, .org REST API (represented by [the `WordPressRestApi` protocol](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/WordPressRestApi.swift)) is implemented by two classes: [`WordPressOrgRestApi`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/WordPressOrgRestApi.swift) for self-hosted sites and [`WordPressComRestApi`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/WordPressComRestApi.swift#L702-L718) for WP.com sites.

As mentioned before, the same backend implementation is exposed by different URL path. The `WordPressRestApi` protocol has [a special `requestPath` function](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/WordPressRestApi.swift#L30) which the callers _have to_ remember to use ([here is an example](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/13.0.0/WordPressKit/BlockEditorSettingsServiceRemote.swift#L17)) before sending HTTP requests. IMO, forgetting to call that special function is an easy mistake to make during development.

Also, [the WP.com implementation of .org REST API](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/7a0217a5b9b3e4f0fa3f7197c06b833f7b23c433/WordPressKit/WordPressComRestApi.swift#L705-L709) reuses the error handling code in `rest/vx.x`. If I understand it correctly, the `rest/` endpoints are entirely different to the endpoints under `wp/v2` and there is no guarantee errors are reported in the same HTTP response format.

### New implementation

The PR rewrites `WordPressOrgRestApi` to be one single implementation which supports [self-hosted sites](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/wordpress-org-rest-api-v2/WordPressKit/WordPressOrgRestApi.swift#L57-L63) and [WordPress.com sites](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/wordpress-org-rest-api-v2/WordPressKit/WordPressOrgRestApi.swift#L53-L55).

The problem with different path now becomes [an internal implementation detail](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/wordpress-org-rest-api-v2/WordPressKit/WordPressOrgRestApi.swift#L249-L268) which callers don't need to worry about ([here is an example](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/wordpress-org-rest-api-v2/WordPressKit/BlockEditorSettingsServiceRemote.swift#L18)), mostly. When we need to support more .org REST API, [WordPressKit crashes with an error](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/wordpress-org-rest-api-v2/WordPressKit/WordPressOrgRestApi.swift#L257-L259) telling developers the exact changes they need to make to add such support.

[The error handling](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/wordpress-org-rest-api-v2/WordPressKit/WordPressOrgRestApi.swift#L3) is now the same for both self-hosted sites and WP.com sites, which makes sense because they are backed by the same backend implementation.

### Testing Details

I have added/updated unit tests in this PR. You can also use https://github.com/wordpress-mobile/WordPress-iOS/pull/22612 to do some manual tests on the app.

---


WordPressKit/BlockEditorSettingsServiceRemote.swift
```diff
@@ -1,8 +1,8 @@
 import Foundation
 
 public class BlockEditorSettingsServiceRemote {
-    let remoteAPI: WordPressRestApi
-    public init(remoteAPI: WordPressRestApi) {
+    let remoteAPI: WordPressOrgRestApi
+    public init(remoteAPI: WordPressOrgRestApi) {
```
mokagio, line 5: Here's a minor nitpick / curiosity. I have a thing for type names using acronyms, like `URL` instead of `Url` or `JSONDecoder` instead of `JsonDecoder`.

How would this look like if it were `WordPressORGRESTAPI`? Too many uppercase letters? :flushed:

WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -51,6 +51,23 @@ final class HTTPRequestBuilder {
         return self
     }
 
+    /// Append path and query to the original URL.
+    ///
+    /// Some may call API client using a string that contains path and query, like `api.get("post?id=1")`. This function
+    /// can be used to support those use cases.
```
mokagio: Formatting nitpick

```suggestion
    /// Some may call API client using a string that contains path and query, like `api.get("post?id=1")`.
    /// This function can be used to support those use cases.
```

WordPressKit/WordPressComRestApi.swift
```diff
@@ -484,11 +484,8 @@ open class WordPressComRestApi: NSObject {
     }
 
     private func requestBuilder(URLString: String) throws -> HTTPRequestBuilder {
-        guard let url = URL(string: URLString, relativeTo: baseURL) else {
-            throw URLError(.badURL)
-        }
-
-        var builder = HTTPRequestBuilder(url: url)
+        var builder = HTTPRequestBuilder(url: baseURL)
+            .appendURLString(URLString)
```
mokagio, line 488: Out of scope nitpick: value names [should start with a lowercase word](https://www.swift.org/documentation/api-design-guidelines/).

`URLString` -> `urlString`

image

WordPressKit/WordPressOrgRestApi.swift
```diff
@@ -1,105 +1,269 @@
...
-                    userInfo: [NSLocalizedDescriptionKey: NSLocalizedString("Failed to serialize request to the REST API.", comment: "Error message to show when wrong URL format is used to access the REST API")])
-            completion(.failure(error), nil)
-            return nil
+public final class WordPressOrgRestApi: NSObject {
+    public struct SelfHostedSiteCredential {
+        public var loginURL: URL
+        public var username: String
+        public var password: Secret<String>
+
+        public var adminURL: URL
```
mokagio: Do these need to be `var`?

mokagio:
```suggestion
        public let loginURL: URL
        public let username: String
        public let password: Secret<String>

        public let adminURL: URL
```

WordPressKit/WordPressOrgRestApi.swift
```diff
@@ -1,105 +1,269 @@
...
+    }
+
+    public func get<Success: Decodable>(
+        path: String,
+        parameters: [String: Any]? = nil,
+        jsonDecoder: JSONDecoder = JSONDecoder(),
+        type: Success.Type = Success.self
+    ) async -> WordPressAPIResult<Success, WordPressOrgRestApiError> {
+        await perform(.get, path: path, parameters: parameters, jsonDecoder: jsonDecoder, type: type)
+    }
```
mokagio, line 96: Inspired by `try JSONDecoder().decode(Resource.self, from: jsonData)`, what do you think of moving `type` at the start?

```swift
get(Post.type, path: "/post", parameters: nil)
```

WordPressKit/WordPressOrgRestApi.swift
```diff
@@ -1,105 +1,269 @@
...
+        options: JSONSerialization.ReadingOptions = []
+    ) async -> WordPressAPIResult<Any, WordPressOrgRestApiError> {
+        await perform(.get, path: path, parameters: parameters, options: options)
+    }
+
+    public func post<Success: Decodable>(
+        path: String,
+        parameters: [String: Any]? = nil,
+        jsonDecoder: JSONDecoder = JSONDecoder(),
+        type: Success.Type = Success.self
```
mokagio, line 110: Similar suggestion to move `type` as the first parameter as in the GET method.

WordPressKitTests/WordPressOrgRestApiTests.swift
```diff
@@ -22,98 +22,96 @@ class WordPressOrgRestApiTests: XCTestCase {
...
-            }
-        }
-        waitForExpectations(timeout: 2, handler: nil)
+        let response = try await api.post(path: "wp/v2/blocks/6", parameters: parameters, type: Response.self).get()
+        XCTAssertEqual(response.content.raw, blockContent, "The API should return the block")
     }
 
-    /// Verify that parameters in POST requests are sent as urlencoded form.
-    func testPostParametersContent() throws {
+        /// Verify that parameters in POST requests are sent as urlencoded form.
```
mokagio:
```suggestion
    /// Verify that parameters in POST requests are sent as urlencoded form.
```

WordPressKit/Plugin Management/SelfHostedPluginManagementClient.swift
```diff
@@ -7,89 +7,74 @@ public class SelfHostedPluginManagementClient: PluginManagementClient {
...
-        remote.request(method: .put, path: path, parameters: parameters) { (result, _) in
-            switch result {
-                case .success:
-                    success()
-
-                case .failure(let error):
-                    failure(error)
-            }
+        Task { @MainActor in
+            await remote.perform(.put, path: path, parameters: parameters, type: AnyResponse.self)
```
mokagio, line 34: Have you considered adding methods for `PUT` and other HTTP verbs, for consistency with the `POST` and `GET` implementations?

mokagio: Also, the `AnyResponse` type and related `.map { _ in }` took me a bit to understand.

I read them as a way to deal with an API which returns no data, or for which we don't care about the data.

What do you think of pushing this knowledge down at the `WordPressOrgRestApi` level?

We could have a `perform` implementation that returns `WordPressAPIResult<Void, WordPressOrgRestApiError>`, allowing us to segregate the `AnyResponse + .map { _ in }` "trick" lower into the implementation layer.

If instead it was implemented at this level because it's a one-off kind of thing, I'd still suggest centralizing the implementation into a `fileprivate extension WordPressOrgRestApi`

WordPressKit/Plugin Management/SelfHostedPluginManagementClient.swift
```diff
@@ -163,3 +137,26 @@ public class SelfHostedPluginManagementClient: PluginManagementClient {
...
 }
+
+private struct PluginStateResponse: Decodable {
+    enum CodingKeys: String, CodingKey {
+        case plugin = "plugin"
+        case status = "status"
+        case name = "name"
+        case author = "author"
+        case version = "version"
+        case pluginURI = "plugin_uri"
```
mokagio, line 148: _There's always one that gets in the way of the automatic key generation_ 😅

WordPressKitTests/SelfHostedPluginManagementClientTests.swift
```diff
@@ -0,0 +1,108 @@
...
+
+    override func setUp() {
+        super.setUp()
+
+        stub(condition: { _ in true }) {
+            XCTFail("Unexpected request: \($0)")
+            return HTTPStubsResponse(error: URLError(URLError.Code.networkConnectionLost))
+        }
+    }
+
```
mokagio, line 17: I believe we should also remove the stubs?

```suggestion
    override func tearDown() {
        super.tearDown()
        HTTPStubs.removeAllStubs()
    }
```

WordPressKit/BlockEditorSettingsServiceRemote.swift
```diff
@@ -43,30 +28,18 @@ public extension BlockEditorSettingsServiceRemote {
...
-        remoteAPI.GET(modifiedPath, parameters: parameters) { [weak self] (result, _) in
-            guard let `self` = self else { return }
-            switch result {
-            case .success(let response):
-                self.processBlockEditorSettingsResponse(response) { blockEditorSettings in
-                    completion(.success(blockEditorSettings))
+    func fetchBlockEditorSettings(completion: @escaping BlockEditorSettingsCompletionHandler) {
+        Task { @MainActor in
+            let result = await self.remoteAPI.get(path: "/wp-block-editor/v1/settings", parameters: ["context": "mobile"], type: RemoteBlockEditorSettings.self)
+                .map { settings -> RemoteBlockEditorSettings? in settings }
```
mokagio, line 34: This might be a change for later on, but it seems like the only reason the `RemoteBlockEditorSettings` the API call returns is wrapped in an `Optional` here is because that's the type the "legacy" `BlockEditorSettingsCompletionHandler` expects?

If so, and since we're making breaking changes, I wonder whether we could make the type in the completion typealias non-optional. It would remove this step, and I suspect also simplify caller code by removing the need to deal with `Result` _and_ `Optional`

WordPressKitTests/BlockEditorSettingsServiceRemoteTests.swift
```diff
@@ -171,13 +171,32 @@ extension BlockEditorSettingsServiceRemoteTests {
...
+    }
+
     func testFetchBlockEditorSettingsNotThemeJSON() {
         stub(condition: isHost("public-api.wordpress.com") && isPath("/wp-block-editor/v1/sites/1/settings") && containsQueryParams(["context": "mobile"])) { _ in
             fixture(filePath: OHPathForFile(self.blockSettingsNOTThemeJSONResponseFilename, Self.self)!, headers: ["Content-Type": "application/json"])
         }
 
         let waitExpectation = expectation(description: "Block Settings should be successfully fetched")
-        service.fetchBlockEditorSettings(forSiteID: siteID) { (response) in
+        service.fetchBlockEditorSettings { (response) in
```
mokagio: I'd expect SwiftLint to complain about the `()` at some point. But, it should also be something it can autocorrect easily 🤷‍♂️

mokagio (approved):
>  IMO, forgetting to call that special function is an easy mistake to make during development.

As a forgetful developer, I can attest to that. Definitely best to remove the need for people to have to remember to do things and have the code and design either do that for them or force them to do them.

> The problem with different path now becomes an internal implementation detail which callers don't need to worry about

Perfect 👌 

---

None of my comments are blocking.

WordPressKit/BlockEditorSettingsServiceRemote.swift
crazytonyli: I'd probably use `Org` instead of `ORG`. But this PR simply reuses the existing name 😄 .

WordPressKit/Plugin Management/SelfHostedPluginManagementClient.swift
crazytonyli: > Have you considered adding methods for PUT and other HTTP verbs, for consistency with the POST and GET implementations?

The post and get functions are convenience functions because the majority of HTTP requests uses these two methods. I didn't add convenience functions for other methods simply because they are not used by many places (maybe only one or two places?).

> What do you think of pushing this knowledge down at the WordPressOrgRestApi level?

IMO, It's kind of strange to ignore the response. Adding such support to `WordPressOrgRestApi` feels like we are encouraging it, but I don't think we should... Also, I believe this is the only place that ignores HTTP API response. I don't want to complicate `WordPressOrgRestApi` too much for this one case.

WordPressKit/BlockEditorSettingsServiceRemote.swift
crazytonyli: It's not just for `fetchBlockEditorSettings`'s syntax level backward compatibility, but most importantly runtime behaviour backward compatibility.

`/wp-block-editor/v1/settings` endpoint returns a JSON that contains many properties. The `fetchBlockEditorSettings` function only cares about a subset of these properties. See the model type [`RemoteBlockEditorSettings`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/trunk/WordPressKit/RemoteBlockEditorSettings.swift), which contains mostly non-optional properties.

If the JSON response contains all the properties declared in the Swift type, this function of course should return a model instance.

However, if this API returns 200 with valid "block editor settings" JSON response, which _does not_ contains all the non-optional proprieties declared in the `RemoteBlockEditorSettings` type, the function should return a nil, instead of a "decoding error".

That's the behaviour of the existing function. And the refactored implementation (mainly the line here and the code below that maps decoding error to a nil successful result) maintains the same runtime behaviour for backward compatibility.

WordPressKitTests/SelfHostedPluginManagementClientTests.swift
crazytonyli: Good catch! I always forget about it. Addressed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/724/commits/c44bb13e7c5551471274efe47ee59f1374f6f9e4.

WordPressKit/WordPressOrgRestApi.swift
crazytonyli: I feel like the decoding parameters are "secondary": The "GET /api/path" part is the main one. Plus, they have default values thus should be place at the end?

PR #725, opened 2024-02-15
Decouple `PluginDirectoryServiceRemote` from Alamofire

crazytonyli:
### Description

`PluginDirectoryServiceRemote` sends HTTP requests to `api.wordpress.org`, which we don't have a dedicated API client for. I used `URLSession` directly, instead of the customized `URLSession.perform` function, because the extra error handling implemented in the `perform` function is useless in `PluginDirectoryServiceRemote`.

The `Endpoint` abstraction is removed from this PR, because `PluginDirectoryServiceRemote`—the only type that implements `Endpoint` protocol—no longer uses it.

### Testing Details

See updated and added unit tests.

---


CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- Decouple `PluginDirectoryServiceRemote` from Alamofire. [#725]
```
mokagio, line 38:
```suggestion
- Decouple `PluginDirectoryServiceRemote` from Alamofire. [#725]
- Remove `Endpoint`. [#725]
```

`Endpoint` might have been unused, but it was still a `public` type. Do you think it would be worth to track it in the changelog?

mokagio (approved):

CHANGELOG.md
crazytonyli: Sure thing. Added in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/725/commits/dce379835de69e333a9748a481ab380c069fbd41.

PR #727, opened 2024-02-19
Use a path, not URL, in fetch posts from endpoint request

mokagio:
### Description

The recent work on the WordPress.com API requires GET, POST, etc methods to be called using a path, but this method passed a whole URL.

### Testing Details

The change didn't affect unit tests because there are no tests for this method 😢 

image

See how https://github.com/wordpress-mobile/WordPress-iOS/pull/22622 passes in CI:

image

I left a note about the Gutenberg Gallery block failure here: https://github.com/wordpress-mobile/WordPress-iOS/pull/22612#issuecomment-1951690269

---

### DIFF
--- a/WordPressKit/ReaderPostServiceRemote.m
+++ b/WordPressKit/ReaderPostServiceRemote.m
@@ -200,7 +200,7 @@ NSString * const ParamKeyMetaValue = @"site,feed";
                            success:(void (^)(NSArray<RemoteReaderPost *> *posts, NSString *algorithm))success
                            failure:(void (^)(NSError *))failure
 {
-    NSString *path = [endpoint absoluteString];
+    NSString *path = [endpoint path];
     [self.wordPressComRestApi GET:path
            parameters:params
               success:^(id responseObject, NSHTTPURLResponse *httpResponse) {


WordPressKit/ReaderPostServiceRemote.m
```diff
@@ -200,7 +200,7 @@ - (void)fetchPostsFromEndpoint:(NSURL *)endpoint
                            success:(void (^)(NSArray<RemoteReaderPost *> *posts, NSString *algorithm))success
                            failure:(void (^)(NSError *))failure
 {
-    NSString *path = [endpoint absoluteString];
+    NSString *path = [endpoint path];
```
mokagio, line 203: I wonder how many of these migration issues are left in the library, and what it would take to discover them on our side before they reach the users... 🤔 

I suppose one way would be to audit each `GET...` call. There are "only" 118 of them...

crazytonyli:
> The change didn't affect unit tests because there are no tests for this method 😢

I have added one in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/729.

WordPressKit/ReaderPostServiceRemote.m
crazytonyli: I haven't looked where the app gets `endpoint`. If they are constructed by the app, I'm thinking maybe we can dumb down this `fetchPostsFromEndpoint` function to not accepting a `URL` argument?

PR #731, opened 2024-02-20
Add `apiURL` parameter for changing WP.com API URL

crazytonyli:
### Description

This PR brings back support of mocking WP.com .org REST API. See https://github.com/wordpress-mobile/WordPress-iOS/pull/22612#pullrequestreview-1891508629 for context in the WordPress app.

### Testing Details

I have added one unit test. Let me know if you think there are more can be added.

---

### DIFF
--- a/WordPressKit/WordPressOrgRestApi.swift
+++ b/WordPressKit/WordPressOrgRestApi.swift
@@ -40,7 +40,7 @@ public final class WordPressOrgRestApi: NSObject {
     }
 
     enum Site {
-        case dotCom(siteID: UInt64, bearerToken: String)
+        case dotCom(siteID: UInt64, bearerToken: String, apiURL: URL)
         case selfHosted(apiURL: URL, credential: SelfHostedSiteCredential)
     }
 
@@ -49,8 +49,8 @@ public final class WordPressOrgRestApi: NSObject {
 
     var selfHostedSiteNonce: String?
 
-    public convenience init(dotComSiteID: UInt64, bearerToken: String, userAgent: String? = nil) {
-        self.init(site: .dotCom(siteID: dotComSiteID, bearerToken: bearerToken), userAgent: userAgent)
+    public convenience init(dotComSiteID: UInt64, bearerToken: String, userAgent: String? = nil, apiURL: URL = WordPressComRestApi.apiBaseURL) {
+        self.init(site: .dotCom(siteID: dotComSiteID, bearerToken: bearerToken, apiURL: apiURL), userAgent: userAgent)
     }
 
     public convenience init(selfHostedSiteWPJSONURL apiURL: URL, credential: SelfHostedSiteCredential, userAgent: String? = nil) {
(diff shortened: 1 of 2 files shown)


WordPressKit/WordPressOrgRestApi.swift
```diff
@@ -40,7 +40,7 @@ public final class WordPressOrgRestApi: NSObject {
     }
 
     enum Site {
-        case dotCom(siteID: UInt64, bearerToken: String)
+        case dotCom(siteID: UInt64, bearerToken: String, apiURL: URL)
         case selfHosted(apiURL: URL, credential: SelfHostedSiteCredential)
```
mokagio, line 44: Nitpick: What do you think of making `apiURL` the first associated value, so that it's in line with the  `selfHosted` case?

One argument against doing so would be that `siteID` is more important when it comes to identifying a .com site, so it belongs in first position. I'd be fine with that. This is just a style nitpick / question that I'm not sure on.

WordPressKit/WordPressOrgRestApi.swift
```diff
@@ -68,7 +68,7 @@ public final class WordPressOrgRestApi: NSObject {
         if let userAgent {
             additionalHeaders["User-Agent"] = userAgent
         }
-        if case let Site.dotCom(siteID: _, bearerToken: token) = site {
+        if case let Site.dotCom(siteID: _, bearerToken: token, _) = site {
```
mokagio: Nitpick: The new `_` entry made me realize that the `siteID:` label is unnecessary

```suggestion
        if case let Site.dotCom(_, bearerToken: token, _) = site {
```

---

Site note: I looked for a SwiftLint rule for this but did not find one that would have picked that up. At first, I browsed the [rules docs](https://realm.github.io/SwiftLint/rule-directory.html). Then, I tried with

```
cat WordPressKit/WordPressOrgRestApi.swift \
  | ./Pods/SwiftLint/swiftlint lint --quiet --enable-all-rules --use-stdin \
  | sort -v
```

WordPressKit/WordPressOrgRestApi.swift
```diff
@@ -203,8 +203,8 @@ public final class WordPressOrgRestApi: NSObject {
 private extension WordPressOrgRestApi {
     func apiBaseURL() -> URL {
         switch site {
-        case .dotCom:
-            return URL(string: "https://public-api.wordpress.com")!
+        case let .dotCom(_, _, apiURL):
+            return apiURL
```
mokagio, line 207: A nice side effect is that now we are one step closer to having a single source of truth for what this URL should be.

image

WordPressKitTests/WordPressOrgRestApiTests.swift
```diff
@@ -114,4 +126,10 @@ extension WordPressOrgRestApi {
     }
 }
 
+extension WordPressOrgRestApi.Site {
+    static func dotCom(siteID: UInt64, bearerToken: String) -> Self {
+        .dotCom(siteID: siteID, bearerToken: bearerToken, apiURL: WordPressComRestApi.apiBaseURL)
+    }
+}
```
mokagio, line 133: Note for future readers: This method looks unused in the diff, but it's actually a way to bridge the new implementation, which expects an `apiURL` parameter, with the existing code that did not have it.

See for example line 103 in this file.

mokagio (approved):

WordPressKit/WordPressOrgRestApi.swift
crazytonyli: I'd agree with that argument, considering 'apiURL' is an optional parameter in the initialiser. 😄

PR #733, opened 2024-02-21
Do not add default query if it exists in the original URL

crazytonyli:
### Description

While updating unit tests, I have caught an bug in the exiting implementation of appending default 'locale' query to all WP.com REST API call. The expected behaviour is in [`testThatLocaleIsNotAppendedIfAlreadyIncludedInPath`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/2e67d20e276223f2511642ef8e2ea9655fe9878a/WordPressKitTests/WordPressComRestApiTests%2BLocale.swift#L44-L55).

#### Details

`WordPressComRestApi` adds a `locale=en` (or whatever the device locale is) query to all HTTP requests, unless the 'local' is passed as parameter. For example:
1. `WordPressComRestApi.get("/post/1")` should send an HTTP request with the default 'locale=en' in the URL: `https://.../post/1?locale=en`.
1. `WordPressComRestApi.get("/post/1", parameters: ["locale": "zh"])` should send an HTTP request with `locale=zh`, instead of the default query "locale=en".

However, there is a case missed in the current implementation, `WordPressComRestApi.get("/post/1?locale=zh")` should behave like the second example above and send an HTTP request with `locale=zh`.

### Testing Details

The bug fix is covered in the existing `testThatLocaleIsNotAppendedIfAlreadyIncludedInPath` and I have added a new dedicate unit test for it.

---

### DIFF
--- a/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
+++ b/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
@@ -205,6 +205,15 @@ class HTTPRequestBuilderTests: XCTestCase {
         try XCTAssertEqual(builder.query(name: "foo", value: "bar").build().url?.query, "locale=zh&foo=bar")
     }
 
+    func testDefaultQueryExistsInOriginalURL() throws {
+        let url = try HTTPRequestBuilder(url: URL(string: "https://wordpress.org/hello?locale=foo")!)
+            .query(defaults: [URLQueryItem(name: "locale", value: "en")])
+            .build()
+            .url
+
+        XCTAssertEqual(url?.query, "locale=foo")
+    }
+
     func testJSONBody() throws {
         var request = try HTTPRequestBuilder(url: URL(string: "https://wordpress.org")!)
             .method(.post)
--- a/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
+++ b/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
@@ -205,7 +205,7 @@ class HTTPRequestBuilderTests: XCTestCase {
         try XCTAssertEqual(builder.query(name: "foo", value: "bar").build().url?.query, "locale=zh&foo=bar")
     }
 
-    func testDefaultQueryExistsInOriginalURL() throws {
+    func testDefaultQueryDoesNotOverriedQueryItemInOriginalURL() throws {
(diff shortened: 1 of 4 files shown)


WordPressKitTests/WordPressComRestApiTests+Locale.swift
```diff
@@ -1,272 +1,98 @@
...
 import WordPressShared
 @testable import WordPressKit
 
 extension WordPressComRestApiTests {
 
-    func testThatAppendingLocaleWorks() {
-        // Given
-        let path = "/path/path"
-        let preferredLanguageIdentifier = WordPressComLanguageDatabase().deviceLanguage.slug
+    func testThatAppendingLocaleWorks() async throws {
```
mokagio: What do you think of

```suggestion
    func testAddsLocaleToURLQueryByDefault() async throws {
```

WordPressKitTests/WordPressComRestApiTests+Locale.swift
```diff
@@ -1,272 +1,98 @@
...
-
-        let expectedQueryItemValue = preferredLanguageIdentifier
-        XCTAssertEqual(expectedQueryItemValue, actualQueryItemValue!)
+        let preferredLanguageIdentifier = WordPressComLanguageDatabase().deviceLanguage.slug
+        XCTAssertEqual(request?.url?.query, "locale=\(preferredLanguageIdentifier)")
     }
 
-    func testThatAppendingLocaleWorksWithExistingParams() {
-        // Given
+    func testThatAppendingLocaleWorksWithExistingParams() async throws {
```
mokagio: What do you think of

```suggestion
    func testAddsLocaleToURLQueryByDefaultAndMaintainsInputParameters() async throws {
```

WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
```diff
@@ -205,6 +205,15 @@ class HTTPRequestBuilderTests: XCTestCase {
         try XCTAssertEqual(builder.query(name: "foo", value: "bar").build().url?.query, "locale=zh&foo=bar")
     }
 
+    func testDefaultQueryExistsInOriginalURL() throws {
```
mokagio: What do you think of 

```suggestion
    func testDoesNotOverriedQueryItemInOriginalURL() throws {
```

mokagio (approved):

PR #734, opened 2024-02-21
Fix an XMLRPC uploading file issue 

### LINKED ISSUE #22659: URLSession feature flag leads to broken video uploads on self-hosted sites
### Expected behavior

Videos uploaded to self-hosted sites should function as expected.

### Actual behavior

With non-production builds, videos are currently being uploaded with an invalid format to self-hosted sites. This happened following the change in https://github.com/wordpress-mobile/WordPress-iOS/pull/22540, which defaulted to using URLSession in WordPressKit (instead of Alamofire).

An error like the following can be seen when attempting to view iOS-uploaded videos in the media library on the web: 

`Media error: Format(s) not supported or source(s) not found[Download](https://example.com.mp4?_=4)`

Additionally, when uploaded to the editor, the videos won't play back.

Note, this isn't an issue on dotcom or Jetpack-connected sites, only self-hosted sites with no Jetpack connection. 

* Set up a self-hosted site that _doesn't_ have the Jetpack plugin enabled.
* Download a non-production build of the iOS, such as the installable build generated with this PR.
* Navigate to the post/page editor for your self-hosted site.
* Add the video block and choose the option to upload a video from your device.
* Observe that the block's thumbnail is blank and tapping to play results in a blank player.
* On the web, navigate to the site's media library and view the following error alongside the uploaded video: `Media error: Format(s) not supported or source(s) not found[Download](https://example.com.mp4?_=4)`

The feature flag that enables URLSession in WordPressKit can be toggled off via _Me_ → _App Settings_ → _Feature Flags_ (note, I found it necessary to reboot the app for these changes to take effect). Toggling the flag off, however, doesn't fix the issue and leads to failed video uploads.

crazytonyli:
### Description

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/22659

### Testing Details

See the added unit tests.

---

### DIFF
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -228,7 +228,7 @@ final class HTTPRequestBuilder {
         } else {
             let data = try encoder.dataEncoded()
             request.setValue("\(data.count)", forHTTPHeaderField: "Content-Length")
-            return try .left(encoder.dataEncoded())
+            return .left(data)
         }
     }
 }
--- a/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
+++ b/WordPressKitTests/Utilities/HTTPRequestBuilderTests.swift
@@ -407,12 +407,14 @@ class HTTPRequestBuilderTests: XCTestCase {
         let file = try XCTUnwrap(Bundle(for: type(of: self)).url(forResource: "me-settings-success", withExtension: "json"))
         let fileContentBase64 = try Data(contentsOf: file).base64EncodedString()
         let fileStream = try XCTUnwrap(InputStream(url: file))
-
         let request = try HTTPRequestBuilder(url: URL(string: "https://w.org/xmlrpc.php")!)
             .method(.post)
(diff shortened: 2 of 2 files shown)


mokagio (approved):
Approving to unblock. I checked it out locally and saw that the test fails with the original code.

However, I don't understand how this works 🤔 I've been looking into the `WPXMLRPCEncoder` [method](https://github.com/wordpress-mobile/wpxmlrpc/blob/810695e0455396cfc5fad460f16f64eefd8b5087/WPXMLRPC/WPXMLRPCEncoder.m#L77-L88) and a bit down the chain and I don't see anything that would result in the second call (however unnecessary) to it returning a different value from the first 🤔

crazytonyli:
The root cause is at [the `encodingInputStream` function](https://github.com/wordpress-mobile/wpxmlrpc/blob/trunk/WPXMLRPC/WPBase64Utils.m#L307). The first call works as expected with full base64 encoded content. Since the first call reads the stream to its end, the second call start reading from the end, which of course returns nothing, and get an empty base64 encoded content.

mokagio:
Got it. Thanks for following up @crazytonyli

PR #736, opened 2024-02-23
Remove Alamofire as a library dependency

crazytonyli:
### Description

The apps has been running with `useURLSession` turned on for a couple of weeks. See https://github.com/wordpress-mobile/WordPress-iOS/pull/22540 and https://github.com/woocommerce/woocommerce-ios/pull/11927. There is no major issue reported. And it's likely we'll fix bugs in the new implementation, instead of reverting back to the old one.

This PR removes Alamofire as a dependency to the library, but it's still used in the unit tests to verify some encoding request encoding implementations, i.e. multipart-form.

### Testing Details

Changes are covered by unit tests.

---


mokagio (approved):
🎉 

> This PR removes Alamofire as a dependency to the library, but it's still used in the unit tests to verify some encoding request encoding implementations, i.e. multipart-form.

I hadn't considered the option of keeping it in the unit tests temporarily. It's a great idea, because it lets us focus on removing Alamofire from the production code faster. 👍

PR #739, opened 2024-03-03
Fix crash in background url session

crazytonyli:
### Description

I thought I tested background uploads from Jetpack share extension with [the "user URLSession" feature flag][PR] on. I turns out I didn't, because the feature flag only takes effect within the app, and doesn't change make the share extension to use URLSession instead of Alamofire.

#### Issue

The issue is pretty obvious once the feature flag is removed in https://github.com/wordpress-mobile/WordPress-iOS/pull/22707: the share extension would crash because `URLSession.uploadTask(with:fromFile:completion:)` crashes on background URLSession instances.

#### Fix

In the `perform(request:...)` async function, we rely on `URLSession` APIs which accept a completion handler to call the continuation API. As I mentioned above, those APIs are not available to background URLSession instances. That means we'll have to use a customized `URLSessionDelegate` to receive HTTP response and call the completion block created in `perform(request:...)`. Those are done in the new `BackgroundURLSessionDelegate` type.

Please note, this new delegate is only used in background URLSession instances. That means in theory this PR has no effect on the JP/WP iOS app process which uses non-background URLSession instances.

[PR]: https://github.com/wordpress-mobile/WordPress-iOS/pull/22540

### Testing Details

As I mentioned in the code comment, background URLSession doesn't work in current unit test set up, but I have added some simple code to ensure we can still test `BackgroundURLSessionDelegate` from unit tests.

~I'll open a PR on the WordPress-iOS repo, so that we can test its share extension, which uses this PR's changes.~
See the "Test Instructions" in https://github.com/wordpress-mobile/WordPress-iOS/pull/22755.

---

### DIFF
--- a/WordPressKit/HTTPClient.swift
+++ b/WordPressKit/HTTPClient.swift
@@ -22,6 +22,18 @@ extension HTTPAPIResponse where Body == Data {
 
 extension URLSession {
 
+    /// Create a background URLSession instance that can be used in the `perform(request:...)` async function.
+    ///
+    /// The `perform(request:...)` async function can be used in all non-background URLSession instances without any
+    /// extra work. However, there is a requirement to make the function works with with background URLSession instances.
+    /// That is the URLSession must have a delegate of `BackgroundURLSessionDelegate` type.
+    static func backgroundSession(configuration: URLSessionConfiguration) -> URLSession {
+        assert(configuration.identifier != nil)
+        // Pass `delegateQueue: nil` to get a serial queue, which is required to ensure thread safe access to
+        // `WordPressKitSessionDelegate` instances.
+        return URLSession(configuration: configuration, delegate: BackgroundURLSessionDelegate(), delegateQueue: nil)
+    }
+
     /// Send a HTTP request and return its response as a `WordPressAPIResult` instance.
     ///
     /// ## Progress Tracking and Cancellation
@@ -49,6 +61,10 @@ extension URLSession {
         fulfilling parentProgress: Progress? = nil,
         errorType: E.Type = E.self
     ) async -> WordPressAPIResult<HTTPAPIResponse<Data>, E> {
+        if configuration.identifier != nil {
+            assert(delegate is BackgroundURLSessionDelegate, "Unexpected URLSession delegate type. See the `backgroundSession(configuration:)`")
+        }
+
         if let parentProgress {
             assert(parentProgress.completedUnitCount == 0 && parentProgress.totalUnitCount > 0, "Invalid parent progress")
             assert(parentProgress.cancellationHandler == nil, "The progress instance's cancellationHandler property must be nil")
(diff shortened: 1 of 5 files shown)


WordPressKit/HTTPClient.swift
```diff
@@ -22,6 +22,18 @@ extension HTTPAPIResponse where Body == Data {
 
 extension URLSession {
 
+    /// Create a background URLSession instance that can be used in the `perform(request:...)` async function.
+    ///
+    /// The `perform(request:...)` async function can be used in all non-background URLSession instances without any
+    /// extra work. However, there is a requirement to make the function works with with background URLSession instances.
+    /// That is the URLSession must have a delegate of `BackgroundURLSessionDelegate` type.
```
mokagio:
```suggestion
    /// The `perform(request:...)` async function can be used in all non-background `URLSession` instances without any
    /// extra work. However, there is a requirement to make the function works with with background `URLSession` instances.
    /// That is the `URLSession` must have a delegate of `BackgroundURLSessionDelegate` type.
```

WordPressKit/HTTPClient.swift
```diff
@@ -49,6 +61,10 @@ extension URLSession {
         fulfilling parentProgress: Progress? = nil,
         errorType: E.Type = E.self
     ) async -> WordPressAPIResult<HTTPAPIResponse<Data>, E> {
+        if configuration.identifier != nil {
+            assert(delegate is BackgroundURLSessionDelegate, "Unexpected URLSession delegate type. See the `backgroundSession(configuration:)`")
```
mokagio:
```suggestion
            assert(delegate is BackgroundURLSessionDelegate, "Unexpected `URLSession` delegate type. See the `backgroundSession(configuration:)`")
```

WordPressKit/HTTPClient.swift
```diff
@@ -97,23 +113,50 @@ extension URLSession {
     ) throws -> URLSessionTask {
         var request = try builder.build(encodeBody: false)
 
+        // This additional `callCompletionFromDelegate` is added so that we can test `BackgroundURLSessionDelegate`
+        // in unit tests. Background URLSession doesn't work on unit tests, we have to create a non-background URLSession
+        // which has a `BackgroundURLSessionDelegate` delegate in order to test `BackgroundURLSessionDelegate`.
```
mokagio:
```suggestion
        // This additional `callCompletionFromDelegate` is added to unit test `BackgroundURLSessionDelegate`.
        // Background `URLSession` doesn't work on unit tests, we have to create a non-background `URLSession`
        // which has a `BackgroundURLSessionDelegate` delegate in order to test `BackgroundURLSessionDelegate`.
```

WordPressKit/HTTPClient.swift
```diff
@@ -97,23 +113,50 @@ extension URLSession {
...
-            return dataTask(with: request, completionHandler: completion)
+            if callCompletionFromDelegate {
+                task = dataTask(with: request)
+            } else {
+                task = dataTask(with: request, completionHandler: completion)
+            }
         }
+
+        if callCompletionFromDelegate {
+            assert(delegate is BackgroundURLSessionDelegate, "Unexpected URLSession delegate type. See the `backgroundSession(configuration:)`")
```
mokagio:
```suggestion
            assert(delegate is BackgroundURLSessionDelegate, "Unexpected `URLSession` delegate type. See the `backgroundSession(configuration:)`")
```

WordPressKit/HTTPClient.swift
```diff
@@ -205,3 +248,82 @@ extension Progress {
...
+    var responseBody = Data()
+    var completion: ((Data?, URLResponse?, Error?) -> Void)?
+}
+
+class BackgroundURLSessionDelegate: NSObject, URLSessionDataDelegate {
+
+    private var taskData = [Int: SessionTaskData]()
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
+        session.recieved(data, forTaskWithIdentifier: dataTask.taskIdentifier)
```
mokagio:
```suggestion
        session.received(data, forTaskWithIdentifier: dataTask.taskIdentifier)
```

WordPressKit/HTTPClient.swift
```diff
@@ -205,3 +248,82 @@ extension Progress {
...
+        self.taskData[taskID] = task
+    }
+
+    func set(completion: @escaping (Data?, URLResponse?, Error?) -> Void, forTaskWithIdentifier taskID: Int) {
+        updateData(forTaskWithIdentifier: taskID) {
+            $0.completion = completion
+        }
+    }
+
+    func recieved(_ data: Data, forTaskWithIdentifier taskID: Int) {
```
mokagio:
```suggestion
    func received(_ data: Data, forTaskWithIdentifier taskID: Int) {
```

WordPressKitTests/Utilities/URLSessionHelperTests.swift
```diff
@@ -227,6 +236,92 @@ class URLSessionHelperTests: XCTestCase {
...
+        if delegate.startedReceivingResponse {
+            XCTAssertTrue(delegate.completionCalled)
+        }
+    }
+
+}
+
+private class TestBackgroundURLSessionDelegate: BackgroundURLSessionDelegate {
+    var startedReceivingResponse = false
+    var completionCalled = false
```
mokagio, line 314: Nitpick: When building a spy I like making these `private(set)` to ensure only the spy itself can change their value and the tests cannot accidentally mess up with the state it tracks.

```suggestion
    private(set) var startedReceivingResponse = false
    private(set) var completionCalled = false
```

WordPressKitTests/Utilities/URLSessionHelperTests.swift
```diff
@@ -227,6 +236,92 @@ class URLSessionHelperTests: XCTestCase {
...
+        }
+
+        let builder = HTTPRequestBuilder(url: URL(string: "https://wordpress.org/hello")!)
+        let response = try await session.perform(request: builder, errorType: TestError.self).get()
+
+        try XCTAssertEqual(
+            sha256(XCTUnwrap(InputStream(url: file))),
+            sha256(InputStream(data: response.body))
+        )
+    }
```
mokagio, line 257: This is now the slowest test in the suite, multiplied by the number of subclasses. [Source](https://buildkite.com/organizations/automattic/analytics/suites/wordpress-ios-kit/runs/3b366dac-1daa-8c9d-87c0-d0d8c6d43ac6).

image

I suppose part of that is generating the large file at runtime. I wonder what the implications would be of tracking the file in the test suite for repo size and test duration. Not a big deal at this point, anyways.

mokagio (commented):
Tracking my comments, mostly so I can commit the typo-fix suggestions and eventually merge to move forward with the release.

Before approving, I'll take this for a ride via https://github.com/wordpress-mobile/WordPress-iOS/pull/22755.

mokagio (approved):
Followed the instructions in https://github.com/wordpress-mobile/WordPress-iOS/pull/22755 to test these changes. 

I first verified the crash with a recent App Center build, `pr22748-1a9193e`, then used the build from that PR.

IMG_3524

Worked as expected. I also run a photo upload directly from the app (that is, using the non-background `URLSession`) and verified it worked.

mokagio:
Taking over the merge so this can be part of the new version and be into WordPress 24.4

crazytonyli:
@mokagio Thanks for the review and code changes!

PR #740, opened 2024-03-04
Remove temp file once HTTP request completes

crazytonyli:
> [!Note]
> This PR is built on top of https://github.com/wordpress-mobile/WordPressKit-iOS/pull/739.

### Description

In some HTTP requests (i.e. uploading a large image from a background URL session), HTTP request body is written into a temp file first. Once the request completes (success of failure), the temp file should be deleted to avoid taking up device disk space.

### Testing Details

See the added unit tests.

---

### DIFF
--- a/WordPressKit/HTTPRequestBuilder.swift
+++ b/WordPressKit/HTTPRequestBuilder.swift
@@ -215,7 +215,7 @@ final class HTTPRequestBuilder {
         request.setValue("text/xml", forHTTPHeaderField: "Content-Type")
         let encoder = WPXMLRPCEncoder(method: xmlrpcRequest.method, andParameters: xmlrpcRequest.parameters)
         if forceWriteToFile {
-            let fileName = "\(ProcessInfo.processInfo.globallyUniqueString)_file.xmlrpc"
+            let fileName = "\(UUID().uuidString).xmlrpc"
             let fileURL = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent(fileName)
             try encoder.encode(toFile: fileURL.path)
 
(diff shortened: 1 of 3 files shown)


WordPressKit/HTTPRequestBuilder.swift
```diff
@@ -215,7 +215,7 @@ final class HTTPRequestBuilder {
         request.setValue("text/xml", forHTTPHeaderField: "Content-Type")
         let encoder = WPXMLRPCEncoder(method: xmlrpcRequest.method, andParameters: xmlrpcRequest.parameters)
         if forceWriteToFile {
-            let fileName = "\(ProcessInfo.processInfo.globallyUniqueString)_file.xmlrpc"
+            let fileName = "\(UUID().uuidString).xmlrpc"
```
crazytonyli, line 218: [Naming temp files using PID](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/e7d2c04c1be6c602c2f285643f6e1e19f935a8d7/WordPressKit/WordPressOrgXMLRPCApi.swift#L257-L261) can be problematic if there are concurrent uploads.

crazytonyli: I can't verify this assumption though, maybe Foundation API is safe enough to avoid concurrent writes to the same file. Nonetheless, I don't think we should reuse the same file, given temp files will now be removed after request completes.

mokagio: Using a unique name sounds like a good idea regardless 👍

mokagio (approved):

WordPressKit/HTTPRequestBuilder.swift
crazytonyli: Actually, I misread the code, the API is "globally unique string", not "PID". 🤦‍♂️ 

The original code is an alternative to `UUID`, which means this code change is kind of meaningless. 😄

PR #756, opened 2024-03-19
SPM Prep - Use XCTUnwrap instead of force-unwrapping for fixture paths in tests

mokagio:
### Description

This will make it easier to run the tests once the fixture files move around. 

If a file is misplaced, the test will fail instead of crashing.

Operationally, failing is better than crashing because it generates a
list of all misplaced tests, so that they can be addressed in one go. The alternative would be a run tests, tests crash, fix individual file path loop for as many misplaced fixture files there are.

Unfortunately, the diff is extensive. However, with the exception of `RemoteTestCase.swift` which changed in isolation in 007e11a, the pattern for all the changes is the same:

From:

```swift
func testSomething() {
    stub(condition: {  ... }) { _ in
      let stubPath = OHPathForFile("fixture.json", type(of: self))
      return fixture(filePath: stubPath!, ...)
    }
    ...
```

To:

```swift
func testSomething() throws { // test method now throws because of try XCTUnwrap
    // stubPath defined outside stub closure because the closure does not throw
    let stubPath = try XCTUnwrap(OHPathForFile("fixture.json", type(of: self))) 
    stub(condition: {  ... }) { _ in
      return fixture(filePath: stubPath, ...) // direct access instead of force unwrap
    }
    ...
```

### Testing Details

Green CI. This is only a unit tests implementation change.

### Next up

- Move files into `Sources/` and `Tests/`
- Then, new ground, define protocol abstraction to work around the Objective-C / Swift inter-op issue

---

### DIFF
--- a/WordPressKitTests/ReaderSiteServiceRemoteTests.swift
+++ b/WordPressKitTests/ReaderSiteServiceRemoteTests.swift
@@ -289,12 +289,12 @@ class ReaderSiteServiceRemoteTests: XCTestCase {
         XCTAssertTrue(failure)
     }
 
-    func testCheckSiteExistsAtURLSuccess() {
+    func testCheckSiteExistsAtURLSuccess() throws {
         let testURLString = "http://www.wordpress.com"
         let testURL = URL(string: testURLString)!
+        let stubPath = try XCTUnwrap(OHPathForFile("empty.json", type(of: self)))
         stub(condition: {request in request.url?.absoluteString == testURLString}) { _ in
-            let stubPath = OHPathForFile("empty.json", type(of: self))
-            return fixture(filePath: stubPath!, headers: ["Content-Type" as NSObject: "application/json" as AnyObject])
+            return fixture(filePath: stubPath, headers: ["Content-Type" as NSObject: "application/json" as AnyObject])
         }
 
         let expect = self.expectation(description: "One callback should be invoked")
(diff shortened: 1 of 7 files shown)


WordPressKitTests/ReaderSiteServiceRemoteTests.swift
```diff
@@ -289,12 +289,12 @@ class ReaderSiteServiceRemoteTests: XCTestCase {
...
 
-    func testCheckSiteExistsAtURLSuccess() {
+    func testCheckSiteExistsAtURLSuccess() throws {
         let testURLString = "http://www.wordpress.com"
         let testURL = URL(string: testURLString)!
+        let stubPath = try XCTUnwrap(OHPathForFile("empty.json", type(of: self)))
         stub(condition: {request in request.url?.absoluteString == testURLString}) { _ in
-            let stubPath = OHPathForFile("empty.json", type(of: self))
-            return fixture(filePath: stubPath!, headers: ["Content-Type" as NSObject: "application/json" as AnyObject])
+            return fixture(filePath: stubPath, headers: ["Content-Type" as NSObject: "application/json" as AnyObject])
```
mokagio, line 297: The header dictionary for content type JSON is repeated all over the place.

It could be DRYed in some form. I decided to keep it out because this PR is already noisy. Besides, the focus of this stream of work is to prepare for SPM. DRYing this definition would be useful, but doesn't bring us closer to that goal.

WordPressKitTests/RemoteTestCase.swift
```diff
@@ -134,15 +146,20 @@ extension RemoteTestCase {
                 return HTTPStubsResponse(error: notConnectedError)
             }
 
-            let stubPath = OHPathForFile(files[callCounter], type(of: self))
+            guard let stubPath = OHPathForFile(files[callCounter], type(of: self)) else {
+                XCTFail("Could not find file at path '\(files[callCounter])'.")
+                let error = NSError(domain: "RemoteTestCase", code: 0, userInfo: nil)
+                return HTTPStubsResponse(error: error)
+            }
```
mokagio, line 153: Had to fail in this way, also returning an error because we're inside a closure and the file path is computed based on the `callCounter` value.

crazytonyli (approved):

PR #759, opened 2024-03-20
Modernize `NSDate` and `Date` implementation introducing breaking changes

mokagio:
Follows up on @crazytonyli comments from the PR https://github.com/wordpress-mobile/WordPressKit-iOS/pull/758, which it targets.

I don't know why I felt like making a dedicated PR instead of pushing to that one directly 🤷‍♂️ Maybe because of the breaking changes...

---

### DIFF
--- a/WordPressKit/Date+WordPressCom.swift
+++ b/WordPressKit/Date+WordPressCom.swift
@@ -1,0 +1,19 @@
+extension Date {
+
+    /// Parses a date string
+    ///
+    /// Dates in the format specified in http://www.w3.org/TR/NOTE-datetime should be OK.
+    /// The kind of dates returned by the REST API should match that format, even if the doc promises ISO 8601.
+    ///
+    /// Parsing the full ISO 8601, or even RFC 3339 is more complex than this, and makes no sense right now.
+    ///
+    /// - SeeAlso: [WordPress.com REST API docs](https://developer.wordpress.com/docs/api/)
+    /// - Warning: This method doesn't support fractional seconds or dates with leap seconds (23:59:60 turns into 23:59:00)
+    static func with(wordPressComJSONString jsonString: String) -> Date? {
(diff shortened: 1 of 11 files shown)


WordPressKit/Date+WordPressCom.swift
```diff
@@ -0,0 +1,19 @@
...
+    /// - SeeAlso: [WordPress.com REST API docs](https://developer.wordpress.com/docs/api/)
+    /// - Warning: This method doesn't support fractional seconds or dates with leap seconds (23:59:60 turns into 23:59:00)
+    static func with(wordPressComJSONString jsonString: String) -> Date? {
+        DateFormatter.wordPressCom.date(from: jsonString)
+    }
+
+    var wordPressComJSONString: String {
+        DateFormatter.wordPressCom.string(from: self)
+    }
+}
```
crazytonyli, line 19: I feel like these extension functions don't really help much. They can just use the formatter instance `DateFormatter.wordPressCom` directly, right?

crazytonyli: Sorry, I just noticed `wordPressCom` is an internal API. Please ignore my comment above, if you intend to not expose the `DateFormatter` as public API, which I think is a good idea (leaving us option to switch over to [ISO8601DateFormatter](https://developer.apple.com/documentation/foundation/iso8601dateformatter/options/1643217-withinternetdatetime)).

crazytonyli (approved):
I'm okay with merging these two PR together as one. It might be easier for code review 🙈

WordPressKit/Date+WordPressCom.swift
mokagio: 👍 

Yes. This whole custom formatting feels odd. I think once we get the code in a better place in term of SPM, we could look at trimming some of those unnecessary custom bits in favor of standards

PR #760, opened 2024-03-21
SPM Prep – Add `@protocol` defining REST GET and POST methods

mokagio:
This PR is meant to get early feedback on the design direction for the abstraction layer required for SPM support.

It:

- Defines a minimal abstraction layer via Objective-C protocol tentatively named `WordPressComRESTAPIInterfacing`
- Makes `WordPressComRestApi` conform to the protocol
- Updates a few existing requests to use the abstraction, to ensure the various pieces are wired together correctly. 

_I'm aware of the CocoaPods validation failure. Working on fixing it by updating the paths definitions in the `podspec`._

### Testing Details

CI is enough.

## Next up

- Move files into Sources/ and Tests/

See also, https://github.com/wordpress-mobile/WordPressKit-iOS/pull/756 and #758 which is related to the SPM work but independent from this one.

---


Sources/APIInterface/include/WordPressComRESTAPIInterfacing.h
```diff
@@ -0,0 +1,15 @@
+@import Foundation;
+
+@protocol WordPressComRESTAPIInterfacing
```
mokagio: Two notes on the naming:

- I used "REST" and "API", instead of "Rest" and "Api" because they are acronyms and it feels more correct to have them all uppercase (when not used as the first word of a method).
- I used "Interfacing" to align with Swift naming conventions. In that context, an object conforming to this protocol should "interface with the WordPress.com REST API" which is exactly what `WordPressComRestApi` does.

I'm a bit concerned by the ambiguity between "interfacing" as a verb and "interface" as a noun often used in programming languages to define something similar to an Objective-C or Swift protocol. Any suggestion for an alternative name is welcome.

Sources/APIInterface/include/WordPressComRESTAPIInterfacing.h
```diff
@@ -0,0 +1,15 @@
...
+
+- (void)get:(NSString * _Nonnull)URLString
+ parameters:(NSDictionary<NSString *, NSObject *> * _Nullable)parameters
+    success:(void (^ _Nonnull)(id _Nonnull, NSHTTPURLResponse * _Nullable))success
+    failure:(void (^ _Nonnull)(NSError * _Nonnull, NSHTTPURLResponse * _Nullable))failure;
+
+- (void)post:(NSString * _Nonnull)URLString
```
mokagio: I decided not to uppercase `GET` and `POST` like elsewhere in the code because of the naming convention of starting methods names in lowercase.

Interestingly, even if uppercase, the conversion from Objective-C to Swift results in the word being lowercased in the generated Swift version.

mokagio: Also interesting, Xcode suggested the method signature. I think it's a new Xcode 15.2/15.3 feature? I was pleasantly surprised. Definitely saved me some time in figuring out the block signatures.

WordPressKit.xcodeproj/project.pbxproj
```diff
@@ -1550,6 +1552,30 @@
...
+			sourceTree = "<group>";
+		};
+		3FFCC0562BABC7D20051D229 /* include */ = {
+			isa = PBXGroup;
+			children = (
+				3FFCC0552BABC78B0051D229 /* WordPressComRESTAPIInterfacing.h */,
+			);
+			path = include;
+			sourceTree = "<group>";
+		};
```
mokagio: I took the chance of adding this new file to setup the `Sources/` SPM structure. 

After this lands, I'll make a dedicated PR that moves all the files in `Sources/WordPressKit` and a new `Tests/WordPressKitTests`.

Once that happens, I'll follow up with another PR that extracts `WordPressComRestApi` and the related types in something I'm tentatively naming "CoreAPI", as per https://github.com/wordpress-mobile/WordPressKit-iOS/pull/738

WordPressKit/ServiceRemoteWordPressComREST.h
```diff
@@ -23,6 +24,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, strong, readonly) WordPressComRestApi *wordPressComRestApi;
 
+/**
+ *  @brief      The interface to the WordPress.com API to use for performing REST requests.
+ *              This is meant to gradually replace `wordPressComRestApi`.
+ */
+@property (nonatomic, strong, readonly) id<WordPressComRESTAPIInterfacing> wordPressComRESTAPI;
```
mokagio, line 31: The aim of this PR is to get feedback on the direction. As such, I haven't gone through the considerable effort of replacing all usages of the concrete `WordPressComRestApi *wordPressComRestApi` with the abstracted `id<WordPressComRESTAPIInterfacing> wordPressComRESTAPI`.

Assuming the direction I envisioned resonates, that'll be my next step.

WordPressKit/ServiceRemoteWordPressComREST.m
```diff
@@ -26,6 +26,7 @@ - (instancetype)initWithWordPressComRestApi:(WordPressComRestApi *)wordPressComR
     self = [super init];
     if (self) {
         _wordPressComRestApi = wordPressComRestApi;
+        _wordPressComRESTAPI = wordPressComRestApi;
```
mokagio, line 29: Once all the concrete `WordPressComRestApi` usages will have been replaced, this init will take an `id<WordPressComRESTAPIInterfacing>` input, effectively decoupling `ServiceRemoteWordPressComREST` (Objective-C) from `WordPressComRestApi` (Swift) via `WordPressComRESTAPIInterfacing` (Objective-C).

This is the model implemented in https://github.com/Automattic/spm-abstraction-layer-demo which should allow us to organize the code in a Swift package that Xcode can compile.

crazytonyli (approved):

PR #766, opened 2024-03-25
SPM Prep – Use `WordPressComRESTAPIInterfacing` in `DomainServiceRemote` and `AccountServiceRemoteREST`

mokagio:
### Description

A small PR to get feedback on my approach to adopt the `WordPressComRESTAPIInterfacing` abstraction layer from #760 in Swift. 

This required a few minor tweaks, mainly to make sure the code across Objective-C and Swift deals with `Any` instead of `NSObject`. I think this is okay. `Any` is as loose a type as it comes in Swift, but the loss of type safety is only a matter of making the two languages working together. All the code I've dealt with so far uses stricter types at runtime, of course.

### Testing Details

Ensure CI is green.

### Next Steps

Assuming the implementation direction I've taken is appropriate, after this lands I'll move on with porting it to the rest of the Swift files. 

---

### DIFF
--- a/Sources/WordPressKit/WordPressAPI/WordPressComRestApi.swift
+++ b/Sources/WordPressKit/WordPressAPI/WordPressComRestApi.swift
@@ -615,20 +615,34 @@ extension WordPressComRestApi: WordPressComRESTAPIInterfacing {
     // The same applies for the other methods below.
     public func get(
         _ URLString: String,
-        parameters: [String: NSObject]?,
+        parameters: [String: Any]?,
         success: @escaping (Any, HTTPURLResponse?) -> Void,
         failure: @escaping (any Error, HTTPURLResponse?) -> Void
     ) -> Progress? {
-        GET(URLString, parameters: parameters, success: success, failure: failure)
+        GET(
+            URLString,
+            // It's possible `WordPressComRestApi` could be updated to use `[String: Any]` instead.
+            // But leaving that investigation for later.
+            parameters: parameters as? [String: AnyObject],
+            success: success,
+            failure: failure
+        )
     }
 
     public func post(
         _ URLString: String,
-        parameters: [String: NSObject]?,
+        parameters: [String: Any]?,
         success: @escaping (Any, HTTPURLResponse?) -> Void,
         failure: @escaping (any Error, HTTPURLResponse?) -> Void
     ) -> Progress? {
-        POST(URLString, parameters: parameters, success: success, failure: failure)
+        POST(
+            URLString,
+            // It's possible `WordPressComRestApi` could be updated to use `[String: Any]` instead.
+            // But leaving that investigation for later.
+            parameters: parameters as? [String: AnyObject],
+            success: success,
+            failure: failure
+        )
     }
 
     public func multipartPOST(
--- a/Sources/WordPressKit/Services/AccountServiceRemoteREST+SocialService.swift
+++ b/Sources/WordPressKit/Services/AccountServiceRemoteREST+SocialService.swift
@@ -23,7 +23,7 @@ extension AccountServiceRemoteREST {
                                        oAuthClientID: String,
                                        oAuthClientSecret: String,
                                        success: @escaping (() -> Void),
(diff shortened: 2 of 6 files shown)


WordPressKit/AccountServiceRemoteREST+SocialService.swift
```diff
@@ -23,7 +23,7 @@ extension AccountServiceRemoteREST {
                                        oAuthClientID: String,
                                        oAuthClientSecret: String,
                                        success: @escaping (() -> Void),
-                                       failure: @escaping ((NSError) -> Void)) {
+                                       failure: @escaping ((Error) -> Void)) {
```
mokagio: I don't think this counts as a breaking change, does it?

The API became looser not stricter. Any code passing `(NSError) -> Void` to failure should still compile now that the expected type is `(Error) -> Void`.

_Anyways, if it's a breaking change, I'll just add it to the existing list_ 😅

WordPressKit/WordPressComRESTAPIInterfacing.h
```diff
@@ -4,13 +4,17 @@
 
 @property (strong, nonatomic, readonly) NSURL * _Nonnull baseURL;
 
+/// - Note: `parameters` has `id` instead of the more common `NSObject *` as its value type so it will convert to `AnyObject` in Swift.
+///         In Swift, it's simpler to work with `AnyObject` than with `NSObject`. For example `"abc" as AnyObject` over `"abc" as NSObject`.
 - (NSProgress * _Nullable)get:(NSString * _Nonnull)URLString
-                   parameters:(NSDictionary<NSString *, NSObject *> * _Nullable)parameters
+                   parameters:(NSDictionary<NSString *, id> * _Nullable)parameters
```
mokagio: Does the explanation make sense?

crazytonyli: Yep. It makes sense to me. Some Foundation API uses the same "fake generic" signature, i.e. [UserDefaults](https://developer.apple.com/documentation/foundation/nsuserdefaults/1408563-dictionaryforkey)

WordPressKit/AccountServiceRemoteREST+SocialService.swift
crazytonyli: IMO, it's still a breaking change. The consumers may use `NSError.domain` in their code, which won't compile with this updated code.

One way to avoid this breaking change is casing as NSError in the `failure` callback below: `failure(error as NSError)`.

Sources/WordPressKit/WordPressAPI/WordPressComRestApi.swift
```diff
@@ -615,20 +615,34 @@ extension WordPressComRestApi: WordPressComRESTAPIInterfacing {
...
+            parameters: parameters as? [String: AnyObject],
+            success: success,
+            failure: failure
+        )
     }
 
     public func post(
         _ URLString: String,
-        parameters: [String: NSObject]?,
+        parameters: [String: Any]?,
```
crazytonyli, line 634: I don't _think_ this is a breaking change, because `[String: NSObject]` is a "sub-type" of `[String: Any]`, thus can be passed in here?

crazytonyli (approved):

WordPressKit/WordPressComRESTAPIInterfacing.h
mokagio: Awesome! TIL

WordPressKit/AccountServiceRemoteREST+SocialService.swift
mokagio: If that's okay with you, I'd keep `Error` and count this as a breaking change, then.

`Error` is more Swifty, which I think is the best approach going forward.

I'll merge under that assumption and the approval you already gave. Happy to followup with a fix PR if needed.

crazytonyli: 👍 I don't have a strong opinion on this.

PR #777, opened 2024-04-04
SPM Prep - Use `WordPressComRESTAPIInterfacing` for all sync GET and POST requests in the Swift code

mokagio:
### Description

Follows up on #766 applying the same pattern to all the `.GET` and `.POST` usages.

I changed each file/group of files with extensions in dedicated commits, in case one wants to review in chunks.

### Testing Details

Not all code is covered by unit tests, unfortunately. But, these changes are really about swapping interfaces without changing the implementation. They should be safe—last famous words.

---

### DIFF
--- a/Sources/WordPressKit/Services/ActivityServiceRemote.swift
+++ b/Sources/WordPressKit/Services/ActivityServiceRemote.swift
@@ -59,7 +59,7 @@ open class ActivityServiceRemote: ServiceRemoteWordPressComREST {
 
         let finalPath = self.path(forEndpoint: endpoint, withVersion: ._2_0)
 
-        wordPressComRestApi.GET(finalPath,
+        wordPressComRESTAPI.get(finalPath,
                                 parameters: nil,
                                 success: { response, _ in
                                     do {
@@ -105,7 +105,7 @@ open class ActivityServiceRemote: ServiceRemoteWordPressComREST {
             parameters["on"] = formatter.string(from: on) as AnyObject
         }
 
-        wordPressComRestApi.GET(path,
+        wordPressComRESTAPI.get(path,
                                 parameters: parameters,
                                 success: { response, _ in
                                     do {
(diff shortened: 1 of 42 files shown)


Sources/WordPressKit/Services/ActivityServiceRemote.swift
```diff
@@ -154,20 +154,25 @@ open class ActivityServiceRemote: ServiceRemoteWordPressComREST {
                                 }, failure: { error, _ in
                                     // FIXME: A hack to support free WPCom sites and Rewind. Should be obsolote as soon as the backend
                                     // stops returning 412's for those sites.
-                                    if error.domain == WordPressComRestApiEndpointError.errorDomain, error.code == WordPressComRestApiErrorCode.preconditionFailure.rawValue {
-                                        let status = RewindStatus(state: .unavailable)
-                                        success(status)
+                                    let nsError = error as NSError
```
crazytonyli, line 157: Not a blocker, but IMO it's better to avoid casting to NSError here by using pattern matching.

```swift
if let apiError = error as? WordPressAPIError<WordPressRestApiEndpointError>,
    case let .endpointError(endpointError) = apiError,
    endpointError.code == .preconditionFailure {
  // ....
}
```

Sources/WordPressKit/Services/JetpackServiceRemote.swift
```diff
@@ -74,16 +74,17 @@ public class JetpackServiceRemote: ServiceRemoteWordPressComREST {
...
+                                 success: { response, _ in
                                     if let response = response as? [String: Bool],
                                         let success = response[Constants.status] {
                                         completion(success, nil)
                                     } else {
                                         completion(false, JetpackInstallError(type: .installResponseError))
                                     }
-        }) { (error: NSError, _: HTTPURLResponse?) in
+        }) { error, _ in
+            let error = error as NSError
```
crazytonyli, line 87: Same as above.

crazytonyli (approved):

Sources/WordPressKit/Services/ActivityServiceRemote.swift
mokagio: Oh, right. 

I guess you are suggesting to used the better typed error to ensure what we are handling is a legit API error. If the casting fails, we are in a edge case and should act differently.

Under that assumption, I opened https://github.com/wordpress-mobile/WordPressKit-iOS/pull/779 to address. Thanks!

PR #778, opened 2024-04-04
SPM Prep - Isolate API version enum and convert to modern Objective-C

mokagio:
### Description

The end goal is to extract / duplicate the `pathForEndpoint:withVersion:` logic so that it can be used outside of `ServiceRemoteWordPressComRESTApiVersion`. The reason for that is that while all the sync calls can be used across Swift and Objective-C via #777, the `async` one cannot be used in Objective-C and therefore need Swift-only types that do not depend on `WordPressComRESTAPIInterfacing`

### Testing Details

No behavior change here. If the unit tests compile, we should be good.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- Remove unused `anonymousWordPressComRestApiWithUserAgent` method from `ServiceRemoteWordPressComREST` [TBD]
 
 ### New Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@ _None._
 
 ### Breaking Changes
 
-- Remove unused `anonymousWordPressComRestApiWithUserAgent` method from `ServiceRemoteWordPressComREST` [TBD]
+- Removed `anonymousWordPressComRestApiWithUserAgent` method from `ServiceRemoteWordPressComREST` [#778]
+- Renamed `ServiceRemoteWordPressComRESTApiVersion` to `WordPressComRestAPIVersion` [#778]
 
 ### New Features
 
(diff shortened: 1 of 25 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/WordPressKit/Services/AccountServiceRemoteREST+SocialService.swift
```diff
@@ -24,7 +24,7 @@ extension AccountServiceRemoteREST {
                                        oAuthClientSecret: String,
                                        success: @escaping (() -> Void),
                                        failure: @escaping ((Error) -> Void)) {
-        let path = self.path(forEndpoint: "me/social-login/connect", withVersion: ._1_1)
+        let path = self.path(forEndpoint: "me/social-login/connect", with: ._1_1)
```
mokagio: The compiler changed this, which was a bit annoying because it made the diff quite noisy.

I guess the compiler is happy with just "with" when the type on the other side is "strong", i.e. a `typedef NS_ENUM` that converts neatly to `enum` in Swift.

crazytonyli: You can avoid this name change by adding a `NS_SWIFT_NAME(path(forEndpoint:withVersion:))` to the function in the ObjC header.

mokagio: Nice! Much better diff, thanks for the tip @crazytonyli 

I think the Swift API is better this way, too. `with: ._1_1` was not as clear as `withVersion: ._1_1`.

crazytonyli (approved):

mokagio:
_Force pushed to resolve merge conflict without having to add merge commit._

PR #780, opened 2024-04-07
SPM Prep – Rewrite `RemoteBlogSettings` encoding and decoding in Swift

mokagio:
### Description

The Objective-C `BlogServiceRemoteREST` object [references](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/d1493069abec97755c69eaa3ea7bd1dc8e85c30b/Sources/WordPressKit/Services/BlogServiceRemoteREST.h#L5-L6) the Swift types `RemoteBlog` and `RemoteBlogSettings`.

This Objective-C to Swift dependency is incompatible with the separation we require for the SPM setup. 

The first step to address this is to move the custom encoding and decoding logic implemented within `BlogServiceRemoteREST` to Swift.

I did this by first writing tests for the existing implementation and then moving it to Swift using those tests as a guideline.

### Testing Details

See new tests in CI.

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/WordPressKit/Models/RemoteBlogSettings.swift
```diff
@@ -186,6 +186,182 @@ public class RemoteBlogSettings: NSObject {
...
+        let rawSettings = json[CodingKeys.settings.rawValue] as? [String: Any] ?? [:]
+
+        name = json[CodingKeys.name.rawValue] as? String
+        tagline = json[CodingKeys.tagline.rawValue] as? String
+        privacy = rawSettings[CodingKeys.privacy.rawValue] as? NSNumber
+        // The value here can be a String, so we need a custom conversion
+        languageID = rawSettings.number(forKey: CodingKeys.languageID.rawValue)
+        iconMediaID = rawSettings[CodingKeys.iconMediaID.rawValue] as? NSNumber
+        // The value here can be a String, so we need a custom conversion
+        gmtOffset = rawSettings.number(forKey: CodingKeys.gmtOffset.rawValue)
```
crazytonyli, line 249: Maybe I worry too much, but do you think we should use `number(forKey:)` to parse all properties instead of just these keys, because that's what the original Objective-C uses?.

We happen to know these weird cases, simply because we have one test json file for it. There is no indication in the original Objective-C code that these keys need special treatment. And, may be there are other keys needs a special "string" to "number" parsing, which is not detected by that one single test json file?

crazytonyli: Also, [the `string(forKey:)` parsing](https://github.com/wordpress-mobile/NSObject-SafeExpectations/blob/3cc5d8cca23f36efe711feadc37e3b0e13318d8c/Sources/NSObject-SafeExpectations/NSDictionary%2BSafeExpectations.m#L101-L102) in the original Objective-C code can parse a number to a string, which we don't do here. So, maybe we should use `string(forKey:)` and `number(forKey:)` instead of type casting?

PR #782, opened 2024-04-08
SPM Prep – Define new CoreAPI package

mokagio:
### Description

Moves some of the lower level Swift files that implement API requests into `Sources/CoreAPI` with matching tests.

Getting the tests to compile in this setup required some additional work in terms of conditional imports and an SPM-aware `Bundle` loader.

Notice that the package exists but is not used anywhere. I just wanted to lock this in, so to speak, to build on top of it later on.

### Testing Details

See CI. Although, it only shows that no regression has been introduced by this setup.

To test the SPM setup alone, one needs to:

1. Delete the `.xc-` workspace and project folder
2. Test via `xcodebuild`, for example with a lane like this one: https://github.com/wordpress-mobile/WordPressKit-iOS/blob/5c002d70d2eceb56159eeb871e3042e811e5ebc9/fastlane/Fastfile#L10-L29

OR

1. Open the project in Xcode via `Package.swift` (`open -a /Applications/Xcode.app .` should do it)
2. Run the tests

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Tests/CoreAPITests/WordPressOrgXMLRPCApiTests.swift
```diff
@@ -59,8 +63,12 @@ class WordPressOrgXMLRPCApiTests: XCTestCase {
             },
             failure: { (error, _) in
                 expect.fulfill()
-
+                // When building for SPM, the compiler doesn't generate the domain constant.
+#if SWIFT_PACKAGE
+                XCTAssertEqual(error.domain, "CoreAPI.WordPressOrgXMLRPCApiError")
+#else
                 XCTAssertEqual(error.domain, WordPressOrgXMLRPCApiErrorDomain)
+#endif
```
mokagio, line 71: @crazytonyli what do you think of this?

I worry it might backfire in the real world if people where to do a comparison on the domain with a `String` literal value.

Maybe it would be safer to explicitly define the domain in a `let` and update the code that inits `WordPressOrgXMLRPCApiError` to set it?

Sources/CoreAPI/WordPressComOAuthClient.swift
```diff
@@ -187,7 +187,7 @@ public final class WordPressComOAuthClient: NSObject {
             .mapSuccess { response in
                 let responseObject = try JSONSerialization.jsonObject(with: response.body)
 
-                WPKitLogVerbose("Received OAuth2 response: \(self.cleanedUpResponseForLogging(responseObject as AnyObject? ?? "nil" as AnyObject))")
+                // WPKitLogVerbose("Received OAuth2 response: \(self.cleanedUpResponseForLogging(responseObject as AnyObject? ?? "nil" as AnyObject))")
```
mokagio, line 190: Might need to introduce a new abstraction layer for logging, too.

Or, is logging at this level useful? Can we simply get rid of it altogether?

mokagio: I opened an issue about it https://github.com/wordpress-mobile/WordPressKit-iOS/issues/787

Sources/CoreAPI/WordPressComRestApiErrorDomain.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+/// Error domain of `NSError` instances that are converted from `WordPressComRestApiEndpointError`
+/// and `WordPressAPIError<WordPressComRestApiEndpointError>` instances.
+///
+/// See `extension WordPressComRestApiEndpointError: CustomNSError` for context.
+let WordPressComRestApiErrorDomain = "WordPressKit.WordPressComRestApiError"
+
+// WordPressComRestApiErrorDomain is accessible only in Swift and, since it's a global, cannot be made @objc.
+// As a workaround, here is a builder to init NSError with that domain and a method to check the domain.
```
crazytonyli: An alternative could be defining the same constant again in Objective-C and making it unavailable to Swift. That way you won't need to change those Objective-C files.

Tests/CoreAPITests/WordPressOrgXMLRPCApiTests.swift
crazytonyli: My preference would be going with the safer option, which is avoiding breaking changes. You can implement `CustomNSError` to set a custom domain, which means we shouldn't need to change this assertion here.

mokagio: Partially addressed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/782/commits/3f578954acf42bb50699d1885779f7d947e609fd. 

However, the tests still fail when build for SPM. I believe that has to do with the XML RPC returning `WordPressOrgXMLRPCApiError` or `WordPressOrgXMLRPCApiFault`.

I'll have to look into the latter as well, I guess.

mokagio: I found the source of the issue.

If one removes the `@objc` annotation, then the `CustomNSError` conversion code is used in its entirety. Otherwise, with the error `enum` declared as `@objc`, it seems like only the part that computes `userInfo` is called.

I verified this with an ad hoc package. I might put it online later. I'm unsure how to move forward in the meantime, because we do need the `@objc` visibility for `WordPressOrgXMLRPCApiError`. I'll keep thinking about it...

mokagio: Move my WIP work to https://github.com/wordpress-mobile/WordPressKit-iOS/pull/790 and rebased it off this branch.

Sources/CoreAPI/WordPressComRestApiErrorDomain.swift
mokagio: See 6fba3a19460d392576003415dd8d54576b16ad21

crazytonyli (approved):

PR #784, opened 2024-04-09
Remove `PostServiceRemoteExtended` and implementations

mokagio:
They will be moved into WordPress and Jetpack.

The reason for this is that they are only used in WordPress and Jetpack and that they are Swift extensions of Objective-C types which we are trying to isolate.

It's simpler to move these leaf types into the only consumer that uses them than trying to re-architect them.

### Description

Fixes #

### Testing Details

---


Sources/WordPressKit/Models/RemotePostParameters.swift
```diff
@@ -177,10 +177,14 @@ private enum RemotePostWordPressComCodingKeys: String, CodingKey {
...
-struct RemotePostCreateParametersWordPressComEncoder: Encodable {
+public struct RemotePostCreateParametersWordPressComEncoder: Encodable {
     let parameters: RemotePostCreateParameters
 
-    func encode(to encoder: Encoder) throws {
+    public init(parameters: RemotePostCreateParameters) {
+        self.parameters = parameters
+    }
+
+    public func encode(to encoder: Encoder) throws {
```
mokagio, line 187: When the type became `public`, a `public init` became necessary because the compiler does not synthesize one outside of the package.

As for `encode(to:)`, it needs to be `public` because `Encodable` is declared as part of the `public` type interface. 

🤔 In hindsight, I suppose the `Encodable` part could be moved into an `internal` extension, but I think it's okay to leave it as is. Let me know if you can think of a strong reason to hide it.

Sources/WordPressKit/WordPressAPI/WordPressAPIError.swift
```diff
@@ -18,13 +18,10 @@ public enum WordPressAPIError<EndpointError>: Error where EndpointError: Localiz
...
     case unacceptableStatusCode(response: HTTPURLResponse, body: Data)
     /// The API call returned an HTTP response that WordPressKit can't parse. Receiving this error could be an indicator that there is an error response that's not handled properly by WordPressKit.
-    case unparsableResponse(response: HTTPURLResponse?, body: Data?, underlyingError: Error)
+    case unparsableResponse(response: HTTPURLResponse?, body: Data?, underlyingError: Error = URLError(.cannotParseResponse))
     /// Other error occured.
     case unknown(underlyingError: Error)
 
-    static func unparsableResponse(response: HTTPURLResponse?, body: Data?) -> Self {
-        return WordPressAPIError<EndpointError>.unparsableResponse(response: response, body: body, underlyingError: URLError(.cannotParseResponse))
-    }
```
mokagio, line 27: I didn't look up when, but relatively recently, I think, Swift acquired default values for associated parameters. So we can remove this builder method in favor of defining the underlying error as a default for the case.

Sources/WordPressKit/WordPressAPI/WordPressComRestApi.swift
```diff
@@ -173,9 +177,6 @@ open class WordPressComRestApi: NSObject {
         }
     }
 
-    @objc func setInvalidTokenHandler(_ handler: @escaping () -> Void) {
-        invalidTokenHandler = handler
-    }
```
mokagio, line 178: Became redundant once `invalidTokenHandler` became a property in the interface.

kean:
@mokagio This code is a work in progress, and the respective PRs that integrate it in WordPress-iOS haven't been merged yet. Please hold off on making changes to this code until 24.8. 

Initially, the idea and the naming of `PostServiceRemoteExtended` came from the need to add just two new methods to the service, but it turned out most of the existing method needed to be replaced.

mokagio:
Hi @kean 🤔 

There I was thinking hacking around WordPressKit was safe because it was relatively quiet these days 😅 

What's your timeline for that work?

My priority right now is to get WordPressKit split into packages so that the majority of it can then go into WordPress and we can avoid back and forth like this.

**Maybe I can help get your project over the line?**

> Initially, the idea and the naming of `PostServiceRemoteExtended` came from the need to add just two new methods to the service, but it turned out most of the existing method needed to be replaced.

Interesting. Did you consider making all the addition in WordPress itself? Or perhaps there's internal utils that need to be accessed in WordPressKit to get what you need done?

"Replaced" is an interesting choice of word. Do you mean that the now there are methods that are no longer in use? Or is it a matter of the new feature you and team are working on needs a lot of new methods that will live in parallel to the existing ones?

Thanks for the info.

kean:
> Did you consider making all the addition in WordPress itself? Or perhaps there's internal utils that need to be accessed in WordPressKit to get what you need done?
> Do you mean that the now there are methods that are no longer in use? Or is it a matter of the new feature you and team are working on needs a lot of new methods that will live in parallel to the existing ones?

We've added new methods for creating and updating posts. I'm not sure it's possible to implement them outside of the framework, but I haven't tried or considered it. The old methods will no longer be used by the app once 24.8 rolls out.

PR #788, opened 2024-04-12
SPM Prep – Move some `async perform` types at the `WordPressComRESTAPIInterface` level

mokagio:
### Description

This was part of https://github.com/wordpress-mobile/WordPressKit-iOS/pull/784 but I've been asked to wait on the actual `extension` removal because of conflicting WIP work. 

All these changes should be fine to merge regardless, as are mostly about changing access control, injecting new params, and moving methods at the protocol level.

### Testing Details

See CI.

---

### DIFF
--- a/Sources/APIInterface/include/WordPressComRESTAPIInterfacing.h
+++ b/Sources/APIInterface/include/WordPressComRESTAPIInterfacing.h
@@ -6,6 +6,12 @@
 
 @property (strong, nonatomic, readonly) NSURL * _Nonnull baseURL;
 
+/// The key with which to specify locale in the parameters of a request.
+@property (strong, nonatomic, readonly) NSString * _Nonnull localeKey;
+
+/// Whether the user's preferred language locale should be appended.
+@property (nonatomic, readonly) BOOL appendsPreferredLanguageLocale;
+
 /// - Note: `parameters` has `id` instead of the more common `NSObject *` as its value type so it will convert to `AnyObject` in Swift.
 ///         In Swift, it's simpler to work with `AnyObject` than with `NSObject`. For example `"abc" as AnyObject` over `"abc" as NSObject`.
 - (NSProgress * _Nullable)get:(NSString * _Nonnull)URLString
--- a/Sources/APIInterface/include/WordPressComRESTAPIInterfacing.h
+++ b/Sources/APIInterface/include/WordPressComRESTAPIInterfacing.h
@@ -18,6 +18,10 @@
 /// The value with which to specify locale in the parameters of a request.
 @property (strong, nonatomic, readonly) NSString * _Nonnull localeValue;
 
+@property (strong, nonatomic, readonly) NSURLSession * _Nonnull urlSession;
+
+@property (strong, nonatomic, readonly) void (^ _Nullable invalidTokenHandler)(void);
+
 /// - Note: `parameters` has `id` instead of the more common `NSObject *` as its value type so it will convert to `AnyObject` in Swift.
 ///         In Swift, it's simpler to work with `AnyObject` than with `NSObject`. For example `"abc" as AnyObject` over `"abc" as NSObject`.
 - (NSProgress * _Nullable)get:(NSString * _Nonnull)URLString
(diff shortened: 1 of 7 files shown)


Sources/APIInterface/include/WordPressComRESTAPIInterfacing.h
```diff
@@ -6,6 +6,22 @@
...
+
+/// The key with which to specify locale in the parameters of a request.
+@property (strong, nonatomic, readonly) NSString * _Nonnull localeKey;
+
+/// The value with which to specify locale in the parameters of a request.
+@property (strong, nonatomic, readonly) NSString * _Nonnull localeValue;
+
+@property (strong, nonatomic, readonly) NSURLSession * _Nonnull urlSession;
+
+@property (strong, nonatomic, readonly) void (^ _Nullable invalidTokenHandler)(void);
```
mokagio, line 23: All these properties went in the protocol to make the `async perform` methods work.

The `async perform` methods, which like in `WordPressComRestApi` are Swift-only, of course.

I considered using an extension or a different protocol, but it seemed neat to have it all in `WordPressComRESTAPIInterfacing`, so that Swift code can used an upgraded version of the same interface that the Objective-C code uses. This will actually be useful in the cases where we have an ObjC superclass with the API client conforming to `WordPressComRESTAPIInterfacing` and a subclass using the API client, too.

Sources/WordPressKit/Models/RemotePostParameters.swift
```diff
@@ -198,10 +198,14 @@ private enum RemotePostWordPressComCodingKeys: String, CodingKey {
...
-struct RemotePostCreateParametersWordPressComEncoder: Encodable {
+public struct RemotePostCreateParametersWordPressComEncoder: Encodable {
     let parameters: RemotePostCreateParameters
 
-    func encode(to encoder: Encoder) throws {
+    public init(parameters: RemotePostCreateParameters) {
+        self.parameters = parameters
+    }
+
+    public func encode(to encoder: Encoder) throws {
```
mokagio, line 208: When the type became `public`, a `public init` became necessary because the compiler does not synthesize one outside of the package.

As for `encode(to:)`, it needs to be `public` because `Encodable` is declared as part of the `public` type interface. 

🤔 In hindsight, I suppose the `Encodable` part could be moved into an `internal` extension, but I think it's okay to leave it as is. Let me know if you can think of a strong reason to hide it.

Sources/WordPressKit/WordPressAPI/WordPressAPIError.swift
```diff
@@ -18,14 +18,10 @@ public enum WordPressAPIError<EndpointError>: Error where EndpointError: Localiz
...
     case unacceptableStatusCode(response: HTTPURLResponse, body: Data)
     /// The API call returned an HTTP response that WordPressKit can't parse. Receiving this error could be an indicator that there is an error response that's not handled properly by WordPressKit.
-    case unparsableResponse(response: HTTPURLResponse?, body: Data?, underlyingError: Error)
+    case unparsableResponse(response: HTTPURLResponse?, body: Data?, underlyingError: Error = URLError(.cannotParseResponse))
     /// Other error occured.
     case unknown(underlyingError: Error)
 
-    static func unparsableResponse(response: HTTPURLResponse?, body: Data?) -> Self {
-        return WordPressAPIError<EndpointError>.unparsableResponse(response: response, body: body, underlyingError: URLError(.cannotParseResponse))
-    }
```
mokagio, line 27: I didn't look up when, but relatively recently, I think, Swift acquired default values for associated parameters. So we can remove this builder method in favor of defining the underlying error as a default for the case.

Sources/WordPressKit/WordPressAPI/WordPressComRestApi.swift
```diff
@@ -173,9 +177,6 @@ open class WordPressComRestApi: NSObject {
         }
     }
 
-    @objc func setInvalidTokenHandler(_ handler: @escaping () -> Void) {
-        invalidTokenHandler = handler
-    }
```
mokagio, line 178: Became redundant once `invalidTokenHandler` became a property in the protocol.

PR #790, opened 2024-04-15
Make `WordPressOrgXMLRPCApiError` conform to `CustomNSError`

mokagio:
See discussion at https://github.com/wordpress-mobile/WordPressKit-iOS/pull/782#discussion_r1561853022

---


Sources/CoreAPI/WordPressOrgXMLRPCApiErrorDomain.swift
```diff
@@ -0,0 +1,30 @@
...
+    @objc
+    static func wordPressOrgXMLRPCApiError(
+        code: Int,
+        userInfo: [String: Any]?
+    ) -> NSError {
+        NSError(
+            domain: WordPressOrgXMLRPCApiErrorDomain,
+            code: code,
+            userInfo: userInfo
+        )
```
crazytonyli: If you create a `WordPressOrgXMLRPCApiError` error here, cast it to `NSError`, and return it to Objective-C code, the instance can be cast back to `WordPressOrgXMLRPCApiError` in Swift code when the error is passed from Objective-C code to Swift.

I don't suppose you can cast this `NSError` instance to `WordPressOrgXMLRPCApiError`? It would be pretty magical if you can. 🤔

mokagio: Oh, this is a shortsighted leftover from before I realized that the error domain constant could be defined in `APIInterface` and therefore be available to packages in both Objective-C and Swift.

I shall remove this utils like I did for `WordPressComRestApiErrorDomain`.

Sources/CoreAPI/WordPressOrgXMLRPCApiError.swift
```diff
@@ -0,0 +1,40 @@
...
+        /// The serialization of the request failed.
+        case requestSerializationFailed
+        /// The serialization of the response failed.
+        case responseSerializationFailed
+        /// An unknown error occurred.
+        case unknown
+    }
+
+    let code: Code
+}
```
mokagio, line 18: @crazytonyli I ended up having to wrap the error code enum into an `Error` type like you did for the REST API error.

Otherwise, between the `@objc` visibility (needed because WordPress-Authenticator uses it 🙄 ) and the `Error` conformance, I couldn't find a way to stop the compiler from automatically generating `WordPressOrgXMLRPCApiErrorDomain` for `@objc public enum WordPressOrgXMLRPCApiError: Int, Error`.

I think this is an okay solution. Apart for the fact that's the only one I could get to work, it's, as far as I can see, the same setup as the other error, which makes the implementation consistent.

Tests/CoreAPITests/WordPressOrgXMLRPCApiTests.swift
```diff
@@ -128,15 +119,10 @@ class WordPressOrgXMLRPCApiTests: XCTestCase {
             failure: { (error, _) in
                 expect.fulfill()
 
-                XCTAssertTrue(error is WordPressOrgXMLRPCApiError)
```
mokagio, line 131: This is the only change in behavior from the rewrite.

The `NSError`-converted error (via `asNSerror()`) that the API passes to this failure branch doesn't convert back to `WordPressOrgXMLRPCApiError`.

crazytonyli: I might have missed something. Does that mean the `error` in the failure block was `WordPressOrgXMLRPCApiError`, but in this PR it's changed to something else?

PR #793, opened 2024-04-18
Fix TCP connection issues

kean:
### Description

Related PR https://github.com/wordpress-mobile/WordPress-iOS/pull/23035

### Testing Details

---

### DIFF
--- a/Sources/CoreAPI/WordPressComRestApi.swift
+++ b/Sources/CoreAPI/WordPressComRestApi.swift
@@ -152,12 +152,6 @@ open class WordPressComRestApi: NSObject {
         super.init()
     }
 
-    deinit {
-        for session in [urlSession, uploadURLSession] {
-            session.finishTasksAndInvalidate()
-        }
-    }
-
     /// Cancels all outgoing tasks asynchronously without invalidating the session.
     public func cancelTasks() {
         for session in [urlSession, uploadURLSession] {
@@ -321,6 +315,12 @@ open class WordPressComRestApi: NSObject {
             builder = builder.query(defaults: [URLQueryItem(name: localeKey, value: preferredLanguageIdentifier)])
         }
 
+        if let oAuthToken {
+            builder = builder.header(name: "Authorization", value: "Bearer \(oAuthToken)")
+        }
+        if let userAgent {
+            builder = builder.header(name: "User-Agent", value: userAgent)
+        }
         return builder
     }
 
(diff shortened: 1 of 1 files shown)


Sources/CoreAPI/WordPressComRestApi.swift
```diff
@@ -333,34 +333,28 @@ open class WordPressComRestApi: NSObject {
...
 
-    private func sessionConfiguration(background: Bool) -> URLSessionConfiguration {
-        let configuration = background ? URLSessionConfiguration.background(withIdentifier: self.backgroundSessionIdentifier) : URLSessionConfiguration.default
+    private static var urlSession = URLSession(configuration: .default)
 
+    private func makeBackgroundSession() -> URLSession {
+        let configuration = URLSessionConfiguration.background(withIdentifier: self.backgroundSessionIdentifier)
         var additionalHeaders: [String: AnyObject] = [:]
-        if let oAuthToken = self.oAuthToken {
+        if let oAuthToken {
```
kean, line 350: I don't know if this is required for background sessions? Probably not.

Sources/CoreAPI/WordPressComRestApi.swift
```diff
@@ -333,34 +333,28 @@ open class WordPressComRestApi: NSObject {
...
-            return URLSession(configuration: configuration)
+        if backgroundUploads {
+            return makeBackgroundSession()
         }
+        return WordPressComRestApi.urlSession
     }()
 
-    private func sessionConfiguration(background: Bool) -> URLSessionConfiguration {
-        let configuration = background ? URLSessionConfiguration.background(withIdentifier: self.backgroundSessionIdentifier) : URLSessionConfiguration.default
+    private static var urlSession = URLSession(configuration: .default)
```
kean, line 345: TBD: apply the same treatment to .org and XMLRPC.

Sources/CoreAPI/WordPressComRestApi.swift
```diff
@@ -152,12 +152,6 @@ open class WordPressComRestApi: NSObject {
         super.init()
     }
 
-    deinit {
-        for session in [urlSession, uploadURLSession] {
-            session.finishTasksAndInvalidate()
-        }
-    }
```
crazytonyli, line 159: One side effect is ongoing tasks won't be invalidated at all. I wonder if we should make `WordPressComRestApi` accept a URLSession instance, which the app can create upon use signing in and invalidate upon logout.

kean: Is there a need to invalidate it? The requests are signed on a per-request basis now.

Something does need to change with the `WordPressComRestApi/invalidateAndCancelTasks` usage. The lifetime of `_wordPressComRestApi` is tied to a `WPAccount`, which is also problematic: there is one instance per thread, and the instances could get reset.

crazytonyli: > Is there a need to invalidate it?

Good question. I don't know for sure though. Using a shared URLSession singleton between user sessions is a change introduced in this PR, which we'll need to test to make sure it doesn't break things.

IMO, the best practice is URLSession instance should be per user session. Upon user logs in, they get a new URLSession instance and uses it until logs out. Once logs out, the URLSession instance should be destroyed. That's similar to how Safari uses one URLSession for each tab.

> The lifetime of _wordPressComRestApi is tied to a WPAccount, which is also problematic: there is one instance per thread, and the instances could get reset.

Yes, definitely. That's another big piece of tech debt.

kean: > IMO, the best practice is URLSession instance should be per user session. Upon user logs in, they get a new URLSession instance and uses it until logs out. Once logs out, the URLSession instance should be destroyed. That's similar to how Safari uses one URLSession for each tab.

I completely agree. The app should have a repository that could control the lifetime of the objects to match the lifetime of the user session. It will be a better fix for this issue as well (assuming it actually fixes it because I don't have any hard evidence yet).

PR #795, opened 2024-04-23
Add endpoint for fetching subscriber history

guarani:
### Description

Add endpoint for fetching subscriber history.

Part of https://github.com/wordpress-mobile/WordPress-iOS/issues/23044

### Testing Details

- Check if `StatsSubscribersSummaryDataTests` passes

---

### DIFF
--- a/Sources/WordPressKit/Models/Stats/StatsSubscribersSummaryData.swift
+++ b/Sources/WordPressKit/Models/Stats/StatsSubscribersSummaryData.swift
@@ -1,0 +1,52 @@
+import Foundation
+import WordPressShared
+
+public struct StatsSubscribersSummaryData: Decodable, Equatable {
+    let history: [SubscriberData]
+
+    private enum CodingKeys: String, CodingKey {
+        case history = "data"
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        history = try container.decode([SubscriberData].self, forKey: .history)
+    }
+
+    public struct SubscriberData: Decodable, Equatable {
+        let date: Date
+        let count: Int
+
+        private enum CodingKeys: Int, CodingKey {
+            case date = 0
+            case count = 1
+        }
+
+        public init(from decoder: any Decoder) throws {
+            var container = try decoder.unkeyedContainer()
+
+            date = ISO8601DateFormatter().date(from: try container.decode(String.self)) ?? Date()
+            count = try container.decode(Int.self)
+        }
+    }
+}
+
+extension StatsSubscribersSummaryData {
+    public static var pathComponent: String {
+        return "stats/subscribers"
+    }
+
+    public init?(jsonDictionary: [String: AnyObject]) {
+        do {
+            let jsonData = try JSONSerialization.data(withJSONObject: jsonDictionary, options: [])
+            let decoder = JSONDecoder.apiDecoder
+            self = try decoder.decode(Self.self, from: jsonData)
+        } catch {
+            return nil
+        }
+    }
+
+    public static func queryProperties(quantity: Int, unit: String) -> [String: String] {
+        return ["quantity": String(quantity), "unit": unit]
+    }
+}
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/WordPressKit/Models/Stats/StatsSubscribersSummaryData.swift
```diff
@@ -0,0 +1,67 @@
...
+            self.count = count
+        }
+    }
+
+    public init?(jsonDictionary: [String: AnyObject]) {
+        guard
+            let fields = jsonDictionary["fields"] as? [String],
+            let data = jsonDictionary["data"] as? [[Any]],
+            let dateIndex = fields.firstIndex(of: "period"),
+            let countIndex = fields.firstIndex(of: "subscribers")
```
staskus, line 50: This pattern of having an array of fields is in a way super-flexible but unfortunately incompatible with any benefits that `Codable` may provide.

Sources/WordPressKit/Models/Stats/StatsSubscribersSummaryData.swift
```diff
@@ -0,0 +1,67 @@
...
+                let date = StatsSubscribersSummaryData.dateFormatter.date(from: dateString),
+                let count = elements[countIndex] as? Int
+            else {
+                return nil
+            }
+
+            return SubscriberData(date: date, count: count)
+        }
+
+        let sorted = history.compactMap { $0 }.sorted(by: { $0.date.compare($1.date) == .orderedAscending })
```
staskus: nitpick

```suggestion
        let sorted = history.compactMap { $0 }.sorted { $0.date < $1.date }
```

Sources/WordPressKit/Services/StatsServiceRemoteV2.swift
```diff
@@ -316,6 +316,30 @@ private extension StatsServiceRemoteV2 {
     }
 }
 
+// MARK: - Subscribers Data
+
+public extension StatsServiceRemoteV2 {
```
staskus: I think it's not necessary to create a separate method for subscribers endpoint.

I compared the API response and request with `stats/visits` and it looks to be following the same structure:

Same structure between subscribers and visits

What it means that it's enough for `StatsSubscribersSummaryData` to conform to `StatsTimeIntervalData` protocol, and follow the same rules when parsing as `StatsSummaryTimeIntervalData`. 

Then you could reuse `statsService.getData` method, which will also support quantity, end dates, `StatsPeriodUnit` type parameters out of the box. We won't need to be making changes to the API when adding additional date-picking capabilities to subscribers.

Example usage:

```
        statsService.getData(for: .day, endingOn: StatsDataHelper.currentDateForSite(), limit: 30) { (data: StatsSubscribersSummaryData?, error: Error?) in
            DispatchQueue.main.async {
                if let data = data {
                    cache.setValue(data, key: cacheKey)
                    self.chartSummary.send(.success(data))
                }
            else {
                    if cachedData == nil {
                        self.chartSummary.send(.error)
                    }
                }
            }
        }
```

I tested and it all works well. If you agree, feel free to cherry-pick this commit:

https://github.com/wordpress-mobile/WordPressKit-iOS/compare/task/subscribers-api-endpoint...task/subscribers-api-endpoint-review?expand=1

Sources/WordPressKit/Models/Stats/StatsSubscribersSummaryData.swift
guarani: I tried to use codable here with things like `nestedUnkeyedContainer` and `singleValueContainer` but kept running into issues. 

 I think leaving this as-is is fine for now., but please let me know if you happen to come across a way of using `Codable` 🙇

Sources/WordPressKit/Models/Stats/StatsSubscribersSummaryData.swift
guarani: Makes sense 👍

Sources/WordPressKit/Services/StatsServiceRemoteV2.swift
guarani: That's a very good idea! I went ahead and added it 👍

staskus (approved):
LGTM 👍

PR #823, opened 2024-11-20
Release 17.3.0

twstokes:
Cuts version 17.3.0.


mokagio (approved):
Thanks @twstokes 

I'm going to merge this, tag, and release.

mokagio:
Process breadcrumb: I got a server error when attempting to publish 17.3.0.

https://buildkite.com/automattic/wordpresskit-ios/builds/1908#01934d17-d6b4-43c7-9005-0fcb799f5222

Before retrying, I thought I'd update to the latest CocoaPods version. There's nothing in the changelog that suggests there's a fix for the error we run into, but it's usually best to make sure one is on the latest version before investigating failures. 

The publishing build is triggered by pushing a tag. Which mean to retry it we should bump the version again... To keep things simple, and given the only change is a tooling one, I deleted the 17.3.0 tag here on GitHub and will tag the commit with the CP update as the new 17.3.0 then try again.

mokagio:
🙄 so it seems like 17.3.0 was actually published despite the server error

image

For what is worth, `bundle exec pod search WordPressKit` did not and still does not show that version at this time.

PR #835, opened 2025-03-31
Do not show disconnected Jetpack sites

crazytonyli:
### Description

The app displays sites inaccessible Jetpack self-hosted sites(for example, deleted Jurassic Ninja sites). Similar to deleted WP.com sites, when switching to those sites, the app displays almost empty content. This PR removes those sites from API responses.

[WordPressKit.zip](https://github.com/user-attachments/files/19658656/WordPressKit.zip)

### Testing Details

---

### DIFF
--- a/Sources/WordPressKit/Models/RemoteBlog.swift
+++ b/Sources/WordPressKit/Models/RemoteBlog.swift
@@ -37,10 +37,10 @@ import Foundation
     /// Features available for the current blog's plan.
     public var planActiveFeatures = [String]()
 
-    /// Indicates whether the jetpack connection is active.
+    /// Indicates whether the site is a Jetpack site or not.
     public var jetpack: Bool = false
 
-    /// Indicates whether it's a Jetpack connected site.
+    /// Indicates whether the site is connected to WP.com via `jetpack-connection`.
     public var jetpackConnection: Bool = false
 
     /// Boolean indicating whether the current user has Admin privileges, or not.
--- a/Sources/WordPressKit/Models/RemoteBlog.swift
+++ b/Sources/WordPressKit/Models/RemoteBlog.swift
@@ -37,8 +37,11 @@ import Foundation
     /// Features available for the current blog's plan.
     public var planActiveFeatures = [String]()
 
-    /// Indicates whether it's a jetpack site, or not.
+    /// Indicates whether the jetpack connection is active.
     public var jetpack: Bool = false
+
+    /// Indicates whether it's a Jetpack connected site.
+    public var jetpackConnection: Bool = false
 
     /// Boolean indicating whether the current user has Admin privileges, or not.
     public var isAdmin: Bool = false
@@ -70,6 +73,7 @@ import Foundation
         self.url = json.string(forKey: "URL") ?? ""
         self.xmlrpc = json.string(forKeyPath: "meta.links.xmlrpc")
         self.jetpack = json.number(forKey: "jetpack")?.boolValue ?? false
+        self.jetpackConnection = json.number(forKey: "jetpack_connection")?.boolValue ?? false
         self.icon = json.string(forKeyPath: "icon.img")
         self.capabilities = json.object(forKey: "capabilities") as? [String: Bool] ?? [:]
         self.isAdmin = json.number(forKeyPath: "capabilities.manage_options")?.boolValue ?? false
--- a/Sources/WordPressKit/Services/AccountServiceRemoteREST.m
+++ b/Sources/WordPressKit/Services/AccountServiceRemoteREST.m
@@ -391,7 +391,16 @@ MagicLinkFlow const MagicLinkFlowSignup = @"signup";
         // Exclude deleted sites from query result, since the app does not handle deleted sites properly.
         // I tried to use query arguments `site_visibility=visible` and `site_activity=active`, but neither excludes
         // deleted sites.
-        return !blog.isDeleted;
+        if (blog.isDeleted) {
+            return false;
+        }
+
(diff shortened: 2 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pmusolino (commented):
LGTM, but I'm waiting for the final PR in the WordPress-iOS repo before approving it!

Sources/WordPressKit/Models/RemoteBlog.swift
```diff
@@ -37,9 +37,12 @@ import Foundation
     /// Features available for the current blog's plan.
     public var planActiveFeatures = [String]()
 
-    /// Indicates whether it's a jetpack site, or not.
+    /// Indicates whether the jetpack connection is active.
```
kean: Shouldn't this comment apply to the `jetpackConnection` property?

```swift
/// Indicates whether the jetpack connection is active.
public var jetpack: Bool = false
```

Let's follow the Swift naming conversion and name these something like:

```swift
var isJetpackConnectionActive: Bool { get }
var isJetpackConnected: Bool { get } 
```

```swift
Indicates whether it's a Jetpack connected site.
```

Does this mean that the site is a non-WP.com site?

Sources/WordPressKit/Services/AccountServiceRemoteREST.m
```diff
@@ -391,7 +391,16 @@ - (NSArray *)remoteBlogsFromJSONArray:(NSArray *)jsonBlogs
         // Exclude deleted sites from query result, since the app does not handle deleted sites properly.
         // I tried to use query arguments `site_visibility=visible` and `site_activity=active`, but neither excludes
         // deleted sites.
-        return !blog.isDeleted;
+        if (blog.isDeleted) {
+            return false;
+        }
+
+        // Exclude sites that are connected via Jetpack, but without an active Jetpack connection.
+        if (blog.jetpackConnection && !blog.jetpack) {
```
kean, line 399: What does WP.com do in the same scenario? Does it also remove the site from your account? What does "active Jetpack" mean? Can a site temporarily become inactive?

Sources/WordPressKit/Models/RemoteBlog.swift
crazytonyli: I have updated the API doc to match the properties on the HTTP API https://github.com/wordpress-mobile/WordPressKit-iOS/pull/835/commits/777c6515b1b31dd86f427f7cc3ca41ec920d2add.

> Does this mean that the site is a non-WP.com site?

I believe it's possible that atomic sites have `jetpack_connection == true & jetpack == false`. Just by `jetpack_connection` property alone, we can't tell if a site is a WP.com site.

Sources/WordPressKit/Services/AccountServiceRemoteREST.m
crazytonyli: `jetpack_connection = true` means the site was connected to WP.com via jetpack.

`jetpack = true` means WP.com can still communicate with the site via Jetpack (a.k.a, jetpack connection is active).

You typically get `jetpack_connection = true && jetpack = false` when the site was connected to WP.com a while ago, but its domain expired.

crazytonyli: WP.com does not show sites with `jetpack_connection = true && jetpack = false`, hence the fix.

There is more complicated logic on WP.com (see [here](https://github.com/Automattic/wp-calypso/blob/b57c069b819c61f4527c7737094ad2fd8ebed499/client/my-sites/sites/index.jsx#L50)) to decide whether to show a site or not. I didn't try to understand all of them. But this PR fixes the particular case where the site domain has expired.

crazytonyli: > Can a site temporarily become inactive?

I think that's entirely possible, like when the site is down, or for whatever reason that WP.com decides that it can't communicate with the site.

kean: I'm not entirely sure how the logic on WP.com translate to this:

```swift
if ( hasJetpackActivePlugins( site ) && ! isJetpackSiteOrJetpackCloud( site ) ) { return false; }
```

Is there a good way to test it?

> jetpack = true means WP.com can still communicate with the site via Jetpack (a.k.a, jetpack connection is active).

There is also a chance the UX on WP.com isn't optimal. WP.com can't not if the site was actually deleted or if it's a temporarily issue, can it? As a user, I'd probably expect  it to show some kind of a message like "site is not responding" with a way to remove it from my WP.com account.

I'd also suggest moving filtering (the entire logic inside `wpkit_filter`) to the app. It's not `WordPressKit` job to decide what app displays.

crazytonyli: I didn't follow the exact filtering logic on calypso. This is the code I referenced when looking into the API responses.

```javascript
// Sites connected through A4A plugin are listed on wordpress.com/sites even when Jetpack is deactivated.
export const isDisconnectedJetpackAndNotAtomic = ( site: SiteExcerptNetworkData ) => {
	return ! site?.is_wpcom_atomic && site?.jetpack_connection && ! site?.jetpack;
};
```

> I'd also suggest moving filtering (the entire logic inside wpkit_filter) to the app. It's not WordPressKit job to decide what app displays.

Agreed. I'll do that.

kean (approved):
Please make sure to rebase and re-create the binary if needed.

PR #816, opened 2024-07-12
WordPress-iOS Edition [Do Not Merge]

kean:
### Description

This branch contains an archived version of WordPressKit designed to be used exclusively by [WordPress-iOS](https://github.com/wordpress-mobile/WordPress-iOS). Consider it a release branch.

This version of WordPressKit is distributed using SwiftPM as a binary target. It has no dependencies to `WordPressShared` and `UIDeviceIdentifier`, and it imports `wpxmlrpc` and `NSObject_SafeExtensions` as `@_implementationOnly`, so there are not exposed in the public interface.

### Updating WordPressKit

On an off chance that the WordPressKit version used by WordPress-iOS needs a patch, please submit it by targeting this branch.

#### Deployment Instructions

The deployment currently requires a few manual steps, but we don't expect it to be needed often.

- Run `.buildkite/create-xcframeworks.sh` to create binaries
- Upload `.build/xcframeworks/WordPressKit.zip` to this repo
- Update binary target in `Package.swift` with a new URL and checksum*
- Push the new changes to your branch
- (After testing is done) Merge changes in `wpios-edition`

> checksum is echoed at the end of the create-xcframeworks.sh script

PR #838, opened 2025-06-18
AINFRA-283 - Set `queue: mac` explicitly in pipeline

mokagio:
This will allow adopting an improved default pipeline upload step on the CI infrastructure side of the setup. It also makes it clear what queue is in use just by looking at the pipeline file.

Notice it does not introduce the standard `shared-pipeline-vars` file because to use that we first need the default pipeline upload step in the IaC configuration.

See https://linear.app/a8c/issue/AINFRA-283

> [!NOTE]  
> Tests are failing but that's unrelated with the changes in this PR, they fail on #816, too. Given how WordPressKit has changed to ship as an XCFramework _and_ that currently WordPress-Jetpack-Reader is the only consumer, I decided not to investigate. I'm happy to if required, but my current understanding is that this repo is on life support and will see minimal changes, meaning investing in does not give the best RoI.

> [!NOTE]  
> I targeted #816 instead of `trunk` because WordPress-Jetpack-Reader is the only consumer of this library. As such, I propose merging #816 and making it the main branch, adding a note on the state of the project. See also pbArwn-7sp-p2

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,6 +1,9 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 ---
 
+agents:
+  queue: mac
+
 env:
   IMAGE_ID: xcode-15.3-v3
 
@@ -54,5 +57,3 @@ steps:
       - "validate"
       - "lint"
     if: build.tag != null
-    agents:
-      queue: "mac"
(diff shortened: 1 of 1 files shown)


AliSoftware (approved):
> Tests are failing but that's unrelated with the changes in this PR, they fail on https://github.com/wordpress-mobile/WordPressKit-iOS/pull/816, too. Given how WordPressKit has changed to ship as an XCFramework and that currently WordPress-Jetpack-Reader is the only consumer, I decided not to investigate

👍

PR #822, opened 2024-11-14
Add support for sending magic links to inexistent accounts

hichamboushaba:
### Description

Part of https://github.com/woocommerce/woocommerce-ios/issues/14419

This PR updates the magic link request with an option to pass the option `create_account`, this option allows to use the same request for signup flow needed for the above task.

### Testing Details
Use https://github.com/woocommerce/woocommerce-ios/pull/14431

---

- [ ] Please check here if your pull request includes additional test coverage.
- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #839, opened 2025-06-18
Add search_text parameter to Activity Logs

kean:
Add `search_text` and `rewindable` support to Jetpack activity logs endpoint

PR #840, opened 2025-06-19
Merge `trunk` + update README in wpios-edition

mokagio:
See internal discussion at pbArwn-7sp-p2

I opened a PR for this because https://github.com/wordpress-mobile/WordPressKit-iOS/pull/816 cannot be merged without first resolving conflicts. There is only one conflict, the deleted `podspec`. 

While the `podspec` should stay deleted, this PR is useful to show the changes from #822. I think they are fine to merge and bring into WordPress-Jetpack-Reader, but I'll let @kean and @crazytonyli give they're more expert opinion on it.

After this land in #816, we can merge that into `trunk`.


kean (approved):
LGTM

Please note I have an open PR targeting `wpios-edition`. It looks like I need to update some tests though. https://github.com/wordpress-mobile/WordPressKit-iOS/pull/839

PR #842, opened 2025-06-26
Fix CI jobs

crazytonyli:
What says in the title.


kean:
Nice, thanks! My bad – looks like I used Swift 6.2 features there.

kean (approved):

PR #843, opened 2025-07-18
New Stats

kean:
### New Features
  - **Email Opens Stats**: Added support for fetching email opens statistics with new `StatsEmailOpensData` model
  - **Archive Stats**: Introduced `StatsArchiveTimeIntervalData` for generalized archive endpoint parsing
  - **Hourly Stats Support**: Added `StatsPeriodUnit.hour` to enable hourly statistics data
  - **Site Metrics**: Added `StatsSiteMetricsResponse` for improved site metrics handling

  ### API Enhancements
  - Extended `StatsServiceRemoteV2.getData` with `startDate`, `timeZone`, and `fields` parameters
  - Added `summarize` parameter support for stats requests
  - Enhanced `StatsPostDetails` to include all available data fields

[WordPressKit.zip](https://github.com/user-attachments/files/21582269/WordPressKit.zip)


kean:
[WordPressKit.zip](https://github.com/user-attachments/files/21322294/WordPressKit.zip)

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><code>Package.swift</code> was changed without updating its corresponding <code>Package.resolved</code>. Please resolve the Swift packages as appropriate to your project setup (e.g. in Xcode or by running <code>swift package resolve</code>).</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/WordPressKit/Models/Stats/Emails/StatsEmailOpensData.swift
```diff
@@ -0,0 +1,42 @@
...
+        case opensRate = "opens_rate"
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        totalSends = try container.decodeIfPresent(Int.self, forKey: .totalSends)
+        uniqueOpens = try container.decodeIfPresent(Int.self, forKey: .uniqueOpens)
+        totalOpens = try container.decodeIfPresent(Int.self, forKey: .totalOpens)
+        opensRate = try container.decodeIfPresent(Double.self, forKey: .opensRate)
+    }
```
crazytonyli: Do we need this function? It's the same as the default derived implementation, if I'm not mistaken?

Sources/WordPressKit/Models/Stats/Emails/StatsEmailOpensData.swift
```diff
@@ -0,0 +1,42 @@
...
+}
+
+extension StatsEmailOpensData {
+    public init?(jsonDictionary: [String: AnyObject]) {
```
crazytonyli, line 25: `throw` instead of silencing the error?

Sources/WordPressKit/Models/Stats/Emails/StatsEmailOpensData.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+public struct StatsEmailOpensData: Decodable, Equatable {
+    public let totalSends: Int?
+    public let uniqueOpens: Int?
+    public let totalOpens: Int?
+    public let opensRate: Double?
```
crazytonyli, line 7: Should all these properties be null? Will the response ever be null, like no `total_sends` or `total_send: null` in the response?

Sources/WordPressKit/Models/Stats/StatsSubscribersSummaryData.swift
```diff
@@ -17,6 +17,13 @@ extension StatsSubscribersSummaryData: StatsTimeIntervalData {
         return "stats/subscribers"
     }
 
+    static var hourlyDateFormatter: DateFormatter {
+        let df = DateFormatter()
+        df.locale = Locale(identifier: "en_US_POS")
```
crazytonyli: Is the id supposed to be `en_US_POSIX`?

Sources/WordPressKit/Models/Stats/Time Interval/StatsSiteMetricsResponse.swift
```diff
@@ -0,0 +1,108 @@
...
+            visitors: fields.firstIndex(of: Metric.visitors.rawValue),
+            likes: fields.firstIndex(of: Metric.likes.rawValue),
+            comments: fields.firstIndex(of: Metric.comments.rawValue),
+            posts: fields.firstIndex(of: Metric.posts.rawValue)
+        )
+
+        let dateFormatter = makeDateFormatter(for: period)
+
+        self.data = data.compactMap { data in
+            guard let periodDate = dateFormatter.date(from: data[periodIndex] as? String ?? "") else {
```
crazytonyli: Nitpick:

```suggestion
            guard let date = data[periodIndex] as? String, let periodDate = dateFormatter.date(from: date) else {
```

Sources/WordPressKit/Models/Stats/StatsPostDetails.swift
```diff
@@ -6,6 +8,81 @@ public struct StatsPostDetails: Equatable {
...
+            guard
+                let postID = jsonDictionary["ID"] as? Int,
+                let title = jsonDictionary["post_title"] as? String
+            else {
+                return nil
+            }
+
+            let dateFormatter = DateFormatter()
+            dateFormatter.locale = Locale(identifier: "en_US_POSIX")
+            dateFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
```
crazytonyli, line 54: You'd need to set the `timeZone` property to parse GMT date strings that do not have a time zone in them, right?

Sources/WordPressKit/Services/StatsServiceRemoteV2.swift
```diff
@@ -99,30 +107,48 @@ open class StatsServiceRemoteV2: ServiceRemoteWordPressComREST {
...
         let pathComponent = TimeStatsType.pathComponent
         let path = self.path(forEndpoint: "sites/\(siteID)/\(pathComponent)/", withVersion: ._1_1)
 
-        let staticProperties = ["period": period.stringValue,
+        let dateFormatter = period == .hour ? hourlyDateFormatter : periodDataQueryDateFormatter
+
+        var staticProperties = ["period": period.stringValue,
                                 "unit": unit?.stringValue ?? period.stringValue,
-                                "date": periodDataQueryDateFormatter.string(from: endingOn)] as [String: AnyObject]
+                                "date": dateFormatter.string(from: endingOn)] as [String: AnyObject]
```
crazytonyli, line 127: Nitpick: I think you can get rid of all the `as` statements below if you declare the `staticProperties` as `[String: Any]`.

Sources/WordPressKit/Services/StatsServiceRemoteV2.swift
```diff
@@ -342,6 +372,28 @@ public extension StatsServiceRemoteV2 {
     }
 }
 
+// MARK: - Email Opens
+
+public extension StatsServiceRemoteV2 {
+    func getEmailOpens(for postID: Int, completion: @escaping ((StatsEmailOpensData?, Error?) -> Void)) {
```
crazytonyli, line 378: What do you think about using async functions instead of calls for the new API calls? I presume the callers are written in Swift too, which may prefer async functions?

Sources/WordPressKit/Models/Stats/Time Interval/StatsSiteMetricsResponse.swift
```diff
@@ -0,0 +1,108 @@
...
+
+        let indices = (
+            views: fields.firstIndex(of: Metric.views.rawValue),
+            visitors: fields.firstIndex(of: Metric.visitors.rawValue),
+            likes: fields.firstIndex(of: Metric.likes.rawValue),
+            comments: fields.firstIndex(of: Metric.comments.rawValue),
+            posts: fields.firstIndex(of: Metric.posts.rawValue)
+        )
+
+        let dateFormatter = makeDateFormatter(for: period)
```
crazytonyli, line 75: There is this comment above: "/// Periods date in the site timezone."

It's impossible to parse a date string if we don't know what the site timezone is, right? So, it's incorrect to use any `DateFormatter` instance. I think the correct solution is either passing the site timezone from the app[^1] to the parsing function here, or returning a plain string and letting the app parse it.

[^1]: Or, making another HTTP request to get site timezone, which does feel like wasteful.

crazytonyli:
There are merge conflicts and failing unit tests.

Sources/WordPressKit/Models/Stats/Emails/StatsEmailOpensData.swift
kean: No really - removing.

Sources/WordPressKit/Models/Stats/Emails/StatsEmailOpensData.swift
kean: I auto-generated this code with the rest of the file as a reference. It is consistent with the rest of `StatsServiceRemoteV2` and how it's used. I'd suggest keeping it as it's throw-away code anyway.

Sources/WordPressKit/Models/Stats/Emails/StatsEmailOpensData.swift
kean: I'd change it, but it's a throw-away code, so I'm not sure it's worth reworking it and the usage. I'm not sure I have any energy left at this point 😪

Sources/WordPressKit/Models/Stats/StatsSubscribersSummaryData.swift
kean: Updated. This code is unused – the new stats don't request hourly data for subscribers and there is no hourly data for subscribers. The existing formatters in this screen use `en_US_POS`. I didn't want to make changes to the existing code in the scope of this change.

Sources/WordPressKit/Models/Stats/Time Interval/StatsSiteMetricsResponse.swift
kean: Applied.

Sources/WordPressKit/Models/Stats/StatsPostDetails.swift
kean: This is an existing issue across all existing decodes in this service. The date is actually returned in the site time zone. The decoders don't know what is the time zone of the site. The response also doesn't contain it. So it end's up creating a wrong timestamp in your local time zone.

I added code to map it in StatsService https://github.com/wordpress-mobile/WordPress-iOS/blob/b82374372485a73d0a3f49bc5cbc3886a3179767/Modules/Sources/JetpackStats/Services/StatsService.swift#L355, but I think I made a mistake at one of the layers. I opened a ticket and will debug it later on Monday: https://linear.app/a8c/issue/CMM-642/dates-are-in-the-wrong-timezone. The fix needs to be on the app level. I want to keep the WPKit behavior as is so it doesn't break the legacy screens.

Sources/WordPressKit/Services/StatsServiceRemoteV2.swift
kean: Yeah, it support `Any`, but I just didn't want to change it since it's the existing code. There was already too many changes, and I wanted to keep the changes to the existing code minimum.

Sources/WordPressKit/Models/Stats/Time Interval/StatsSiteMetricsResponse.swift
kean: I'll revisit this in the scope of https://github.com/wordpress-mobile/WordPressKit-iOS/pull/843#discussion_r2248943444. The app knows the time zone, so it will need to perform the conversion. 

In reality, `StatsServiceRemoteV2` simply uses incorrect types to represent dates. If It doesn't know what time zone the dates component belong to, it should never returned `Date` (timestamp). It ends up always creating wrong time stamps (in your app current time zone instead of the site time zone). Unfortunately, it is what it is.

kean:
I merged `trunk`, recompiled the framework, and the tests are now successful. Looks like it was related to the SwiftLint changes.

crazytonyli (approved):

